PR #3, opened 2025-01-28
Add the frr configuration rendering and logic

fedepaol:
Add the FRR configuration rendering and logic.


internal/frrconfig/frrconfig.go
```diff
@@ -0,0 +1,45 @@
...
+type Action string
+
+const (
+	Test         = "test"
+	Reload       = "reload"
+	reloaderPath = "/usr/lib/frr/frr-reload.py"
+)
+
+// Config reloads the frr configuration at the given path.
+func Update(path string) error {
```
oribon, line 20: mismatch between comment and name

internal/frrconfig/frrconfig.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"log/slog"
+	"os/exec"
+)
+
+type Action string
+
+const (
+	Test         = "test"
+	Reload       = "reload"
```
oribon, line 15: nit: shouldn't these be Actions?

internal/frrconfig/frrconfig_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	cmd := exec.Command(os.Args[0], cs...)
+	cmd.Env = append(env, os.Environ()...)
+	return cmd
+}
+
+// This is not a real test. It's used in case fakeExecCommand is used in place of exec.Command.
+// In that case the command execution is redirected to this function.
+func TestFakeReloadHelper(t *testing.T) {
```
oribon, line 71: seems like a bit of over-machinery to make CombinedOutput happy (iiuc), can't execCommand be more like `exec.Exec` instead of `exec.Command` ?

oribon (commented):
still didn't go over the fourth commit, need to go over everything again 😅

internal/frrconfig/frrconfig.go
fedepaol: yup, lost in translation :sweat_smile:

internal/frrconfig/frrconfig_test.go
fedepaol: Not sure I understood your comment. Anyhow, this kind of test is heavily inspired (ie, blatantly copied) from this talk https://youtu.be/8hQG7QlcLBk?t=2052

Basically the idea is, to test the interaction with a sub process, we run the very same executable (in this case the test one) in such a way that instead of running a test (TestFakeReloadHelper), it behaves as a test double.
With this machinery in place, we allow the test (the double) to intercept what the real program argument would be.

oribon: sorry let me expand what I meant 😅 
currently the `execCommand` var (that we test and replace) is a `func(name string, arg ...string) *exec.Cmd`, which is what caused the introduction of this `TestFakeReloadHelper` approach (and the whole machinery).
since we don't really use what `exec.Cmd` has to offer (other than CombinedOutput) the `execCommand` var could just be a `func(name string, arg ...string) (string,error)` which is a lot more straight-forward to test imo (and what we usually did so far).

cmd/reloader/main.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/openperouter/openperouter/internal/frrconfig"
+	"github.com/openperouter/openperouter/internal/logging"
+)
+
+var frrConfigPath string
+
+func main() {
+	var bindAddress string
+	var logLevel string
+	flag.StringVar(&bindAddress, "bindaddress", "0.0.0.0:9080", "The address the reloader endpoint binds to. ")
```
oribon, line 39: can the default be localhost (or we intend this to be public)?

internal/frrconfig/frrconfig_test.go
oribon, line 1: some changes went to the 4th commit

oribon (commented):
2 nits, other than that lgtm

internal/frrconfig/frrconfig_test.go
fedepaol: so iiuc, you are a suggesting not to sub process and interact with the external process through the machinery, but just to mock the command execution itself.

cmd/reloader/main.go
fedepaol: the reloader process is going to be part of the frr pod, which is going to be network namespaced, and as such shouldn't be accessible from outside. We need the controller pod to be able to talk to it, and I am not sure localhost would work in this case.

internal/frrconfig/frrconfig_test.go
fedepaol: done

PR #6, opened 2025-01-28
Makefile: do not set images on daemons

fedepaol:
There are no daemons yet, so not setting images / namespace.


fedepaol:
Yoloing, fixing failing tests.

PR #7, opened 2025-01-29
Add an ipam providing package

fedepaol:
We need to assign ips:

- A different IP per VTEP per node
- The same IP for the PE side on each node
- A different IP for the host side on each node

Which is what this package is providing.

### DIFF
--- a/internal/ipam/ipam.go
+++ b/internal/ipam/ipam.go
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package ipam
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/apparentlymart/go-cidr/cidr"
+	gocidr "github.com/apparentlymart/go-cidr/cidr"
+	"github.com/openperouter/openperouter/internal/ipfamily"
+)
+
+type Veths struct {
+	HostSide net.IPNet
+	PeSide   net.IPNet
+}
+
+// VethIPs returns the IPs for the host side and the PE side
+// for a given pool on the ith node.
+func VethIPs(pool string, index int) (Veths, error) {
+	_, cidr, err := net.ParseCIDR(pool)
+	if err != nil {
+		return Veths{}, fmt.Errorf("failed to parse pool %s: %w", pool, err)
+	}
+	peSide, err := cidrElem(cidr, 0)
+	if err != nil {
+		return Veths{}, err
+	}
+	hostSide, err := cidrElem(cidr, index+1)
+	if err != nil {
+		return Veths{}, err
+	}
+	return Veths{HostSide: *hostSide, PeSide: *peSide}, nil
+}
+
+// VTEPIp returns the IP to be used for the local VTEP on the ith node.
+func VTEPIp(pool string, index int) (net.IPNet, error) {
+	_, cidr, err := net.ParseCIDR(pool)
+	if err != nil {
+		return net.IPNet{}, fmt.Errorf("failed to parse pool %s: %w", pool, err)
+	}
+
+	ips, err := sliceCIDR(cidr, index, 1)
+	if err != nil {
+		return net.IPNet{}, err
+	}
+	if len(ips) != 1 {
+		return net.IPNet{}, fmt.Errorf("vtepIP, expecting 1 ip, got %v", ips)
+	}
+	res := net.IPNet{
+		IP:   ips[0].IP,
+		Mask: net.CIDRMask(32, 32),
+	}
+	if ipfamily.ForAddress(res.IP) == ipfamily.IPv6 {
+		res.Mask = net.CIDRMask(128, 128)
+	}
+	return res, nil
+}
+
+// cidrElem returns the ith elem of len size for the given cidr.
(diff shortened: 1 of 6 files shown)


internal/ipam/ipam.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+type Veths struct {
+	HostSide      net.IPNet
+	ContainerSide net.IPNet
+}
+
+// VethIPs returns the IPs for the host side and the PE side
+// for a given pool on the ith node.
+func VethIPs(pool string, index int) (Veths, error) {
```
oribon, line 19: a bit unrelated (without the whole context), where would the index come from (a node will know it is the x node?)

internal/ipam/ipam.go
```diff
@@ -0,0 +1,98 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package ipam
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/apparentlymart/go-cidr/cidr"
```
oribon, line 9: nit: import this as gocidr (since we're using cidr as a var in some places)

internal/ipam/ipam.go
```diff
@@ -0,0 +1,98 @@
...
+		return net.IPNet{}, err
+	}
+	if len(ips) != 1 {
+		return net.IPNet{}, fmt.Errorf("vtepIP, expecting 1 ip, got %v", ips)
+	}
+	return ips[0], nil
+}
+
+// cidrElem returns the ith elem of len size for the given cidr.
+func cidrElem(pool string, mask int, index int) (*net.IPNet, error) {
```
oribon, line 49: can this receive a cidr net.IPNet instead of the pool+mask?

internal/ipam/ipam.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, fmt.Errorf("failed to get %d address from %s: %w", index, pool, err)
+	}
+	return &net.IPNet{
+		IP:   ip,
+		Mask: net.CIDRMask(mask, 32),
+	}, nil
+}
+
+// sliceCIDR returns the ith block of len size for the given cidr.
+func sliceCIDR(pool string, index, size int) ([]net.IPNet, error) {
```
oribon, line 66: can this receive a cidr net.IPNet instead of string?

internal/ipam/ipam.go
```diff
@@ -0,0 +1,98 @@
...
+	res := []net.IPNet{}
+	for i := 0; i < size; i++ {
+		ipIndex := size*index + i
+		ip, err := cidr.Host(ipNet, ipIndex)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get %d address from %s: %w", ipIndex, pool, err)
+		}
+		ipNet := net.IPNet{
+			IP:   ip,
+			Mask: net.CIDRMask(31, 32),
```
oribon, line 81: not sure I understand the 31 here 😅  (I'd expect this to use size)

internal/ipam/ipam.go
```diff
@@ -0,0 +1,98 @@
...
+
+// VTEPIp returns the IP to be used for the local VTEP on the ith node.
+func VTEPIp(pool string, index int) (net.IPNet, error) {
+	ips, err := sliceCIDR(pool, index, 1)
+	if err != nil {
+		return net.IPNet{}, err
+	}
+	if len(ips) != 1 {
+		return net.IPNet{}, fmt.Errorf("vtepIP, expecting 1 ip, got %v", ips)
+	}
```
oribon, line 44: is this needed? (feels like not trusting sliceCIDR)

internal/ipam/ipam.go
```diff
@@ -0,0 +1,98 @@
...
+			Mask: net.CIDRMask(31, 32),
+		}
+
+		res = append(res, ipNet)
+	}
+
+	return res, nil
+}
+
+// IPsInCDIR returns the number of IPs in the given CIDR.
```
oribon, line 90: nit: CDIR

internal/ipam/ipam.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"fmt"
+	"net"
+
+	"github.com/apparentlymart/go-cidr/cidr"
+)
+
+type Veths struct {
+	HostSide      net.IPNet
+	ContainerSide net.IPNet
```
oribon, line 14: is there a reason this is not `PESide`?

internal/ipam/ipam_test.go
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package ipam
+
+import (
+	"testing"
+)
+
+func TestIPam(t *testing.T) {
```
oribon, line 9: should this be TestVTEPIp/sliceCIDR?

internal/ipam/ipam.go
fedepaol: this is mostly because I wanted to ensure I am getting a value here (ie asking for a number out of the range). I will also add a test for this

fedepaol:
Ok, I think I fixed all the comments and improved the unit tests...

PR #57, opened 2025-04-29
Work with podman

fedepaol:
Containerlab has an experimental podman support which works in rootful mode. Here we allow users to deploy / test with both rootful podman and docker by simply setting the CONTAINER_ENGINE variable to podman.


Makefile
```diff
@@ -299,3 +312,9 @@ cutrelease: bumpversion generate-all-in-one helm-docs
 build-validator: ginkgo ## Build Ginkgo test binary.
 	CGO_ENABLED=0 $(GINKGO) build -tags=externaltests ./internal/hostnetwork
 	mv internal/hostnetwork/hostnetwork.test $(VALIDATOR_PATH)
+
+.PHONY: install-containerlab
+install-containerlab:
+	@if ! command -v clab >/dev/null 2>&1; then \
+	    curl -sL https://containerlab.dev/setup | sudo -E bash -s "install-containerlab"; \
```
oribon, line 314: should we pin a clab version here?

clab/setup.sh
```diff
@@ -3,14 +3,30 @@ set -euo pipefail
...
+CONTAINER_ENGINE=${CONTAINER_ENGINE:-"docker"}
+CONTAINER_ENGINE_CLI="docker"
 KUBECONFIG_PATH=${KUBECONFIG_PATH:-"$(pwd)/kubeconfig"}
-KIND_BIN=${KIND_BIN:-"kind"}
+KIND=${KIND:-"kind"}
 CLAB_VERSION=0.64.0
 
 KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-pe-kind}"
 
+PODMAN_OPTION=""
```
oribon: nit: can we rename this to runtime_option?

clab/setup.sh
```diff
@@ -3,14 +3,30 @@ set -euo pipefail
...
+PODMAN_KIND_ENV="KIND_EXPERIMENTAL_PROVIDER=podman"
```
oribon: is this used?

Makefile
fedepaol: yep, this made me feel bad but that setup script runs either apt or dnf to deploy clab and I did not find a straightforward way to pin a version locally, nor I was able to make it work with podman in podman.
Let me try to spend a bit more trying to pursue this last option

clab/setup.sh
fedepaol: probably a leftover

Makefile
fedepaol: after sweating a bit I am giving up. I resorted to notify the user of the need to deploy clab vs deploying it ourselves, seems more respectful as we are deploying something on the host of the user.
The docker variant still runs with docker in docker, so in CI we won't have version skimming issues.

Makefile
```diff
@@ -86,11 +86,19 @@ BRANCH = $(shell git rev-parse --abbrev-ref HEAD)
...
+	fi
 
 .PHONY: docker-push
 docker-push: ## Push docker image with the manager.
-	$(CONTAINER_TOOL) push ${IMG}
+	@if [ "$(CONTAINER_ENGINE)" = "podman" ]; then \
+		sudo $(CONTAINER_ENGINE) push; \
+	else \
+		$(CONTAINER_ENGINE) push; \
+	fi
```
oribon, line 102: why is {IMG} dropped?

clab/kind/setup.sh
oribon, line 1: why are we removing this? (and `setupworker.sh`)

oribon (commented):
I _think_ the commits got into eachother, not sure

clab/kind/setup.sh
fedepaol: because the purpouse of those were only to assign the ip, which is now done by the tool that is fed with the ip matrix.

Makefile
fedepaol: my bad!

clab/setup.sh
```diff
@@ -35,26 +31,38 @@ docker run --rm --privileged \
...
-kind load docker-image quay.io/frrouting/frr:9.1.0 --name pe-kind
-kind load docker-image gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1 --name pe-kind
-kind load docker-image quay.io/metallb/frr-k8s:v0.0.17 --name pe-kind
+load_image_to_kind quay.io/frrouting/frr:9.1.0 frr9
+load_image_to_kind quay.io/frrouting/frr:10.2.1 frr10
+load_image_to_kind gcr.io/kubebuilder/kube-rbac-proxy:v0.13.1 rbacproxy
+load_image_to_kind quay.io/metallb/frr-k8s:v0.0.17 frrk8s
 
-kind --name pe-kind get kubeconfig > $KUBECONFIG_PATH
+${KIND_COMMAND} --name pe-kind get kubeconfig > $KUBECONFIG_PATH
```
oribon, line 50: same

oribon (commented):
last few nits, lgtm otherwise

clab/setup.sh
fedepaol: done

PR #58, opened 2025-05-07
Add Helm charts

oribon:
We add helm charts (to be used eventually by the operator as well), and add an e2e lane that verifies the deployment


charts/openperouter/templates/perouter.yaml
fedepaol, line 1: let's call this "router.yaml"

fedepaol:
Can you copy also the CRDs upon regeneration? Like, when we run make generate the crds under the chart are automatically upgraded.

charts/openperouter/templates/perouter.yaml
oribon: done

fedepaol:
LGTM

Issue #21, opened 2025-02-28
Add a validating webhook


fedepaol:
We should intercept bad configurations upfront

PR #68, opened 2025-06-05
Add validation webhooks

fedepaol:
Adding validating webhooks to ensure the configuration is set properly.

Fixes https://github.com/openperouter/openperouter/issues/21


internal/webhooks/vni_webhook.go
```diff
@@ -0,0 +1,143 @@
...
+			if err := v.decoder.DecodeRaw(req.OldObject, &oldVNI); err != nil {
+				return admission.Errored(http.StatusBadRequest, err)
+			}
+		}
+	}
+
+	var warnings []string
+	switch req.Operation {
+	case v1.Create:
+		w, err := validateVNI(&vni)
```
oribon, line 69: should be validateVNICreate

internal/webhooks/vni_webhook.go
```diff
@@ -0,0 +1,143 @@
...
+		if err != nil {
+			return admission.Denied(err.Error())
+		}
+		warnings = w
+	case v1.Update:
+		w, err := validateVNI(&vni)
```
oribon, line 75: should be validateVNIUpdate

internal/webhooks/mock_validator_test.go
oribon, line 1: I think this is not used?

internal/conversion/validate_vni.go
```diff
@@ -23,6 +23,10 @@ func ValidateVNIs(vnis []v1alpha1.VNI) error {
 
 	for _, vni := range vnis {
 		vrfName := vni.VRFName()
+		if _, _, err := net.ParseCIDR(vni.Spec.LocalCIDR); err != nil {
```
oribon, line 26: can you add a case for this?

e2etests/tests/webhooks.go
```diff
@@ -0,0 +1,188 @@
...
+					Name:      "underlay",
+					Namespace: openperouter.Namespace,
+				},
+				Spec: v1alpha1.UnderlaySpec{
+					ASN:      65000,
+					Nics:     []string{"nic1", "nic2"},
+					VTEPCIDR: "192.168.1.0/24",
+				},
+			}, "can only have one nic"),
+			Entry("when trying to create an underlay with multiple nics", v1alpha1.Underlay{
```
oribon, line 111: nit: with an invalid vtep CIDR

cmd/nodemarker/main.go
```diff
@@ -114,7 +129,7 @@ func main() {
 	})
 
 	startListeners := make(chan struct{})
-	if args.enableWebhook && !args.disableCertRotation {
+	if !args.disableCertRotation {
```
oribon, line 132: I guess everything works because of the hardcoded `--disable-cert-rotation` in helm, but can we still keep checking that the webhooks should run (`!webhookDisabled` or `enabled || only`)?

operator/config/olm-install/install-resources.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
   namespace: openperouter-system
 spec:
   displayName: OpenPERouter Operator
-  image: quay.io/openperouter/openperouter-operator-bundle-index:main
+  image: quay.io/openperouter/openperouter-operator-catalog:v0.0.0
```
oribon, line 13: I think this should not be commited

PR #93, opened 2025-07-23
Enable BFD for the underlay session

fedepaol:
Allowing optional BFD enablement so we can have BFD for the undelay session.


e2etests/pkg/frr/bgp.go
```diff
@@ -110,29 +110,15 @@ type FRRRoute struct {
 	} `json:"nexthops"`
 }
 
+// BFDPeer represents a BFD peer.
 type BFDPeer struct {
```
oribon, line 114: why zap the other fields? some linter complained?

e2etests/tests/sessions.go
```diff
@@ -141,3 +144,142 @@ var _ = Describe("Router Host configuration", Ordered, func() {
...
+			for _, node := range nodes {
+				neighborIP, err := infra.NeighborIP(infra.KindLeaf, node.Name)
+				Expect(err).NotTo(HaveOccurred())
+				validateSessionWithNeighbor(infra.KindLeaf, node.Name, exec, neighborIP, Established)
+			}
+
+			if underlay.Spec.Neighbors[0].BFD != nil && underlay.Spec.Neighbors[0].BFD.TransmitInterval != nil {
+				By("validating BFD parameters are propagated to the router")
+				for _, pod := range routerPods {
+					podExec := executor.ForPod(pod.Namespace, pod.Name, "frr")
```
oribon, line 228: why does this check the config on podExec instead of checking it was accepted on the ToR (using the fields zapped from the struct)?

e2etests/tests/validate_session.go
```diff
@@ -57,7 +57,7 @@ func validateSessionWithNeighbor(fromName, toName string, exec executor.Executor
 	}, 5*time.Minute, time.Second).ShouldNot(HaveOccurred())
 }
 
-func validateNoSuchNeigh(exec executor.Executor, neighborIP string) {
+func validateNoSessionForNeigh(exec executor.Executor, neighborIP string) {
```
oribon, line 60: nit: validateNo**Established**SessionForNeigh ?

e2etests/pkg/frr/bgp.go
fedepaol: Don't remember, probably we weren't using it and I found it confusing.

internal/conversion/frr_conversion.go
```diff
@@ -148,18 +152,39 @@ func neighborToFRR(n v1alpha1.Neighbor) (*frr.NeighborConfig, error) {
...
+	bfdProfile := &frr.BFDProfile{
+		Name:             profileName,
+		ReceiveInterval:  n.BFD.ReceiveInterval,
+		TransmitInterval: n.BFD.TransmitInterval,
+		DetectMultiplier: n.BFD.DetectMultiplier,
+		EchoInterval:     n.BFD.EchoInterval,
+		EchoMode:         ptr.Deref(n.BFD.EchoMode, false),
+		PassiveMode:      ptr.Deref(n.BFD.PassiveMode, false),
+		MinimumTTL:       n.BFD.MinimumTTL,
+	}
```
oribon, line 185: this is a bit odd (returning the profile as a side-effect as this function), can we modify the `NeighborConfig` struct to have the type of its BFDProfile field as `BFDProfile` instead of `string`, and populate it here instead (with the cost of modifying the frr tmpls)?
alternatively, we can just move this to a different function 😅

e2etests/tests/validate_session.go
oribon: can we change this? NoSession is a bit vague, where we want specifically no session in "Established"😅

internal/conversion/frr_conversion.go
fedepaol: that was my original version (embedding it), but then I realized the profile is not related to the neighbor, it's a global thing in frr.
I will move to a separate function.

PR #109, opened 2025-08-12
Track per-node status of Underlay configurations

phoracek:

  /kind feature

  This PR introduces the UnderlayNodeStatus CRD and its lifecycle management to provide per-node
  tracking of Underlay configurations.

  Key components added:
  - **UnderlayNodeStatus CRD**: New custom resource with spec (nodeName, underlayName) and status
  (lastReconciled) fields
  - **Controller logic**: Automatically creates and manages UnderlayNodeStatus resources for each
  Underlay on each node. Cleanups are handled through owner references.
  - **RBAC permissions**: Required cluster role permissions for UnderlayNodeStatus operations

  Example UnderlayNodeStatus YAML:

```
  kind: UnderlayNodeStatus
  metadata:
    name: my-underlay.worker-1
  spec:
    nodeName: worker-1
    underlayName: my-underlay
  status:
    lastReconciled: "2024-01-01T00:00:00Z"
```

  Example kubectl output:

```
  $ kubectl -n openperouter-system get underlaynodestatus
  NAME                    NODE       UNDERLAY      AGE
  my-underlay.worker-1    worker-1   my-underlay   2m
  my-underlay.worker-2    worker-2   my-underlay   2m
  my-underlay.control-1   control-1  my-underlay   2m
```

A follow-up PR will extend this CRD with status of NIC configuration.

Let me know if you'd prefer the 3 tests to be merged into a single one, verifying the whole lifecycle at once.

  **Release note**:

```release-note
Add UnderlayNodeStatus CRD for per-node status tracking of Underlay configurations.
```

### DIFF
--- a/charts/openperouter/templates/rbac.yaml
+++ b/charts/openperouter/templates/rbac.yaml
@@ -133,6 +133,32 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - openpe.openperouter.github.io
+  resources:
+  - underlaynodestatuses
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - openpe.openperouter.github.io
+  resources:
+  - underlaynodestatuses/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - openpe.openperouter.github.io
+  resources:
+  - underlaynodestatuses/status
+  verbs:
+  - get
+  - patch
+  - update
 {{- if .Values.webhook.enabled }}
 - apiGroups:
   - admissionregistration.k8s.io
--- a/config/all-in-one/crio.yaml
+++ b/config/all-in-one/crio.yaml
@@ -649,6 +649,32 @@ rules:
 - apiGroups:
   - openpe.openperouter.github.io
   resources:
+  - underlaynodestatuses
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - openpe.openperouter.github.io
+  resources:
+  - underlaynodestatuses/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - openpe.openperouter.github.io
+  resources:
+  - underlaynodestatuses/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - openpe.openperouter.github.io
+  resources:
   - underlays
   verbs:
   - create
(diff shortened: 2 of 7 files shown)


phoracek:
> > @maiqueb in no particular order, this was my motivation:
> > ```
> > * Prior work: This pattern is following what we already do in kubernetes-nmstate (node network configuration enactment) and frr-k8s (frr node state)
> > ```
> 
> But in kubernetes-nmstate you do not have a control plane component that is not a daemonset right ? i.e. you'd have contention to write to this common resource. Here, we do have one.
> 
> > ```
> > * Scale: If we kept the status as a part of `Underlay`, we would have N (for number of nodes) controllers trying to write to the same status, which may (?) cause issues. Perhaps Kubernetes API server would handle it fine though, I haven't tested it.
> > ```
> 
> Why would we have N controllers ? openperouter's architecture has 3 components:
> 
>     * router (daemonset)
> 
>     * controller (I really dislike the name ...) - also a daemonset
> 
>     * node labeler - this is a deployment
> 
> 
> We could move this sync operation to the "node labeler" component, but ... I guess the name doesn't fit.
> 
> FWIW, I also don't think having the "controller" component do something like moving interfaces to different namespaces is a fit.

It is the controller daemon set that reads the Underlay resource and moves the NIC from host to the router netns (among other things).

That means this is the component that has the information about success / failure of the NIC configuration (this is what I want to get in the status in the following PR). So I want this component to announce the configuration result somehow.

It also means that if this result is reported from controller, writing it to the single Underlay resource may cause congestion.

I hope this answers the questions above.

> 
> > ```
> > * Scale: Another issue may be that if we have lot of nodes, the information may not fit into a single obejct. But here again I don't have anything to back that up with.
> > ```
> 
> Right. Plus ... what's a lot of nodes ?... like 300 ? 500 ? I don't think clusters w/ 10 K nodes actually exist in the wild. At least kubernetes states that 5K nodes is the limit in https://kubernetes.io/docs/setup/best-practices/cluster-large/


Let's say we have 500 nodes. Per each we want to report status of the underlay configuration - VXLAN config, NICs, ... That would make for one unwieldy CR IMO.

> 
> > ```
> > * UX: Personally I find it easier to look for the status in the high-level list of UnderlayNodeStatuses to understand in a glance what's wrong, than to dive into a single Underlay and try parsing the status.
> > ```
> 
> Maybe.
> 
> OK, I guess my arguments are not exactly amazing. I'll check the code now.

Thanks!

phoracek:
Makes sense. I will prioritize writing the enhancement doc, and in the meanwhile continue on the interface status reporting as a PoC.

phoracek:
@fedepaol @maiqueb in your own time https://github.com/openperouter/openperouter/pull/111, thanks!

PR #102, opened 2025-07-30
L3VNI: Make the local session optional

fedepaol:
**Is this a BUG FIX or a FEATURE ?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**What this PR does / why we need it**:

When using a l3vni as a l3 wrapper of a l2vni, the session with the host
is not necessary.

Here we make the session related fields optional and wrapped under a
subfield.

**Special notes for your reviewer**:

**Release note**:

```release-note
Optional hostsession in the L3VNI CRD, now it's not mandatory to setup a session if a l3vni serves as L3 wrapper of a L2 VNI.
```

PR #113, opened 2025-08-31
Have a different routerid per node

fedepaol:

>
> /kind bug
> /kind cleanup
kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

We should have a different routerid per node. Normally FRR picks the lowest ipv4 of the interfaces of the nodes,
but it makes sense to be deterministic and know which node is which (also, it's gonna be useful for route distinguishers).

Built on top of https://github.com/openperouter/openperouter/pull/102

**Release note**:

```release-note
Set explicitly the routerid of each node, defaulting to use the 10.0.0.0/24 CIDR.
```

### DIFF
--- a/api/v1alpha1/underlay_types.go
+++ b/api/v1alpha1/underlay_types.go
@@ -33,6 +33,11 @@ type UnderlaySpec struct {
 	// +required
 	VTEPCIDR string `json:"vtepcidr,omitempty"`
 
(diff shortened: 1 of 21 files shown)


fedepaol:
@maiqueb mind taking a look?

maiqueb (approved):
The code looks good, though I do not understand why there isn't a routerid attribute for L2 VNIs. 

They still need route distinguishers ... Don't we want this "control" there as well ?

PR #122, opened 2025-10-06
Add a knob to run the pods on master nodes

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Currently both the router and the controller pod can run on masters. The nodemarker can't.

This make it unsuitable to run on single node clusters. Because of this, we:

- allow the nodemarker deployment from kustomize to run on master nodes
  too
- introduce a new knob in the helm chart and in the operator to
  selectively choose if the pods should be allowed to run on master
  nodes or no (default: yes)

**Release note**:

```release-note
Allow pods to run on master nodes or not.
```

### DIFF
--- a/operator/api/v1alpha1/openperouter_types.go
+++ b/operator/api/v1alpha1/openperouter_types.go
@@ -42,6 +42,10 @@ type OpenPERouterSpec struct {
 	// MultusNetworkAnnotation specifies the Multus network annotation to be added to the router pod.
 	// +optional
 	MultusNetworkAnnotation string `json:"multusNetworkAnnotation,omitempty"`
+	// RunOnMaster determines if all pods (router, controller, and nodemarker) will run on master/control-plane nodes. (default: true)
+	// +optional
+	// +kubebuilder:default:=true
+	RunOnMaster bool `json:"runOnMaster,omitempty"`
 }
 
 // OpenPERouterStatus defines the observed state of OpenPERouter
--- a/operator/bundle/manifests/openpe.openperouter.github.io_openperouters.yaml
+++ b/operator/bundle/manifests/openpe.openperouter.github.io_openperouters.yaml
@@ -55,6 +55,12 @@ spec:
                 description: MultusNetworkAnnotation specifies the Multus network
                   annotation to be added to the router pod.
                 type: string
+              runOnMaster:
+                default: true
+                description: 'RunOnMaster determines if all pods (router, controller,
+                  and nodemarker) will run on master/control-plane nodes. (default:
+                  true)'
+                type: boolean
             type: object
           status:
             description: OpenPERouterStatus defines the observed state of OpenPERouter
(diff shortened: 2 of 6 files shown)


fedepaol:
cc @zeeke

charts/openperouter/templates/nodemarker.yaml
```diff
@@ -34,6 +34,13 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ template "openperouter.controller.serviceAccountName" . }}
+      tolerations:
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/master
+        operator: Exists
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/control-plane
+        operator: Exists
```
maiqueb: should we have this configurable ? 

i.e. have the nodemarker deployed on the control plane nodes based on some configuration ?

fedepaol:
@maiqueb mind having another look?

charts/openperouter/values.yaml
```diff
@@ -13,6 +13,10 @@ openperouter:
   # -- Controller log level. Must be one of: `debug`, `info`, `warn` or `error`.
   logLevel: info
   tolerateMaster: true
+  # -- If true, router and controller pods will run on master/control-plane nodes
+  routerOnMaster: true
+  # -- If true, nodemarker pod will run on master/control-plane nodes
+  nodeMarkerOnMaster: true
```
maiqueb: Why should we have two different knobs ? While I understand we have multiple components in the system, IIUC, the controller / router components are daemonsets, thus scheduled in all workers nodes in the system. One of these knobs will allow these to be deployed in control-plane nodes.

The node labeler component, on the other hand, is a deployment which annotates the nodes. It is a deployment, running in a configurable subset of nodes. 

Thus, I think the **only** "knob" we actually require is a knob allowing single cluster installation, which would add the toleration to the control plane pods for both labeler, and controller.

maiqueb (commented):
Some questions.

charts/openperouter/values.yaml
fedepaol: my reasoning was:

- a user might want or not to enhance his master nodes with what this router provides. For this, he might need to run the router / controller on masters too. For some scenarios, running on worker might be enough
- the nodemarker is a utility tool, and there's no value in running it on masters UNLESS you run a single node cluster

With one single knob, a user who wants to run the router on masters would have to take also the nodemarker running on masters even if he doesn't care about that.
If you think this is too much and the configuration becomes too complex, I can step back.

maiqueb: I think it all boils down to if we envision people running the controller on control plane pods (what actually does the heavy lifting) but not wanting under any circumstance to have the nodemarker on the control plane. 

I don't see that happening, but maybe I'm short sighted. I would go with a single knob, `schedulableControlPlaneNodes` or something similar. 

Waiting for more people feedback. Having the two knobs is not a deal breaker for me.

fedepaol: makes sense, we have room for making this more complex later. I'll switch to one single knob

PR #135, opened 2025-10-15
api: make L3VNI VRF field mandatory

### LINKED ISSUE #134: Make the vrf field of l3vni mandatory
Using the name of l3vni as the vrf is useful when the user knows it's gonna be used for that, otherwise we might have a l3vni with a name that is not good to be used as vrf name.

For this reason, I think it'd be better to avoid confusing users and always rely on the vrf field of l3vni.

qinqon:
/kind feature
Fixes #134 

Previously, the VRF field in L3VNISpec was optional with a pointer type, falling back to the VNI instance name when not set. This commit makes the VRF field required to enforce explicit VRF naming.

This ensures users explicitly specify VRF names for L3VNI resources rather than relying on implicit naming behavior.

The api is changing the vrf type to a non pointer string, not sure if this can be a problem since I think the API is alpha.

**Release note**:
```release-note
api: make L3VNI VRF field mandatory
```


maiqueb (approved):
Thank you. I think this covers it.

fedepaol:
@qinqon one last thing. Please go over the examples and the website to validate we don't have broken configuration there

e2etests/tests/webhooks.go
```diff
@@ -63,13 +63,30 @@ var _ = Describe("Webhooks", func() {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring(expectedError))
 			},
+			Entry("when trying to create a VNI without VRF field", v1alpha1.L3VNI{
```
maiqueb: please discriminate the VNI as L3.

examples/examples_suite_test.go
```diff
@@ -0,0 +1,108 @@
...
+		CRDDirectoryPaths:     []string{filepath.Join("..", "config", "crd", "bases")},
+		ErrorIfCRDPathMissing: true,
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join("..", "bin", "k8s",
+			fmt.Sprintf("1.31.0-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
maiqueb: is this the k8s version we're using ? 1.31 ? 

What is this number ? Should it be exposed somewhere else ?

website/content/docs/configuration/evpn.md
```diff
@@ -62,23 +62,25 @@ metadata:
...
 | Field | Type | Description | Required |
 |-------|------|-------------|----------|
-| `asn` | integer | Router ASN for BGP session with host | Yes |
+| `vrf` | string | Name of the VRF (Virtual Routing and Forwarding) instance | Yes |
 | `vni` | integer | Virtual Network Identifier (1-16777215) | Yes |
-| `localcidr` | string | CIDR for veth pair IP allocation | Yes |
-| `hostasn` | integer | Host ASN for BGP session | Yes |
+| `hostsession.asn` | integer | Router ASN for BGP session with host | Yes |
+| `hostsession.hostasn` | integer | Host ASN for BGP session | Yes |
+| `hostsession.localcidr` | string | CIDR for veth pair IP allocation | Yes |
```
maiqueb: This has nothing to do with this PR afaict. 

I would have it in a separate PR.

PR #148, opened 2025-10-23
Fixed #147 Flag `--metrics-bind-address` being ignored on controller and nodemarker

### LINKED ISSUE #147: flag `--metrics-bind-address` is a no-op on controller/nodemarker binaries
Using `--metrics-bind-address` doesn't have any impact when running the controller or nodemarker, it is simply ignored.

Spotted the issue on the controller because it runs on hostNetwork and I had a port conflict with another service running on hostNetwork as well using the same TCP port 8080.

```{"time":"2025-10-23T14:20:10.181694112Z","level":"ERROR","msg":"problem running manager","logger":"setup","err":"failed to start metrics server: failed to create listener: listen tcp :8080: bind: address already in use"}```

I tried to manually patch the daemonset by adding the flag `--metrics-bind-address 9080` to the container config, but noticed it didn't change anything and the controller was still trying to bind on :8080. Indeed, it appears that the `metricsAddr` var is not used at all on both binaries (controller & nodemarker). This is less impacting for the nodemarker since it doesn't run on hostNetwork, but in its own network namespace.

By default the controller-runtime starts a metrics server on port 8080, it could be overriden when creating the manager with
```
ctrl.NewManager(..., options: ctrl.Options{
    (...)
    Metrics: metricsserver.Options{
		BindAddress:   args.metricsAddr,
    },
})
```

Also, it would be convenient to be able to tweak the values of these ports directly from the Helm chart. I would be happy to contribute if accepted.

### PR DESCRIPTION
**Pull request by fdomain**

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

This PR makes the flag `--metrics-bind-address` being really interpreted when used on the controller or nodemarker binaries.

Fixes #147 

**Release note**:

```release-note
Fix flag `--metrics-bind-address` being ignored on controller and nodemarker binaries
```

### DIFF
--- a/cmd/hostcontroller/main.go
+++ b/cmd/hostcontroller/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


fdomain:
@fedepaol are you okay with the current default values for the `--metrics-bind-addr` flag (hostcontroller => 0 so disabled, nodemarker => :8080) ?

Issue #22, opened 2025-02-28
Implement node selectors for VNIs and Underlays


fedepaol:
Different nodes might connect to different peers, so the configuration should be per node.

qinqon:
@fedepaol we are going to prepare a design for this.

fedepaol:
https://github.com/openperouter/openperouter/tree/main/enhancements

PR #152, opened 2025-10-28
enhancements: Add per-node configuration enhancement proposal

qinqon:

/kind design

Introduce enhancement proposal for adding node selector support to OpenPERouter's resource. This enables different nodes to connect to different BGP neighbors and use node-specific network configurations, supporting heterogeneous cluster topologies.

It's a design for issue:
- https://github.com/openperouter/openperouter/issues/22

Asisted by claude code, the examples are kind of out of the blue but kind of make sense, the rest was properly reviewed add changed by the author.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/enhancements/per-node-configuration.md
+++ b/enhancements/per-node-configuration.md
@@ -1,0 +1,584 @@
+# Per-Node Configuration with Node Selectors
+
+## Summary
+
+This enhancement proposes adding node selector support to all OpenPERouter CRDs (`Underlay`, `L3VNI`, `L2VNI`, and `L3Passthrough`), enabling different nodes to have different network configurations. This allows OpenPERouter to support heterogeneous cluster topologies including multi-datacenter, multi-rack, and mixed-hardware environments where different nodes may need different BGP configurations, VNI assignments, or passthrough settings.
+
+## Motivation
+
+Currently, OpenPERouter applies all CRD configurations cluster-wide, meaning all nodes in the cluster receive the same configuration. This creates operational limitations in real-world deployments:
+
+**Underlay limitations:**
+- **Multi-rack deployments**: Different nodes are physically connected to different ToR (Top of Rack) switches but cannot be configured to peer with their respective ToRs
+- **Multi-datacenter clusters**: Nodes distributed across availability zones or datacenters need to peer with location-specific BGP routers
+- **Hardware heterogeneity**: Different server models have different NIC naming conventions (e.g., Dell uses `eno1`, HP uses `em1`), preventing a one-size-fits-all configuration
+- **Topology flexibility**: Cannot accommodate nodes in different physical network topologies within the same cluster
+
+**L3VNI/L2VNI limitations:**
+- **Per-rack VNI isolation**: Different racks may need separate VNI configurations for network segmentation
+- **Zone-specific VNIs**: Nodes in different availability zones may require different VNI configurations for compliance or performance reasons
+- **Selective VNI deployment**: Some workloads on specific nodes may need access to certain VNIs while others don't
+- **Gateway placement**: L2 gateway IPs may need to be configured only on specific nodes based on network topology
+
+**L3Passthrough limitations:**
+- **Selective passthrough**: Only specific nodes should participate in direct BGP fabric communication
+- **Security zones**: Different security zones may require different passthrough configurations
+- **Specialized workloads**: Only nodes running certain workloads should have passthrough capabilities
+
+### Goals
+
+- Enable per-node configuration for all OpenPERouter CRDs using Kubernetes node selectors
+- Support heterogeneous network configurations across different hardware platforms, racks, and zones
+- Allow multiple instances of each CRD to coexist, each targeting specific node subsets
+- Maintain backward compatibility with existing cluster-wide configurations
+- Prevent configuration conflicts through validation
+
+### Non-Goals
+
+- Dynamic node selector updates based on workload scheduling (this is handled by node labels)
+- Handle external events unrelated to node labels, like interfaces names and similar ones
+- Support multiple underlays per node
+
+## Proposal
+
+### User Stories
+
+**Underlay stories:**
+- **As a cluster administrator**, I want each rack's nodes to peer with their local ToR switch so that network traffic stays within the rack when possible and I can configure rack-specific ASNs.
+- **As a cluster administrator**, I want nodes in different datacenters to connect to datacenter-local BGP routers so that cross-datacenter BGP traffic is minimized and datacenter-specific network policies can be enforced.
+- **As a cluster administrator**, I want to configure different NIC names for different server vendors so that I can use vendor-specific interface naming without requiring identical hardware across the cluster.
+
+**L3VNI/L2VNI stories:**
+- **As a cluster administrator**, I want to configure different VNIs for different racks so that each rack has isolated network segments for multi-tenancy.
+- **As a cluster administrator**, I want nodes in specific zones to use zone-specific VNI configurations so that compliance requirements for data locality are met.
+- **As a cluster administrator**, I want worker nodes (not control plane nodes) to have VNIs configured so that control plane traffic remains isolated.
+
+**L3Passthrough stories:**
+- **As a cluster administrator**, I want only edge nodes to have L3Passthrough configured so that only designated nodes participate in direct BGP fabric communication.
+- **As a cluster administrator**, I want different security zones to have different passthrough configurations so that traffic policies can be enforced based on node location.
+
+## Design Details
+
+### API Changes
+
+Add an optional `NodeSelector` field to all CRD Spec structures. This field will be identical across all CRDs to maintain consistency.
+
+#### Underlay
+
+```go
+type UnderlaySpec struct {
+    // NodeSelector specifies which nodes this Underlay applies to.
+    // If empty or not specified, applies to all nodes (backward compatible).
+    // Multiple Underlays with overlapping node selectors will be rejected.
+    // +optional
+    NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
+
+    // ... rest of fields
+}
+```
+
+#### L3VNI
+
+```go
+type L3VNISpec struct {
+    // NodeSelector specifies which nodes this L3VNI applies to.
+    // If empty or not specified, applies to all nodes (backward compatible).
+    // Multiple L3VNIs can match the same node (unlike Underlay).
+    // +optional
+    NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
+
+    // ... rest of fields
+}
+```
+
+#### L2VNI
+
+```go
+type L2VNISpec struct {
+    // NodeSelector specifies which nodes this L2VNI applies to.
+    // If empty or not specified, applies to all nodes (backward compatible).
+    // Multiple L2VNIs can match the same node (unlike Underlay).
+    // +optional
+    NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
+
+    // ... rest of fields
+}
+```
+
+#### L3Passthrough
+
+```go
+type L3PassthroughSpec struct {
+    // NodeSelector specifies which nodes this L3Passthrough applies to.
+    // If empty or not specified, applies to all nodes (backward compatible).
+    // Multiple L3Passthroughs can match the same node (unlike Underlay).
+    // +optional
+    NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
+
+    // ... rest of fields
+}
+```
+
+### Example Configurations
+
+#### Underlay: Multi-Rack Configuration
+
+Different racks connect to different ToR switches:
+
+```yaml
+# Rack 1 nodes connect to ToR 1
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-rack-1
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchLabels:
+      topology.kubernetes.io/rack: rack-1
+  asn: 64512
+  evpn:
+    vtepcidr: 100.65.1.0/24
+  nics:
+    - toswitch
+  neighbors:
+    - asn: 64500
+      address: 192.168.1.254  # ToR switch for rack 1
+      bfd:
+        receiveInterval: 300
+        transmitInterval: 300
+---
+# Rack 2 nodes connect to ToR 2
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-rack-2
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchLabels:
+      topology.kubernetes.io/rack: rack-2
+  asn: 64512
+  evpn:
+    vtepcidr: 100.65.2.0/24
+  nics:
+    - toswitch
+  neighbors:
+    - asn: 64500
+      address: 192.168.2.254  # ToR switch for rack 2
+      bfd:
+        receiveInterval: 300
+        transmitInterval: 300
+```
+
+#### Underlay: Multi-Datacenter Configuration
+
+Different datacenters with different ASNs:
+
+```yaml
+# Underlay for datacenter-east nodes
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-dc-east
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchLabels:
+      topology.kubernetes.io/zone: us-east-1a
+  asn: 64512
+  evpn:
+    vtepcidr: 100.65.0.0/24
+  nics:
+    - eth1
+  neighbors:
+    - asn: 64500
+      address: 192.168.10.1  # ToR switch in DC East
+---
+# Underlay for datacenter-west nodes
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-dc-west
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchLabels:
+      topology.kubernetes.io/zone: us-west-1a
+  asn: 64513
+  evpn:
+    vtepcidr: 100.66.0.0/24
+  nics:
+    - eth1
+  neighbors:
+    - asn: 64501
+      address: 192.168.20.1  # ToR switch in DC West
+```
+
+#### Underlay: Hardware-Specific NIC Configuration
+
+Different NIC naming across vendor hardware:
+
+```yaml
+# For Dell servers with specific NIC naming
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-dell-hardware
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchLabels:
+      hardware.vendor: dell
+  asn: 64512
+  evpn:
+    vtepcidr: 100.65.0.0/24
+  nics:
+    - eno1
+    - eno2
+  neighbors:
+    - asn: 64500
+      address: 192.168.10.1
+---
+# For HP servers with different NIC naming
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-hp-hardware
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchLabels:
+      hardware.vendor: hp
+  asn: 64512
+  evpn:
+    vtepcidr: 100.65.0.0/24
+  nics:
+    - em1
+    - em2
+  neighbors:
+    - asn: 64500
+      address: 192.168.10.1
+```
+
+#### L3VNI: Per-Rack VNI Configuration
+
+Different racks use different L3VNIs for network segmentation:
+
+```yaml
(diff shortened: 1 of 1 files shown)


enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,484 @@
...
+  - Multi-rack deployments
+  - Multi-datacenter deployments
+  - Mixed hardware deployments
+- Migration guide from cluster-wide to per-node configuration
+- Troubleshooting guide for selector conflicts
+
+### Phase 5: E2E Testing
+
+**Deliverables:**
+- E2E test for multi-rack scenario
```
maiqueb: what does this look like ? a topology diagram would be incredible, but I just need to understand what does this mean at this point.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,484 @@
...
+- E2E test for mixed hardware scenario
+- E2E test for dynamic node labeling
+- E2E test for validation webhook
+- Performance testing with multiple Underlays and many nodes
```
maiqueb: really ? Are we actually going through with this ? 

Sounds a bit more than what we usually do.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,484 @@
...
+
+- Enable per-node BGP neighbor configuration using Kubernetes node selectors
+- Support heterogeneous NIC configurations across different hardware platforms
+- Allow multiple `Underlay` resources to coexist, each targeting specific node subsets
+- Maintain backward compatibility with existing cluster-wide configurations
+- Prevent configuration conflicts through validation
+
+### Non-Goals
+
+- Per-neighbor node selectors (Option 2 - less flexible, documented as alternative)
```
maiqueb: It might be me and my thick head, but I am finding it very easy to mix this non-goal with the goal above of "Enable per-node BGP neighbor configuration". Could you disambiguate these somehow ? Maybe just adjust the wording.

FWIW, it seems that you're specifically listing one thing that is proposed in one alternative as a non-goal - which AFAIU automatically discards said alternative. 

Just trying to figure out the scope here.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,484 @@
...
+    - toswitch
+  neighbors:
+    - asn: 64500
+      address: 192.168.2.254  # ToR switch for rack 2
+      bfd:
+        receiveInterval: 300
+        transmitInterval: 300
+```
+
+#### Multi-Datacenter Configuration
```
maiqueb: how are these any different from the `Multi-Rack Configuration` ? I see you've dropped the bfd configurations - OK, might make sense - but, afaiu, the "big" difference is the labels used - which can be anything, as long as it matches the labels on the nodes.

Just making sure I get the API.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,484 @@
...
+      address: 192.168.10.1
+```
+
+### Controller Implementation
+
+#### Node Matching Logic
+
+The Underlay controller will be enhanced with:
+
+1. **Node Label Watching**: Add a watch for Node resource label changes to trigger reconciliation
```
maiqueb: we can pinpoint the watcher to these specific bits, right ?

enhancements/per-node-configuration.md
qinqon: I suppose a pair of Underlays for different nodes with different neighbors ASNs,

enhancements/per-node-configuration.md
qinqon: > It might be me and my thick head, but I am finding it very easy to mix this non-goal with the goal above of "Enable per-node BGP neighbor configuration". Could you disambiguate these somehow ? Maybe just adjust the wording.
> 
> FWIW, it seems that you're specifically listing one thing that is proposed in one alternative as a non-goal - which AFAIU automatically discards said alternative. 
> 
> Just trying to figure out the scope here.

I will rephrase the non goals

enhancements/per-node-configuration.md
qinqon: > we can pinpoint the watcher to these specific bits, right ?

What do you mean ?

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,484 @@
...
+
+    // Conditions represent the latest available observations of the Underlay's state
+    // +optional
+    Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+```
+
+**Condition Types:**
+- `Ready`: All matched nodes have configuration successfully applied
+- `NodesMatched`: At least one node matches the selector
```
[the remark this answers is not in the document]
qinqon: > maybe I'm being a bit short-sighted, but would you explain why we want to differentiate between these two ? 
> 
> i.e. couldn't "no matching nodes found" be one of the "messages" / reasons of  condition `Ready: false` ?

To have the "columnt" view with `-o wide` we usually do 

```
// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.status==\\"True\\")].type",description="Status"
```
If we do so we are not propertly has the Status "kubectl -o wide" column since we have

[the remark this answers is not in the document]
qinqon: > maybe I'm being a bit short-sighted, but would you explain why we want to differentiate between these two ? 
> 
> i.e. couldn't "no matching nodes found" be one of the "messages" / reasons of  condition `Ready: false` ?

To have the "columnt" view with `-o wide` we usually do:
```
// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.conditions[?(@.status==\\"True\\")].type",description="Status"
```
This will print something like
```
Status
Ready
```
If we just convert this to a reason we are not going to be able to distinguis between not ready or in progress, we should add a columnt to have something like 
```
Status Nodes
```

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,484 @@
...
+  conditions:
+    - type: NodesMatched
+      status: "True"
+      reason: NodesFound
+      message: "3 nodes match the selector"
+      lastTransitionTime: "2025-01-15T10:30:00Z"
+    - type: Ready
+      status: "True"
+      reason: ConfigurationSuccessful
+      message: "Configuration applied to all 3 matched nodes"
```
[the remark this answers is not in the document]
maiqueb: I don't follow. Could you explain how that would look like ? Maybe consider adding that permutation to the enhancement.

qinqon:
New version [pushed](https://github.com/openperouter/openperouter/compare/2b0e196f6257cd192758d621e7bc85316a8dedc8..e60051be08509666b3840e1eea90e5f2c3d15bec): 

 The enhancement proposal was expanded from focusing solely on Underlay configuration to supporting all OpenPERouter CRDs. Key updates include:

  Scope expansion:
  - Originally proposed node selectors for Underlay only
  - Now includes L3VNI, L2VNI, and L3Passthrough CRDs

  New capabilities:
  - L3VNI: Per-rack VNI isolation, zone-specific VNIs, selective VNI deployment
  - L2VNI: Per-rack gateway configuration, selective deployment on specific node types
  - L3Passthrough: Edge-only deployment, per-security-zone configurations

  Conflict resolution updates:
  - Underlay: Only one instance per node (validation enforced)
  - L3VNI/L2VNI/L3Passthrough: Multiple instances allowed on the same node

  New example configurations added:
  - 6 new L3VNI examples (per-rack, zone-specific, multiple instances)
  - 4 new L2VNI examples (selective deployment, per-rack gateway)
  - 4 new L3Passthrough examples (edge nodes, security zones)
  - Example showing multiple instances of different CRD types on the same node

  Implementation scope changes:
  - All controllers now need node selector support (not just Underlay)
  - Validation webhook only enforces uniqueness for Underlay
  - Status reporting standardized across all CRD types
  - Implementation phases updated to cover all four CRD types

qinqon:
Last [push](https://github.com/openperouter/openperouter/compare/e60051be08509666b3840e1eea90e5f2c3d15bec..dc50a8541c2533fdd7def55e5e1ee5fd494d4521):

- New status fields in RouterNodeConfigurationStatus:
    - matchedResources: Lists configuration resources that matched the node (kind + name)
    - Condition type renamed: NodesMatched → NodeMatched

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,484 @@
...
+  nodeSelector:
+    matchLabels:
+      topology.kubernetes.io/rack: rack-1
+  asn: 64512
+  # ... rest of spec
+status:
+  matchedNodes:
+    - worker-1
+    - worker-2
+    - worker-3
```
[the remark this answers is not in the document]
qinqon: I have refactor the not matching condition mechanism with the status enhancement bits, so this field no longer applies.

qinqon:
> Plus, I would like to see a short disclaimer indicating which of the phases are required, and which are optional; from what I understand, we can only implement - assuming there are no dependencies to other enhancements - phases 1, 2, 4, and 5. Which to me, are all required for this "feature" to be complete. Thus, I think we should have 2 phases: the first which is about API (current phase1), controller impl (current phase2), docs (current phase4), and e2e testing (current phase5), and a second stage, where status reporting is implemented, along with doc updates, plus e2e testing that relies on the newly added status.

That make sense, we can implement first everything except status so we don't depend on the status implementation and later on we we implement the status bits.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,850 @@
...
+    matchLabels:
+      security-zone: internal
+  hostsession:
+    asn: 64512
+    hostasn: 64720
+```
+
+#### Multiple Instances on Same Node
+
+Example showing multiple L3VNIs configured on the same set of nodes for different tenants:
```
fedepaol: is this adding value?

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,850 @@
...
+
+Only one Underlay can match a given node at any time. The validation webhook will prevent conflicts at admission time:
+
+1. List all existing Underlay resources
+2. For each new/updated Underlay, check if any nodes would match multiple Underlays
+3. Reject the operation if overlapping selectors are detected
+
+**L3VNI, L2VNI, and L3Passthrough:**
+
+Multiple instances of L3VNI, L2VNI, and L3Passthrough can match the same node. This is intentional and allows for:
```
fedepaol: this is true, but is not about conflict resolution. 
We need to ensure we don't have incompatible *vnis and passthrough on the same node (for each node)

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,850 @@
...
+4. **Dynamic Updates**: Reconcile when:
+   - Resources are created/updated/deleted
+   - Node labels change
+   - Nodes are added/removed from the cluster
+
+#### Conflict Resolution
+
+**Underlay-specific restriction:**
+
+Only one Underlay can match a given node at any time. The validation webhook will prevent conflicts at admission time:
```
fedepaol: the validation must happen also in the reconcile loop (and we have a pattern to do that). This is because webhooks might let configurations slip in in case of parallel creation of resources.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,850 @@
...
+
+### Validation Rules
+
+The validation webhook enforces:
+
+1. **No Overlapping Selectors for Underlay**: Multiple Underlays cannot match the same node
+2. **Selector Validity**: Node selector must be a valid `metav1.LabelSelector` for all CRD types
+3. **Multiple instances allowed**: L3VNI, L2VNI, and L3Passthrough can have multiple instances matching the same node (no overlap validation)
+
+### RBAC Requirements
```
fedepaol: this is not important here I think

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,850 @@
...
+  verbs: ["get", "update", "patch"]
+```
+
+### Scalability Considerations
+
+- **Node Watch Overhead**: The controllers watch Node resources for label changes. In clusters with frequent node label updates, this may increase reconciliation frequency. We can improve this by making each daemonset pod reconcile only on the running node using controller-runtime cache filter mechanism.
+- **Selector Evaluation**: For each reconciliation, the controller evaluates node selectors against the node where the daemonset is running, so reconciliation is done in parallel between the nodes.
+- **Status Updates**: Status updates are per-resource, avoiding conflicts from concurrent updates
+
+## Alternatives
```
fedepaol: I don't think this section is useful as it's quite a bad alternative :P

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,850 @@
...
+- More complex to manage when multiple parameters vary per-node
+- Doesn't align with Kubernetes patterns (e.g., DaemonSet uses pod-level selectors, not container-level)
+
+## Implementation Plan
+
+**Implementation Disclaimer:**
+
+This implementation is divided into two phases:
+- **Phase 1** (Required): Core functionality including API changes, controller implementation, documentation, and E2E testing. This phase delivers a complete, production-ready feature.
+- **Phase 2** (Optional): Enhanced status reporting that depends on the [status-crd enhancement](https://github.com/openperouter/openperouter/blob/main/enhancements/status-crd.md). This phase can be implemented once the status-crd enhancement is complete.
```
fedepaol: see before

enhancements/per-node-configuration.md
qinqon: But this is already implemented right ? like right now if you for example have two l2vnis with the same vrf they fail, so is all about limiting this for example for l2vnis with the same node selectors.

enhancements/per-node-configuration.md
qinqon: Really bad :-), I will remove alternatives, since I forced that in.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,850 @@
...
+- Update API documentation for all CRDs
+- Create example configurations for:
+  - **Underlay**: Multi-rack, multi-datacenter, mixed hardware deployments
+  - **L3VNI**: Per-rack VNI, zone-specific VNI configurations
+  - **L2VNI**: Selective deployment, per-rack gateway configurations
+  - **L3Passthrough**: Edge nodes, per-security-zone configurations
+- Migration guide from cluster-wide to per-node configuration
+- Troubleshooting guide for selector conflicts
+
+**Testing:**
```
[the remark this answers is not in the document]
qinqon: I will simplify this massive list

qinqon:
[force-push](https://github.com/openperouter/openperouter/compare/6237a0b0356641b7a5e811ee09610a074cd800d5..818eaf69112f681cda67f07932b07335fb50418a):
- Remove status reporting
- Add controller validation
- Simplify testing section
- Removed alternatives

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,850 @@
...
+**Migration Path**:
+N/A
+
+### Validation Rules
+
+The validation webhook enforces:
+
+1. **No Overlapping Selectors for Underlay**: Multiple Underlays cannot match the same node
+2. **Selector Validity**: Node selector must be a valid `metav1.LabelSelector` for all CRD types
+3. **Multiple instances allowed**: L3VNI, L2VNI, and L3Passthrough can have multiple instances matching the same node (no overlap validation)
```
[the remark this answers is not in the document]
fedepaol: Not sure I understand but:

- the webhook will have to filter per node so it won't allow overlapping resources to be on the same node
- the reconcile loop will have to check resources that match the current node, and ensure they don't conflict

fedepaol:
LGTM, @maiqueb wanna give another look? Otherwise we can merge this

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,656 @@
...
+- **Security zones**: Different security zones may require different passthrough configurations
+- **Specialized workloads**: Only nodes running certain workloads should have passthrough capabilities
+
+### Goals
+
+- Enable per-node configuration for all OpenPERouter CRDs using Kubernetes node selectors
+- Support heterogeneous network configurations across different hardware platforms, racks, and zones
+- Allow multiple instances of each CRD to coexist, each targeting specific node subsets
+- Maintain backward compatibility with existing cluster-wide configurations
+- Prevent configuration conflicts through validation (no node should be matched by multiple instances of the same CRD type)
```
maiqueb: wait, I'm confused. For instance, why can't the user have multiple L2 VNIs in the same node ? 

I think this goal is not right - you probably forgot to adjust the text.

qinqon: Right let me updated this.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,656 @@
...
+
+### Non-Goals
+
+- Dynamic node selector updates based on workload scheduling (this is handled by node labels)
+- Act external events no related to node labels, like interfaces names and similar ones
```
maiqueb: IMHO this one does not read well. How about something like 

```suggestion
- Handle external events unrelated to node labels, like interfaces names and similar ones
```

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,656 @@
...
+
+**Documentation and Examples:**
+- Update API documentation for all CRDs
+- Create example configurations for:
+  - **Underlay**: Multi-rack, multi-datacenter, mixed hardware deployments
+  - **L3VNI**: Per-rack VNI, zone-specific VNI configurations
+  - **L2VNI**: Selective deployment, per-rack gateway configurations
+  - **L3Passthrough**: Edge nodes, per-security-zone configurations
+- Migration guide from cluster-wide to per-node configuration
+- Troubleshooting guide for selector conflicts
```
maiqueb: I think we can offload this discussion to the PR.

maiqueb (commented):
I just have some minor cosmetic / content questions. I think you can omit some information from the enhancement. 

I do wonder if we can have the controller reconcile loop exclusively. Could we actually omit the admission webhook ?

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,656 @@
...
+1. **No Overlapping Selectors for Underlay**: Multiple Underlays cannot match the same node
+2. **Selector Validity**: Node selector must be a valid `metav1.LabelSelector` for all CRD types
+3. **Multiple not overlapping per node instances allowed**: L3VNI, L2VNI, and L3Passthrough can have multiple instances matching the same node, but they cannot conflict within same node
+
+#### Reconcile Loop Validation
+
+**Important**: Validation must also happen in the reconcile loop because webhooks might let configurations slip in during parallel creation of resources. For example:
+- Two Underlay resources with overlapping node selectors might be created simultaneously
+- Each webhook validation might pass because the other resource doesn't exist yet in the webhook's view
+- Both resources could be admitted, violating the "one Underlay per node" constraint
```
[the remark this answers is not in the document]
qinqon: > this feels redundant; I think we went over this a few sentences above in `Conflict Resolution` for the underlay section. 
> 
> Maybe we can restructure this a bit, removing this redundancy ?

Right, removing.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,656 @@
...
+#### Conflict Resolution
+
+**Underlay-specific restriction:**
+
+Only one Underlay can match a given node at any time. This is enforced at two levels:
+
+**Admission webhook (first line of defense):**
+1. List all existing Underlay resources
+2. For each new/updated Underlay, check if any nodes would match multiple Underlays
+3. Reject the operation if overlapping selectors are detected
```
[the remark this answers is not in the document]
qinqon: Ahh right, yes we should limit as much as possible

enhancements/per-node-configuration.md
qinqon: I will remove the whole section

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,599 @@
...
+
+1. **Admission webhook**: First-line defense that lists all Underlays and rejects operations with overlapping selectors
+2. **Reconcile loop**: Re-validates to catch race conditions, updates status conditions, and skips invalid configurations
+
+**L3VNI, L2VNI, and L3Passthrough:**
+
+Multiple instances can match the same node (for multi-tenancy), but the shared validation logic prevents conflicts such as:
+- Two L2VNIs with the same VRF on one node
+- VNI number conflicts across resource types
+- Incompatible configurations
```
maiqueb: I am assuming these will be checked in the controller side. Or do you also plan on checking these in a webhook ? I would favor the former (controller side) for anything that requires checking more than the data in the request for creating/updating the CR.

enhancements/per-node-configuration.md
```diff
@@ -0,0 +1,601 @@
...
+
+#### Conflict Resolution
+
+Conflict resolution uses **shared validation logic** between the admission webhook and reconcile loop. Controllers protect themselves without depending on webhooks.
+
+**Underlay-specific restriction:**
+
+Only one Underlay can match a given node. Enforced at two levels using shared validation code:
+
+1. **Admission webhook**: First-line defense that will validate the current resource received at the webhook request without looking at other resurces.
```
maiqueb: seems that we already are looking into other resources: https://github.com/openperouter/openperouter/blob/3ecbac7c2f4eb398b41505e76a58e955c07b7575/internal/webhooks/l2vni_webhook.go#L104

I please put this back the way it was. Sorry for only having noticed it now. 
I'll lgtm it afterwards.

qinqon: I am going to remove it, we need to not break what we have already, but that's implementation details.

PR #161, opened 2025-11-07
Add the design proposal for running the router directly on the host

fedepaol:

>
> /kind bug
> /kind cleanup
> /kind feature
/kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

As the proposal explains, we might want to run the router as a systemd unit BEFORE k8s starts. Here we explaing how and why. Creating backwards to provide the context / discussion space for https://github.com/openperouter/openperouter/pull/158 , which I started working on when there was not much room for discussing with others.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/enhancements/running-on-the-host.md
+++ b/enhancements/running-on-the-host.md
@@ -19,6 +19,7 @@ To solve this chicken-egg problem we need an operating mode where the router is 
 ## Non Goals
 
 - Having an alternative way to provide the static configuration that involves distritbuting the configuration to the routers
+- Dinamically change the static configuration from the kubernetes API
 
 ## Proposal
 
@@ -72,6 +73,8 @@ The static configuration must be provided to the OpenPERouter as files in a know
 
 - A static configuration with the same structure of the CRDs (ie underlays, vnis) to be used for bringing up the basic connectivity
 - An extra configuration to fill the requirements of running on the host. At the time of writing, the only extra configuration required seem to be a node index, to be associated manually (or by automation) to be different per each node.
+- Additionally, it must be possible to set the relevant parameters of the controller / router via a configuration file. This will make the distribution of the systemd unit easier, as the
+binaries will expect their own configuration in a canonical place
 
 #### Consuming the Kubernetes API
 
@@ -82,6 +85,11 @@ Running the controller on the host while it's still be able to access the kubern
 The static configuration must be treated exactly as configuration coming from the CRs. In this sense it must be validated, and the static configuration must be merged with the configuration read by CRs.
 
 This would allow a scenario where the UNDERLAY and a VNI is provided via static configuration, and extra VNIs are provided as day 2 configuration.
+
+#### Modifying the static configuration
+
+The static configuration will be visible as Kubernetes resource (see the part below about observability). However, it will only be a mirror of what is configured locally. Any change of the
+mirror won't have effect and the controller will reconcile it back.
 
 ### Challenges
 


enhancements/running-on-the-host.md
```diff
@@ -0,0 +1,167 @@
...
+- A static configuration with the same structure of the CRDs (ie underlays, vnis) to be used for bringing up the basic connectivity
+- An extra configuration to fill the requirements of running on the host. At the time of writing, the only extra configuration required seem to be a node index, to be associated manually (or by automation) to be different per each node.
+
+#### Consuming the Kubernetes API
+
+Running the controller on the host while it's still be able to access the kubernetes API is a challenge. One possible way to solve this is to have a proxy pod running on the cluster and translating its credentials to the controller running on the host.
+
+#### Merging the two configurations
+
+The static configuration must be treated exactly as configuration coming from the CRs. In this sense it must be validated, and the static configuration must be merged with the configuration read by CRs.
```
zeeke: What if the static configuration collides with a resource in the API server? 
It's likely an implementation detail, but the static configuration should be put into the API server, shouldn't it?

enhancements/running-on-the-host.md
```diff
@@ -0,0 +1,167 @@
...
+### Design Details
+
+The general idea is to run the same set of containers we have in pods today, but on the host. This includes:
+
+- running the router pod, in charge of providing the network control plane with FRR and the data plane with veths and the underlay interface
+- running the controller pod, in charge of reading the static configuration and receiving the additional configuration when the kubernetes api is available
+
+#### Handling the lifecycle of the components
+
+The router and the controller need to run as systemd units. This allows us to run them before the kubelet and the CNI start.
```
qinqon: Have you check if static pods fits this so we don't need external containers ? as far as I know if static pods use hostNetwork: true they run before CNI which is exactly what we need, we will have the pods running as part of kubernetes but we break the chicken egg problem that we have.

enhancements/running-on-the-host.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Interactions
+
+There are some interactions between the controller process and the router (and the surrounding system) that need to be adapted to the new scenario:
+
+##### Retrieving the router's network namespace
+
+We are using the container runtime api to retrieve the network namespace of the pod.
+When running on the host, the target namespace can be reitrieved by telling podman to fill a pid file, and by using the pid to retrieve the network namespace of the process.
+
```
qinqon: Is it possible to run the controller as the router pod so they share the network namespace ? I understand that is not possible becuase the config reload that restart the router pod, but maybe with alternative reload mechanism they can live together so network namespace discovery is not needed.

enhancements/running-on-the-host.md
```diff
@@ -0,0 +1,167 @@
...
+##### Restarting the router pod
+
+The controller pod currently deletes the router pod to restart the underlay creation machinery. When running on the host, the controller pod can interact either with Podman or with Systemd to restart the pod.
+
+#### Consuming the static configuration
+
+The static configuration must be provided to the OpenPERouter as files in a known location. The configuration is composed by:
+
+- A static configuration with the same structure of the CRDs (ie underlays, vnis) to be used for bringing up the basic connectivity
```
qinqon: It is just the structure or it will be exactly the same yamls ? can you show an example in case it's different ? I think it can be directly the same yaml even with the kubevirt bits, so users need to understand just one type and it can be consumed directly as kube api provided resource.

enhancements/running-on-the-host.md
```diff
@@ -0,0 +1,167 @@
...
+- An extra configuration to fill the requirements of running on the host. At the time of writing, the only extra configuration required seem to be a node index, to be associated manually (or by automation) to be different per each node.
+
+#### Consuming the Kubernetes API
+
+Running the controller on the host while it's still be able to access the kubernetes API is a challenge. One possible way to solve this is to have a proxy pod running on the cluster and translating its credentials to the controller running on the host.
```
qinqon: Similar situation happend when CNI's need to access api server, what they have is the kubeconfig stored under their static config dir /etc/cni.d.

We can do similar here, when kubernetes start a openperouter pod just generate the kubeconfig and store it under the openperouter static config the the controller can use it, this way we don't need a running proxy pod at kubernetes it will be a one shot installer pod.

enhancements/running-on-the-host.md
```diff
@@ -0,0 +1,167 @@
...
+
+- The status CRDs that are discussed in the [status enhancement](https://github.com/openperouter/openperouter/blob/main/enhancements/status-crd.md)
+must keep working
+- Prometheus metrics added to the pods must be available from the Kubernetes cluster.
+- The pod responsible of providing the kubernetes credentials to the process running on the host will mirror the statically provided configuration to
+CRs with an appropriate name. This will make it easier to understand the whole configuration of the cluster.
+
+## Testing
+
+When running in host mode, the same test suite we have today must work and the tests must pass.
```
qinqon: Except that we cannot modify static config

qinqon (changes requested):
Some maybe crazy alternatives and questions

enhancements/running-on-the-host.md
fedepaol: This might be already late in the lifecycle, depending on the assumptions made by the CNI on how to deploy it. For example, OVN-Kubernetes relies on the default gateway to be setup, way earlier than the kubelet starts

enhancements/running-on-the-host.md
fedepaol: sorry but I didn't get the ask. The controller runs as a different pod as it does on k8s, because it must be hostnetworked in order to move interfaces around and create veths and move one leg of the veth in the network ns of the router

enhancements/running-on-the-host.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Consuming the Kubernetes API
+
+Running the controller on the host while it's still be able to access the kubernetes API is a challenge. One possible way to solve this is to have a proxy pod running on the cluster and translating its credentials to the controller running on the host.
+
+#### Merging the two configurations
+
+The static configuration must be treated exactly as configuration coming from the CRs. In this sense it must be validated, and the static configuration must be merged with the configuration read by CRs.
+
+This would allow a scenario where the UNDERLAY and a VNI is provided via static configuration, and extra VNIs are provided as day 2 configuration.
```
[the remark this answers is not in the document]
fedepaol: right, and the node won't be able to join the fabric I guess

enhancements/running-on-the-host.md
qinqon: Ahh wait right... you can disregard this, we don't have host network at routing pod since everything runs at the router namespace.

enhancements/running-on-the-host.md
qinqon: I kind of like the GRPC approach, although I see it split the controller.

enhancements/running-on-the-host.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Challenges
+
+#### Troubleshooting and triaging
+
+When the components are running on the host, it's harder to inspect logs (as it requires ssh-ing to the node and checking the logs of the systemd unit / podman containers).
```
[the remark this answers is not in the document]
qinqon: Well since nodes will not be able to join the frabric, we will not reach this point :-/ or maybe there are not that catastrophic problems that we should expose the the api server ?

fedepaol:
@zeeke @qinqon tried to address the concerns in the new commit

enhancements/running-on-the-host.md
fedepaol: my idea was to keep the static config unmutable. This is because that configuration is likely to be required for the basic connectivity, and any change is disruptive. Let me clarify this.
As a starter, I think it's more than valid. We can add what you are suggesting later if needed.

PR #163, opened 2025-11-14
make: Add clean targets for single and multi-cluster teardown

qinqon:
This commit introduces cleanup targets in the Makefile to simplify the teardown of both single and multi-cluster containerlab environments:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,10 +316,7 @@ clab-multi-cluster: kind-node-image-build ## Deploy multi-cluster setup with 2 k
 
 .PHONY: clean
 clean: kind ## Shutdown and clean up kind cluster(s) and containerlab topology.
-	KUBECONFIG_PATH=$(KUBECONFIG_PATH) KIND=$(KIND) CLAB_TOPOLOGY=$(CLAB_TOPOLOGY_FILE) clab/clean.sh
-
-.PHONY: clean-multi
-clean-multi: kind ## Shutdown and clean up multi-cluster setup.
+	KUBECONFIG_PATH=$(KUBECONFIG_PATH) KIND=$(KIND) CLAB_TOPOLOGY=singlecluster/kind.clab.yml clab/clean.sh pe-kind
 	KUBECONFIG_PATH=$(KUBECONFIG_PATH) KIND=$(KIND) CLAB_TOPOLOGY=multicluster/kind.clab.yml clab/clean.sh pe-kind-a pe-kind-b
 
 .PHONY: load-on-kind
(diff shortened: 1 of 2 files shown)


fedepaol:
please remove unnecessary comments (we should probably tweak claude.md for that)

Makefile
```diff
@@ -295,6 +295,14 @@ clab-multi-cluster: ## Deploy multi-cluster setup with 2 kindleafs, 2 kind switc
 	@echo '  - Cluster A: export KUBECONFIG=${KUBECONFIG_PATH}-pe-kind-a'
 	@echo '  - Cluster B: export KUBECONFIG=${KUBECONFIG_PATH}-pe-kind-b'
 
+.PHONY: clean
+clean: kind ## Shutdown and clean up kind cluster(s) and containerlab topology.
```
fedepaol: can we figure out what's running and react accordingly?

clab/clean.sh
```diff
@@ -0,0 +1,96 @@
...
+
+echo "=== Cleaning up bridge interfaces ==="
+if [[ "$CLUSTER_MODE" == "single" ]]; then
+    bridge_name="leafkind-switch"
+    if [[ -d "/sys/class/net/${bridge_name}" ]]; then
+        echo "Removing bridge: ${bridge_name}"
+        sudo ip link delete ${bridge_name} 2>/dev/null || true
+    fi
+else
+    for cluster_name in "${CLUSTER_ARRAY[@]}"; do
```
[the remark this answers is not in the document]
qinqon: You mean always use an array but with single it will have just one entry so we remove the if statements ?

Makefile
fedepaol: I mean having one single clean rule that finds out if the current cluster is single or multi

clab/clean.sh
```diff
@@ -0,0 +1,83 @@
...
+fi
+
+# Initialize kubeconfig files array based on cluster mode
+if [[ ${#CLUSTER_ARRAY[@]} -gt 1 ]]; then
+    CLUSTER_MODE="multi"
+    KUBECONFIG_FILES=()
+    for cluster_name in "${CLUSTER_ARRAY[@]}"; do
+        KUBECONFIG_FILES+=("${KUBECONFIG_PATH}-${cluster_name}")
+    done
+    KUBECONFIG_FILES+=("${KUBECONFIG_PATH}")
```
fedepaol: this can go outside of the if as it's shared

qinqon: You mean make it an array always right ?

PR #187, opened 2025-12-16
e2etests: return error inside eventually block

maiqueb:

>

/kind bug

> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

The Expect inside Eventually prevents retries when bgpd isn't ready;
this change returns the error instead, which will enable the retry
logic.

Another option would be to receive `ginkgo` as an input parameter in
eventually, and run that `ginkgo` instance expectations. This would
allow the eventually block to be re-tried. I chose to follow the pattern
in place in this function, which is to return the errors.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/tests/evpn_routes.go
+++ b/e2etests/tests/evpn_routes.go
@@ -139,7 +139,9 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
 				Eventually(func() error {
 					for exec := range routers.GetExecutors() {
 						evpn, err := frr.EVPNInfo(exec)
-						Expect(err).NotTo(HaveOccurred())
+						if err != nil {
+							return fmt.Errorf("failed to get EVPN info from %s: %w", exec.Name(), err)
+						}
 						for _, prefix := range prefixes {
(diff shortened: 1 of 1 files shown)


e2etests/tests/evpn_routes.go
```diff
@@ -139,7 +139,9 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
 				Eventually(func() error {
 					for exec := range routers.GetExecutors() {
 						evpn, err := frr.EVPNInfo(exec)
-						Expect(err).NotTo(HaveOccurred())
+						if err != nil {
```
fedepaol: makes total sense, I wonder if you saw evidence in some failures as we (in theory) should be covered by the healthprobe that ensures the router pod is ready when the daemons are up

fedepaol: ack, I think it was deliberate as in "this should never fail", but won't retrying won't hurt.

PR #189, opened 2025-12-22
e2etests: add VNI scale tests with metrics collection

maiqueb:
Add scale tests to measure resource consumption (memory) and reconcile time
of router and controller pods as the number of L2VNI and L3VNI resources
increases. This helps identify potential scaling bottlenecks and establish
baseline performance characteristics.

### Scale test suite
- Dedicated `e2etests/scale_suite` with its own `scale_tests` package, decoupled from the main e2e suite
- Table-driven tests for 1, 50, 150, 200, 250, 300, 350, 400, 450, and 500 VNIs
- Four test groups: L2VNI-only and L2+L3VNI, each with linux-bridge and ovs-bridge host masters
- Uses Ginkgo's `gmeasure` experiment API to record reconcile duration and memory per scenario
- Metrics collection polls `kubectl top` until memory readings stabilize (convergence-based, not snapshot)
- Metrics-server availability is checked in `BeforeAll`; suite fails fast if unavailable

### Metrics package (`e2etests/pkg/metrics`)
- Parses `kubectl top pods` output into structured pod metrics
- `FetchMetricsAggregation` polls until two consecutive readings are within tolerance
- `CheckAvailability` verifies metrics-server is reachable
- Unit tests for `parseCPU` and `parseMemory`

### CI and Makefile
- `make scale-tests` runs the dedicated scale suite
- `make deploy-scale` deploys with scale-specific kustomize overlay (no resource limits) and installs metrics-server
- `make deploy-metrics-server` available as standalone target
- `make parse-scale-report` parses the JSON report into text tables and CSV
- CI workflow job runs on `run-scale-tests` label; uploads JSON report as artifact

### Configurable vtysh timeout
- `vtysh.NewCLIWithTimeout` / `vtysh.NewCLI` replace the old hardcoded timeout
- Reloader accepts `--vtysh-timeout` flag, wired through helm and operator

## Possible follow-ups
- Scale tests with pre-existing (shared) OVS bridge
- L3VNI-only scale tests (without associated L2VNIs)
- Graph/plot generation via Ginkgo Reporter on regression

/kind feature

Add scale tests to measure memory consumption and reconcile time of
router and controller pods as VNI count increases, plus make the vtysh
command timeout configurable.

**Release note**:
```release-note
Have configurable vtysh timeout, defined from the helm / operator config.
```

---

Assisted-by: Claude Opus 4.5 <noreply@anthropic.com>
Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,7 @@ name: CI
 on:
   merge_group:
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
     paths-ignore:
       - 'DCO'
       - 'LICENSE'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -292,3 +292,124 @@ jobs:
         uses: ./.github/workflows/composite/collectlogs
         with:
           artifact-name: kind-logs-${{ matrix.deployment }}
+
+  scale-tests:
+    if: contains(github.event.pull_request.labels.*.name, 'run-scale-tests')
+    runs-on: ubuntu-22.04
+    env:
+      NODE_IMAGE: quay.io/openperouter/kind-node-openperouter:v1.32.2
+    needs:
+      - unit-tests
+      - build-test-images
+
+    steps:
+      - name: Free Disk Space
+        working-directory: ${{ github.workspace }}
+        run: |
+          echo "Disk space before cleanup:"
+          df -h
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /usr/local/share/boost
+          sudo apt-get clean
+          echo "Disk space after cleanup:"
+          df -h
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: "go.mod"
+          cache: true
+
+      - name: Install kernel modules
+        run: |
+          sudo apt-get update
+          sudo apt-get install linux-modules-extra-$(uname -r)
+
+      - name: Disable containerd image store
+        run: |
+          sudo bash -c 'cat > /etc/docker/daemon.json << EOF
+          {
+            "features": {
+              "containerd-snapshotter": false
+            }
+          }
+          EOF'
+          sudo systemctl restart docker
+          docker info -f '{{ .DriverStatus }}'
+
+      - name: Setup docker buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build kind-node-openperouter image
+        uses: docker/build-push-action@v5
+        with:
+          context: hack/kind-node-image
+          file: hack/kind-node-image/Dockerfile
+          build-args: |
+            KIND_NODE_VERSION=v1.32.2
+          tags: quay.io/openperouter/kind-node-openperouter:v1.32.2
+          load: true
+          cache-from: type=gha,scope=kind-node-openperouter
+          cache-to: type=gha,mode=max,scope=kind-node-openperouter
+
+      - name: Download openperouter image
+        uses: actions/download-artifact@v4
+        with:
+          path: image
+
+      - name: Load image
+        working-directory: image
+        run: |
+          docker load -i image-tar-openperouter/openperouter.tar
+
+      - name: Deploy the cluster
+        run: |
+          make deploy-scale
+
+      - name: Install metrics-server
+        env:
+          KUBECONFIG: bin/kubeconfig
+        run: |
+          kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml
+          kubectl patch -n kube-system deployment metrics-server --type=json \
+            -p='[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--kubelet-insecure-tls"}]'
+          # Wait for deployment rollout to complete (not just any pod ready)
+          # This ensures the patched pod is fully running, not the original pre-patch pod
+          kubectl -n kube-system rollout status deployment/metrics-server --timeout=300s
+          # Give metrics-server time to collect initial metrics
+          sleep 60
+
+      - name: Run scale tests
+        run: |
+          make scale-tests
+
+      - name: Parse scale test report
+        if: ${{ always() }}
+        run: |
+          python3 hack/parse-scale-report.py /tmp/kind_logs/e2etests_scale_suite_scale-report.json \
+            | tee /tmp/kind_logs/scale-test-summary.txt
+
+      - name: Upload scale test report
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          retention-days: 30
+          name: scale-test-report
+          path: |
+            /tmp/kind_logs/e2etests_scale_suite_scale-report.json
+            /tmp/kind_logs/scale-test-summary.txt
+
+      - name: Export kind logs
+        if: ${{ failure() }}
+        run:
+          make kind-export-logs
+
+      - name: Collect Logs
+        if: ${{ failure() }}
+        uses: ./.github/workflows/composite/collectlogs
+        with:
+          artifact-name: kind-logs-scale-tests
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -381,8 +381,7 @@ jobs:
       - name: Parse scale test report
         if: ${{ always() }}
         run: |
-          python3 hack/parse-scale-report.py /tmp/kind_logs/e2etests_scale_suite_scale-report.json \
-            | tee /tmp/kind_logs/scale-test-summary.txt
+          make parse-scale-report | tee /tmp/kind_logs/scale-test-summary.txt
 
       - name: Upload scale test report
         if: ${{ always() }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -371,18 +371,8 @@ jobs:
         run: |
           make deploy-scale
 
-      - name: Install metrics-server
-        env:
-          KUBECONFIG: bin/kubeconfig
-        run: |
-          kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml
-          kubectl patch -n kube-system deployment metrics-server --type=json \
-            -p='[{"op":"add","path":"/spec/template/spec/containers/0/args/-","value":"--kubelet-insecure-tls"}]'
-          # Wait for deployment rollout to complete (not just any pod ready)
-          # This ensures the patched pod is fully running, not the original pre-patch pod
-          kubectl -n kube-system rollout status deployment/metrics-server --timeout=300s
-          # Give metrics-server time to collect initial metrics
-          sleep 60
+      - name: Wait for metrics-server to collect initial metrics
+        run: sleep 60
 
       - name: Run scale tests
         run: |
--- a/charts/openperouter/templates/router.yaml
+++ b/charts/openperouter/templates/router.yaml
@@ -177,7 +177,10 @@ spec:
         {{- with .Values.openperouter.logLevel }}
         - --loglevel={{ . }}
         {{- end }}
-        securityContext: 
+        {{- with .Values.openperouter.frr.reloader.vtyshTimeout }}
+        - --vtysh-timeout={{ . }}
+        {{- end }}
+        securityContext:
           seLinuxOptions:
             type: spc_t
         volumeMounts:
--- a/cmd/reloader/main.go
+++ b/cmd/reloader/main.go
@@ -44,6 +44,7 @@ type Args struct {
 	frrConfigPath string
 	logLevel      string
 	unixSocket    string
+	vtyshTimeout  time.Duration
 }
 
 func main() {
@@ -52,6 +53,8 @@ func main() {
 	flag.StringVar(&args.unixSocket, "unixsocket", "", "Unix socket path to listen on")
 	flag.StringVar(&args.logLevel, "loglevel", "info", "The log level of the process")
 	flag.StringVar(&args.frrConfigPath, "frrconfig", "/etc/frr/frr.conf", "The path the frr configuration is at")
+	flag.DurationVar(&args.vtyshTimeout, "vtysh-timeout", vtysh.DefaultTimeout,
+		"Timeout for vtysh commands used in health checks")
 	flag.Parse()
 
 	_, err := logging.New(args.logLevel)
@@ -87,10 +90,11 @@ func serveReload(args Args) error {
 		[]handlerConfig{{pattern: "/", handler: reloadHandler(args.frrConfigPath)}},
 	)
 
+	healthHandler := health(vtysh.NewCLIWithTimeout(args.vtyshTimeout))
 	healthServer := newServer(
 		[]handlerConfig{
-			{pattern: "/healthz", handler: health()},
-			{pattern: "/readyz", handler: health()},
+			{pattern: "/healthz", handler: healthHandler},
+			{pattern: "/readyz", handler: healthHandler},
 		},
 		withEndpoint(args.bindAddress),
 	)
@@ -157,14 +161,14 @@ func reloadHandler(frrConfigPath string) func(w http.ResponseWriter, req *http.R
 	}
 }
 
-func health() func(w http.ResponseWriter, req *http.Request) {
+func health(frrCli vtysh.Cli) func(w http.ResponseWriter, req *http.Request) {
 	return func(w http.ResponseWriter, req *http.Request) {
 		if req.Method != http.MethodGet {
 			http.Error(w, "invalid method", http.StatusMethodNotAllowed)
 			return
 		}
 
-		if err := liveness.PingFrr(vtysh.Run); err != nil {
+		if err := liveness.PingFrr(frrCli); err != nil {
 			slog.Error("health check ping frr", "error", err)
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
--- a/config/scale/kustomization.yaml
+++ b/config/scale/kustomization.yaml
@@ -1,0 +1,19 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+- ../default
+
+patches:
+- target:
+    kind: DaemonSet
+    name: controller
+  patch: |-
+    - op: remove
+      path: /spec/template/spec/containers/0/resources
+- target:
+    kind: Deployment
+    name: nodemarker
+  patch: |-
+    - op: remove
+      path: /spec/template/spec/containers/0/resources
--- a/e2etests/pkg/k8s/nad.go
+++ b/e2etests/pkg/k8s/nad.go
@@ -17,16 +17,25 @@ import (
 
 var nadClient *nadclientset.Clientset
 
-func init() {
-	var err error
-	config := ctrl.GetConfigOrDie()
+func initNADClient() error {
+	if nadClient != nil {
+		return nil
+	}
+	config, err := ctrl.GetConfig()
+	if err != nil {
+		return fmt.Errorf("failed to load kubeconfig for NAD client: %w", err)
+	}
 	nadClient, err = nadclientset.NewForConfig(config)
 	if err != nil {
-		panic(err)
+		return fmt.Errorf("failed to create NAD client: %w", err)
 	}
+	return nil
 }
 
 func CreateMacvlanNad(name, namespace, master string, gatewayIPs []string) (nad.NetworkAttachmentDefinition, error) {
+	if err := initNADClient(); err != nil {
+		return nad.NetworkAttachmentDefinition{}, err
+	}
 	// Build routes based on gateway IPs (IPv4 and/or IPv6)
 	var routes []string
 	for _, gwIP := range gatewayIPs {
--- a/e2etests/pkg/k8sclient/client.go
+++ b/e2etests/pkg/k8sclient/client.go
@@ -18,6 +18,6 @@ func New() *clientset.Clientset {
 	return res
 }
 
-func RestConfig() *restclient.Config {
-	return ctrl.GetConfigOrDie()
+func RestConfig() (*restclient.Config, error) {
+	return ctrl.GetConfig()
 }
--- a/e2etests/pkg/metrics/metrics.go
+++ b/e2etests/pkg/metrics/metrics.go
@@ -1,0 +1,179 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package metrics
+
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"net/url"
+	"os/exec"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// Aggregated contains aggregate statistics for a set of pods.
+type Aggregated struct {
+	TotalCPU float64
+	TotalMem float64
+	AvgCPU   float64
+	AvgMem   float64
+}
+
+// MemoryConvergenceConfig controls the polling behavior of FetchMetricsAggregation.
+type MemoryConvergenceConfig struct {
+	PollInterval time.Duration
+	Timeout      time.Duration
+	ToleranceMB  float64
+}
+
+func DefaultMemoryConvergenceConfig() MemoryConvergenceConfig {
+	return MemoryConvergenceConfig{
+		PollInterval: 5 * time.Second,
+		Timeout:      90 * time.Second,
+		ToleranceMB:  1.0,
+	}
+}
+
+// CheckAvailability verifies that metrics-server is running by attempting
+// to collect pod metrics for the given label selector.
+func CheckAvailability(kubectl, namespace, labelSelector string) error {
+	_, err := forPod(kubectl, namespace, labelSelector)
+	return err
+}
+
+// FetchMetricsAggregation polls kubectl top until two consecutive total-memory
+// readings are within ToleranceMB of each other, ensuring at least one
+// metrics-server refresh has been observed.
+func FetchMetricsAggregation(
+	kubectl, namespace, labelSelector string,
+	cfg MemoryConvergenceConfig,
+) (Aggregated, error) {
+	fetchAggregation := func() (Aggregated, error) {
+		pods, err := forPod(kubectl, namespace, labelSelector)
+		if err != nil {
+			return Aggregated{}, err
+		}
+		return summarize(pods), nil
+	}
+
+	prev, err := fetchAggregation()
+	if err != nil {
+		return Aggregated{}, fmt.Errorf("initial metrics poll failed: %w", err)
+	}
+
+	ticker := time.NewTicker(cfg.PollInterval)
+	defer ticker.Stop()
+	deadline := time.After(cfg.Timeout)
+
+	for {
+		select {
+		case <-deadline:
+			return prev, fmt.Errorf("metrics did not stabilize within %s (last two readings: %.2f MB, %.2f MB)",
+				cfg.Timeout, prev.TotalMem, prev.TotalMem)
+		case <-ticker.C:
+			curr, err := fetchAggregation()
+			if err != nil {
+				return Aggregated{}, fmt.Errorf("metrics poll failed: %w", err)
+			}
+			if math.Abs(curr.TotalMem-prev.TotalMem) <= cfg.ToleranceMB {
+				return curr, nil
+			}
+			prev = curr
+		}
+	}
+}
+
+func summarize(podsMetrics []podMetrics) Aggregated {
+	if len(podsMetrics) == 0 {
+		return Aggregated{}
+	}
+
+	var result Aggregated
+	for _, p := range podsMetrics {
+		result.TotalCPU += p.cpuMillicores
+		result.TotalMem += p.memoryMB
+	}
+
+	result.AvgCPU = result.TotalCPU / float64(len(podsMetrics))
+	result.AvgMem = result.TotalMem / float64(len(podsMetrics))
+	return result
+}
+
+type podMetrics struct {
+	podName       string
+	namespace     string
+	cpuMillicores float64
+	memoryMB      float64
+	timestamp     time.Time
+}
+
+type podMetricsList struct {
+	Items []podMetricsItem `json:"items"`
+}
+
+type podMetricsItem struct {
+	Metadata   podMetricsMetadata   `json:"metadata"`
+	Containers []containerMetrics   `json:"containers"`
+}
+
+type podMetricsMetadata struct {
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+}
+
+type containerMetrics struct {
+	Usage map[string]string `json:"usage"`
+}
+
+func forPod(kubectl, namespace, labelSelector string) ([]podMetrics, error) {
+	path := fmt.Sprintf("/apis/metrics.k8s.io/v1beta1/namespaces/%s/pods?labelSelector=%s",
+		namespace, url.QueryEscape(labelSelector))
+	out, err := exec.Command(kubectl, "get", "--raw", path).CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("metrics API request failed (is metrics-server running?): %s: %w", string(out), err)
+	}
+
+	return parseMetricsAPIResponse(out)
+}
+
+func parseMetricsAPIResponse(data []byte) ([]podMetrics, error) {
+	var list podMetricsList
+	if err := json.Unmarshal(data, &list); err != nil {
+		return nil, fmt.Errorf("failed to parse metrics API response: %w", err)
+	}
+
+	result := make([]podMetrics, 0, len(list.Items))
+	for _, item := range list.Items {
+		var totalCPUMilli int64
+		var totalMemBytes int64
+
+		for _, c := range item.Containers {
+			if cpuStr, ok := c.Usage["cpu"]; ok {
+				q, err := resource.ParseQuantity(cpuStr)
+				if err != nil {
+					return nil, fmt.Errorf("failed to parse CPU quantity %q for pod %s: %w", cpuStr, item.Metadata.Name, err)
+				}
+				totalCPUMilli += q.MilliValue()
+			}
+			if memStr, ok := c.Usage["memory"]; ok {
+				q, err := resource.ParseQuantity(memStr)
+				if err != nil {
+					return nil, fmt.Errorf("failed to parse memory quantity %q for pod %s: %w", memStr, item.Metadata.Name, err)
+				}
+				totalMemBytes += q.Value()
+			}
+		}
+
+		result = append(result, podMetrics{
+			podName:       item.Metadata.Name,
+			namespace:     item.Metadata.Namespace,
+			cpuMillicores: float64(totalCPUMilli),
+			memoryMB:      float64(totalMemBytes) / (1024 * 1024),
+			timestamp:     time.Now(),
+		})
+	}
+
+	return result, nil
+}
--- a/e2etests/scale_tests/scale.go
+++ b/e2etests/scale_tests/scale.go
@@ -1,0 +1,298 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package scale_tests
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gmeasure"
+	"github.com/openperouter/openperouter/api/v1alpha1"
+	"github.com/openperouter/openperouter/e2etests/pkg/config"
+	"github.com/openperouter/openperouter/e2etests/pkg/executor"
+	"github.com/openperouter/openperouter/e2etests/pkg/infra"
+	"github.com/openperouter/openperouter/e2etests/pkg/metrics"
+	"github.com/openperouter/openperouter/e2etests/pkg/openperouter"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type scenarioUnderTest int
+
+const (
+	l2vni = iota
+	l3vni
+	l2l3vni
+)
+
+type scaleTestCase struct {
+	hostMaster string
+	scenario   scenarioUnderTest
+	vniCount   int
+}
+
+func (tc scaleTestCase) String() string {
+	if tc.scenario == l2l3vni {
+		return fmt.Sprintf("%d pairs", tc.vniCount)
+	}
+	return fmt.Sprintf("%d VNIs", tc.vniCount)
+}
+
+const (
+	routerLabelSelector     = "app=router"
+	controllerLabelSelector = "app=controller"
+)
+
+var _ = Describe("VNI Scale Tests", Ordered, func() {
+	var experiment *gmeasure.Experiment
+
+	BeforeAll(func() {
+		experiment = gmeasure.NewExperiment("VNI Scale")
+		AddReportEntry(experiment.Name, experiment)
+
+		By("Verifying metrics-server is available")
+		err := metrics.CheckAvailability(executor.Kubectl, openperouter.Namespace, routerLabelSelector)
+		Expect(err).NotTo(HaveOccurred(), "metrics-server must be running for scale tests")
+
+		By("Cleaning up any existing VNI resources")
+		Expect(Updater.CleanButUnderlay()).To(Succeed())
+
+		By("Ensuring underlay configuration exists")
+		err = Updater.Update(config.Resources{
+			Underlays: []v1alpha1.Underlay{infra.Underlay},
+		})
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		By("Cleaning up VNIs")
+		Expect(Updater.CleanButUnderlay()).To(Succeed())
+
+		By("Waiting for VNI resources to be fully removed")
+		waitForVNIsGone(Updater.Client())
+	})
+
+	AfterAll(func() {
+		By("Cleaning up all resources")
+		err := Updater.CleanAll()
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	Describe("L2VNI only - Linux Bridge", func() {
+		DescribeTable("VNI Scale Measurements",
+			func(tc scaleTestCase) {
+				runScaleTest(tc, experiment)
+			},
+			EntryDescription("%v"),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 1}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 50}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 150}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 200}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 250}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 300}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 350}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 400}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 450}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 500}),
+		)
+	})
+
+	Describe("L2VNI only - OVS Bridge", func() {
+		DescribeTable("VNI Scale Measurements",
+			func(tc scaleTestCase) {
+				runScaleTest(tc, experiment)
+			},
+			EntryDescription("%v"),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 1}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 50}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 150}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 200}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 250}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 300}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 350}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 400}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 450}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2vni, vniCount: 500}),
+		)
+	})
+
+	Describe("L3VNI+L2VNI - Linux Bridge", func() {
+		DescribeTable("VNI Scale Measurements",
+			func(tc scaleTestCase) {
+				runScaleTest(tc, experiment)
+			},
+			EntryDescription("%v"),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 1}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 50}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 150}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 200}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 250}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 300}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 350}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 400}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 450}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2l3vni, vniCount: 500}),
+		)
+	})
+
+	Describe("L3VNI+L2VNI - OVS Bridge", func() {
+		DescribeTable("VNI Scale Measurements",
+			func(tc scaleTestCase) {
+				runScaleTest(tc, experiment)
+			},
+			EntryDescription("%v"),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 1}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 50}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 150}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 200}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 250}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 300}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 350}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 400}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 450}),
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 500}),
+		)
+	})
+})
+
+func runScaleTest(tc scaleTestCase, experiment *gmeasure.Experiment) {
+	testLabel := CurrentSpecReport().FullText()
+
+	By("Creating " + testLabel)
+	resources := buildResources(tc)
+
+	experiment.MeasureDuration(testLabel+" reconcile", func() {
+		err := Updater.Update(resources)
+		Expect(err).NotTo(HaveOccurred())
+	}, gmeasure.Annotation(fmt.Sprintf("%d VNIs", tc.vniCount)))
+
+	By("Collecting stable memory metrics after reconcile")
+	routerMem := collectMetrics(routerLabelSelector)
+	ctrlMem := collectMetrics(controllerLabelSelector)
+
+	experiment.RecordValue(testLabel+" router memory",
+		routerMem.TotalMem,
+		gmeasure.Units("MB"), gmeasure.Precision(2),
+		gmeasure.Annotation(fmt.Sprintf("%d VNIs", tc.vniCount)),
+	)
+	experiment.RecordValue(testLabel+" controller memory",
+		ctrlMem.TotalMem,
+		gmeasure.Units("MB"), gmeasure.Precision(2),
+		gmeasure.Annotation(fmt.Sprintf("%d VNIs", tc.vniCount)),
+	)
+
+}
+
+func collectMetrics(labelSelector string) metrics.Aggregated {
+	result, err := metrics.FetchMetricsAggregation(
+		executor.Kubectl,
+		openperouter.Namespace,
+		labelSelector,
+		metrics.DefaultMemoryConvergenceConfig(),
+	)
+	Expect(err).NotTo(HaveOccurred(), "metrics did not stabilize for %s", labelSelector)
+	return result
+}
+
+func buildResources(tc scaleTestCase) config.Resources {
+	switch tc.scenario {
+	case l2l3vni:
+		l3vnis, l2vnis := generateL3VNIsWithL2VNIs(tc.vniCount, openperouter.Namespace, tc.hostMaster)
+		return config.Resources{
+			L3VNIs: l3vnis,
+			L2VNIs: l2vnis,
+		}
+	default:
+		return config.Resources{
+			L2VNIs: generateL2VNIs(tc.vniCount, openperouter.Namespace, tc.hostMaster),
+		}
+	}
+}
+
+func waitForVNIsGone(cli crclient.Client) {
+	Eventually(func(g Gomega) int {
+		l2list := &v1alpha1.L2VNIList{}
+		g.Expect(cli.List(context.Background(), l2list, crclient.InNamespace(openperouter.Namespace))).To(Succeed())
+		l3list := &v1alpha1.L3VNIList{}
+		g.Expect(cli.List(context.Background(), l3list, crclient.InNamespace(openperouter.Namespace))).To(Succeed())
+		return len(l2list.Items) + len(l3list.Items)
+	}, 3*time.Minute, time.Second).Should(BeZero())
+}
+
+func generateL2VNIs(count int, namespace, bridgeType string) []v1alpha1.L2VNI {
+	const baseVNI = 1000
+
+	vnis := make([]v1alpha1.L2VNI, count)
+	for i := 0; i < count; i++ {
+		vrfName := fmt.Sprintf("vrf%03d", i+1)
+		vnis[i] = v1alpha1.L2VNI{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("l2vni-%03d", i+1),
+				Namespace: namespace,
+			},
+			Spec: v1alpha1.L2VNISpec{
+				VRF:        ptr.To(vrfName),
+				VNI:        uint32(baseVNI + i + 1),
+				HostMaster: newHostMaster(bridgeType),
+			},
+		}
+	}
+	return vnis
+}
+
+func generateL3VNIsWithL2VNIs(count int, namespace, bridgeType string) ([]v1alpha1.L3VNI, []v1alpha1.L2VNI) {
+	const baseL3VNI = 2000
+	const baseL2VNI = 3000
+
+	l3vnis := make([]v1alpha1.L3VNI, count)
+	l2vnis := make([]v1alpha1.L2VNI, count)
+
+	for i := 0; i < count; i++ {
+		vrfName := fmt.Sprintf("vrf%03d", i+1)
+
+		l3vnis[i] = v1alpha1.L3VNI{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("l3vni-%03d", i+1),
+				Namespace: namespace,
+			},
+			Spec: v1alpha1.L3VNISpec{
+				VRF: vrfName,
+				VNI: uint32(baseL3VNI + i + 1),
+			},
+		}
+
+		l2vnis[i] = v1alpha1.L2VNI{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("l2vni-%03d", i+1),
+				Namespace: namespace,
+			},
+			Spec: v1alpha1.L2VNISpec{
+				VRF:        ptr.To(vrfName),
+				VNI:        uint32(baseL2VNI + i + 1),
+				HostMaster: newHostMaster(bridgeType),
+			},
+		}
+	}
+	return l3vnis, l2vnis
+}
+
+func newHostMaster(bridgeType string) *v1alpha1.HostMaster {
+	switch bridgeType {
+	case v1alpha1.LinuxBridge:
+		return &v1alpha1.HostMaster{
+			Type:        v1alpha1.LinuxBridge,
+			LinuxBridge: &v1alpha1.LinuxBridgeConfig{AutoCreate: true},
+		}
+	case v1alpha1.OVSBridge:
+		return &v1alpha1.HostMaster{
+			Type:      v1alpha1.OVSBridge,
+			OVSBridge: &v1alpha1.OVSBridgeConfig{AutoCreate: true},
+		}
+	default:
+		return nil
+	}
+}
--- a/internal/frr/vtysh/vtysh.go
+++ b/internal/frr/vtysh/vtysh.go
@@ -3,14 +3,24 @@
 package vtysh
 
 import (
+	"context"
 	"os/exec"
+	"time"
 )
+
+const DefaultTimeout = 10 * time.Second
 
 type Cli func(args string) (string, error)
 
-func Run(args string) (string, error) {
-	out, err := exec.Command("/usr/bin/vtysh", "-c", args).CombinedOutput()
-	return string(out), err
+func NewCLIWithTimeout(timeout time.Duration) Cli {
+	return func(args string) (string, error) {
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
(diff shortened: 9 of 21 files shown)


e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,191 @@
...
+	"time"
+)
+
+var (
+	cpuRegex = regexp.MustCompile(`^(\d+)(m)?$`)
+	memRegex = regexp.MustCompile(`^(\d+)(Ki|Mi|Gi)?$`)
+)
+
+// PodMetrics represents CPU and memory metrics for a single pod.
+type PodMetrics struct {
```
fedepaol, line 20: metrics.PodMetrics doesn't read well, I'd just call this Pod

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,191 @@
...
+	TestType             string        `json:"test_type"` // "l2vni_only" or "l3vni_with_l2vni"
+	RouterPodMetrics     []PodMetrics  `json:"router_pod_metrics"`
+	ControllerPodMetrics []PodMetrics  `json:"controller_pod_metrics"`
+	CollectionTime       time.Time     `json:"collection_time"`
+	Duration             time.Duration `json:"duration"`
+}
+
+// CollectPodMetrics uses kubectl top to collect metrics for pods matching the label.
+// If metrics-server is not available, it returns empty metrics with pod names only.
+func CollectPodMetrics(kubectl, namespace, labelSelector string) ([]PodMetrics, error) {
```
fedepaol, line 40: ForPod -> metrics.ForPod

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,191 @@
...
+		// metrics-server might not be available, try to get pod names at least
+		return collectPodNamesOnly(kubectl, namespace, labelSelector)
+	}
+
+	return parseKubectlTopOutput(string(out), namespace)
+}
+
+// collectPodNamesOnly gets pod names when metrics-server is not available.
+// Returns PodMetrics with zero CPU/memory values.
+func collectPodNamesOnly(kubectl, namespace, labelSelector string) ([]PodMetrics, error) {
```
fedepaol, line 53: the name of this is misleading, as it doesn't collect pod names only but the metrics too (albeit empty). Let's call this collectEmptyMetrics or just emptyMetrics

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,191 @@
...
+		return val, nil
+	case "Gi":
+		return val * 1024, nil
+	default:
+		return val / (1024 * 1024), nil // Bytes to MiB
+	}
+}
+
+// IsMetricsServerAvailable checks if metrics-server is available in the cluster.
+func IsMetricsServerAvailable(kubectl string) bool {
```
fedepaol, line 162: IsServerAvailable

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+		err := Updater.CleanAll()
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Setting up underlay configuration")
+		err = Updater.Update(config.Resources{
+			Underlays: []v1alpha1.Underlay{infra.Underlay},
+		})
+		Expect(err).NotTo(HaveOccurred())
+
+		By("Waiting for router pods to be ready")
```
fedepaol, line 57: what's the purpouse of this? If it's because we are CleanAll-ing, it's racy. Otherwise it's useless.
I think tests are written in such a way that when you run one, you have the cluster ready to be tested

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+		nodes, err := k8s.GetNodes(cs)
+		Expect(err).NotTo(HaveOccurred())
+
+		routerPods, err := k8s.PodsForLabel(cs, openperouter.Namespace, routerLabelSelector)
+		Expect(err).NotTo(HaveOccurred())
+
+		controllerPods, err := k8s.PodsForLabel(cs, openperouter.Namespace, controllerLabelSelector)
+		Expect(err).NotTo(HaveOccurred())
+
+		metricsAvailable := metrics.IsMetricsServerAvailable(executor.Kubectl)
```
fedepaol, line 71: I think this is unnecessary. Whoever runs the metrics tests must know if the metrics server is availabe (and we must document we need it!)

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+		}
+	}
+
+	err := Updater.Update(resources)
+	Expect(err).NotTo(HaveOccurred())
+
+	By("Waiting for pods to stabilize")
+	waitForPodsReady(cs, openperouter.Namespace, routerLabelSelector)
+	waitForPodsReady(cs, openperouter.Namespace, controllerLabelSelector)
+	time.Sleep(stabilizationDelay)
```
fedepaol, line 379: we are waiting for pods to be ready, is this sleep necessary? Also, waiting for them to ready might not be necessary at all

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+	testDescription := describeScaleTest(tc)
+
+	By(fmt.Sprintf("Collecting baseline metrics before %s", testDescription))
+	baselineMetrics := collectAllMetrics()
+
+	By(fmt.Sprintf("Creating %s", testDescription))
+	startTime := time.Now()
+
+	var resources config.Resources
+	if tc.l3VNICount > 0 {
```
fedepaol, line 360: instead of relying on the count, can we make this an explicit parameter of the table?

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+
+	By("Waiting for pods to stabilize")
+	waitForPodsReady(cs, openperouter.Namespace, routerLabelSelector)
+	waitForPodsReady(cs, openperouter.Namespace, controllerLabelSelector)
+	time.Sleep(stabilizationDelay)
+
+	By("Collecting scaled metrics")
+	scaledMetrics := collectAllMetrics()
+	scaledMetrics.VNICount = tc.l2VNICount
+	if tc.l3VNICount > 0 {
```
fedepaol, line 384: same here, let's be driven by a parameter

internal/frr/vtysh/vtysh.go
```diff
@@ -3,13 +3,22 @@
...
 import (
+	"context"
 	"os/exec"
+	"time"
 )
 
+// vtyshTimeout is the maximum time to wait for a vtysh command to complete.
+// Under heavy load (e.g. hundreds of VRFs), FRR daemons can be slow to respond
+// to vtysh IPC, so we use a generous timeout to avoid blocking indefinitely.
+const vtyshTimeout = 10 * time.Second
```
fedepaol, line 14: can we make this larger / customizable?

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// CollectPodMetrics uses kubectl top to collect metrics for pods matching the label.
+// If metrics-server is not available, it returns empty metrics with pod names only.
+func CollectPodMetrics(kubectl, namespace, labelSelector string) ([]PodMetrics, error) {
+	args := []string{"top", "pods", "-n", namespace, "-l", labelSelector, "--no-headers"}
+	out, err := exec.Command(kubectl, args...).CombinedOutput()
+	if err != nil {
+		// metrics-server might not be available, try to get pod names at least
+		return collectPodNamesOnly(kubectl, namespace, labelSelector)
```
fedepaol, line 45: I'd rather explode here as somebody requested to run these tests on an invalid cluster

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+		report.PrintConsole()
+
+		By("Writing scale test report to JSON file")
+		reportPath := filepath.Join(ReportPath, "scale-test-report.json")
+		err = report.WriteJSON(reportPath)
+		Expect(err).NotTo(HaveOccurred())
+		GinkgoWriter.Printf("Scale test report written to: %s\n", reportPath)
+	})
+
+	DescribeTable("VNI Scale Measurements",
```
fedepaol, line 99: even though having a graph to see how the memory / cpu goes with the number, I feel like that for CI we could / should settle only to a given number and check if the output for that number is:

- stable on multiple runs
- regressing across prs

the graph is probably nice to have when we investigate the breaking point if any, but for CI imo it makes sense only to understand if we regress

e2etests/tests/scale.go
maiqueb: adding yet another parameter is imho redundant, and confusing to the person writing tests since now they have two parameters that must be aligned. 

I fail to see the advantage it would bring to the table - would you explain ?

internal/frr/vtysh/vtysh.go
maiqueb: sure. help me think - config map ? CLI param ? ...

e2etests/tests/scale.go
maiqueb: point is we can get the artifacts from CI and get a visual representation of how good / bad it is. 

imho that adds some value. 

I'm happy to be rid of this if you prefer.

e2etests/tests/scale.go
fedepaol: here we are implementing two different class of tests based on the number of l3vnis / l2vnis. 

Imo, we should explicit about what we want to test in each entry instead of having the behavior as a side effect of the input parameter, to the point that it probably makes sense to split this table test in two, where the l2vni-only doesn't even have the l3vni parameter.

internal/frr/vtysh/vtysh.go
fedepaol: I'd say cli param that is driven by helm / operator - in case we need to address it in production

e2etests/tests/scale.go
fedepaol: but will the graph say how good / bad is? What matters (I think) is if we regress once we have a baseline.

I am fine with leaving it, I'd like to understand what's the value provided by having it in CI (vs, being able to produce it after we find a performance regression)

e2etests/tests/scale.go
maiqueb: > here we are implementing two different class of tests based on the number of l3vnis / l2vnis. 

I honestly don't agree - to me, this is the same class of test: I am measuring how much CPU / memory it costs those resources. But I also think it's not farfetched to have 2 tables instead. I'll refactor the test to have 2 tables. I just hope there isn't much duplicated code ... 

I found it obvious that when testing L3 VNI + L2 VNI we were wrapping the L3 VNI over the L2 VNI, but, I could be wrong. I won't die on this hill.

fedepaol: > I found it obvious that when testing L3 VNI + L2 VNI we were wrapping the L3 VNI over the L2 VNI, but, I could be wrong. I won't die on this hill.

yep, but what's bugging me is that we have effectively two implementations hidden here and driven by the input parameters. That's not obvious at all looking at the table entries. It's not critical and I won't hold the fort either, it was just the impression I got by reading this.

maiqueb: I never understood your argument, tbh, but, I've split this into 4 different tables, based on the following attributes:
- L2 VNI linux bridge
- L2 VNI ovs bridge
- l2 + l3 VNI linux bridge
- l2 + l3 VNI ovs bridge

What made me do the change even when not getting your argument was that with this split, if one test fails, the other configurations will still run.

e2etests/tests/scale.go
maiqueb: > If it's because we are CleanAll-ing, it's racy.

yes, I need to ensure the routers are ready after the underlay removal.

Racy how ? Can you add more info ? How do you work-around it elsewhere ? I've seen the pattern of waiting for a roll out after a destructive update elsewhere in this repo for sure. 

What do you suggest ?..

fedepaol: you remove the underlay (if any), create a new one and check for pods to be ready. If the removal of old pods is slow, you might end up checking for the old instance and assume it's ready.

What I _think_ we do elsewhere is to assume we are starting from a clean slate, we save the old pods as "routerPods", we create the underlay and we clean up all in AfterAll, ensuring pods are rolled out, like here https://github.com/openperouter/openperouter/blob/main/e2etests/tests/evpn_l2.go#L98

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/tests/scale.go
maiqueb: Done.

internal/frr/vtysh/vtysh.go
maiqueb: Done.

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/tests/scale.go
maiqueb: Is it better now @fedepaol ?

e2etests/tests/scale.go
maiqueb: it's a result of cleaning the underlay. resolving.

e2etests/tests/scale.go
maiqueb: Removed it.

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		routerPods, err := k8s.PodsForLabel(cs, openperouter.Namespace, routerLabelSelector)
+		Expect(err).NotTo(HaveOccurred())
+
+		controllerPods, err := k8s.PodsForLabel(cs, openperouter.Namespace, controllerLabelSelector)
+		Expect(err).NotTo(HaveOccurred())
+
+		metricsAvailable := metrics.IsMetricsServerAvailable(executor.Kubectl)
+		if !metricsAvailable {
```
qinqon: I think we should make this mandatory, with Expect at the beginnig  at BeforeAll, like doing scale testing without CPU/memory does not make much sense.

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+	DescribeTable("VNI Scale Measurements",
+		func(tc scaleTestCase) {
+			runScaleTest(cs, tc, report)
+		},
+		// L2VNI only tests - Linux Bridge
+		Entry("L2VNI only: 1 VNI, linux-bridge", scaleTestCase{
+			hostMaster: v1alpha1.LinuxBridge,
+			l2VNICount: 1,
+			l3VNICount: 0,
+		}),
```
qinqon: nit: you can use Ginkgo dynamic test description to compose the test name, so you don't have to harcode what you are using

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+		Entry("L2VNI only: 50 VNIs, linux-bridge", scaleTestCase{
```
qinqon: I don't think having all this tests  make sense, you just need a pair of this per VNI type
range: 0-600
step: 50
So all this tests will be something like 
```golang
Entry("L2VNI only: from 1 to 600 with 50 size steps of VNIs, linux-bridge", scaleTestCase{
			hostMaster: v1alpha1.LinuxBridge,
			l2VNI: Range {
                Min: 1
                Max: 600,
                Step: 50
		})

```

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+			l2VNICount: 550,
+			l3VNICount: 0,
+		}),
+		Entry("L2VNI only: 600 VNIs, ovs-bridge", scaleTestCase{
+			hostMaster: v1alpha1.OVSBridge,
+			l2VNICount: 600,
+			l3VNICount: 0,
+		}),
+		// L3VNI with L2VNI tests - Linux Bridge
+		Entry("L3VNI+L2VNI: 1 pair, linux-bridge", scaleTestCase{
```
qinqon: ditto

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+			hostMaster: v1alpha1.LinuxBridge,
+			l2VNICount: 550,
+			l3VNICount: 550,
+		}),
+		Entry("L3VNI+L2VNI: 600 pairs, linux-bridge", scaleTestCase{
+			hostMaster: v1alpha1.LinuxBridge,
+			l2VNICount: 600,
+			l3VNICount: 600,
+		}),
+		// L3VNI with L2VNI tests - OVS Bridge
```
qinqon: ditto

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+			hostMaster: v1alpha1.OVSBridge,
+			l2VNICount: 550,
+			l3VNICount: 550,
+		}),
+		Entry("L3VNI+L2VNI: 600 pairs, ovs-bridge", scaleTestCase{
+			hostMaster: v1alpha1.OVSBridge,
+			l2VNICount: 600,
+			l3VNICount: 600,
+		}),
+	)
```
qinqon: Don't we want to add scale test for L3VNI without L2VNI ?

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+})
+
+func runScaleTest(cs clientset.Interface, tc scaleTestCase, report *metrics.ScaleReport) {
+	testDescription := describeScaleTest(tc)
```
qinqon: Can we use `CurrentSpecReport().FullText()` instead of custom test descriptor ?

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+		resources = config.Resources{
+			L2VNIs: l2vnis,
+		}
+	}
+
+	err := Updater.Update(resources)
+	Expect(err).NotTo(HaveOccurred())
+
+	By("Waiting for pods to stabilize")
+	waitForPodsReady(cs, openperouter.Namespace, routerLabelSelector)
```
qinqon: This is not right, you can be facing a false possitive if something get broken and pods never get updated, you need to retrive tiemstamp/pods before Update call and use it to really verfiy that pods are rollout and updated.

Like we do at other parts with 
```
openperouter.DaemonsetRolled(routers, newRouters)
```

e2etests/tests/scale.go
qinqon: if we are waiting for pods why do we have to also wait for this ? maybe we should check something else instead of just add time ?

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+		scaledMetrics.TestType = fmt.Sprintf("l2vni_only_%s", tc.hostMaster)
+	}
+	scaledMetrics.Duration = time.Since(startTime)
+
+	report.AddDataPoint(baselineMetrics, scaledMetrics)
+
+	By("Cleaning up VNIs")
+	err = Updater.CleanButUnderlay()
+	Expect(err).NotTo(HaveOccurred())
+	time.Sleep(cleanupDelay)
```
qinqon: ditto, maybe we should check something else ? statabilization delays are hard to make them stable.

e2etests/tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+	Expect(err).NotTo(HaveOccurred())
+
+	return &metrics.ScaleMetrics{
+		RouterPodMetrics:     routerMetrics,
+		ControllerPodMetrics: controllerMetrics,
+		CollectionTime:       time.Now(),
+	}
+}
+
+func waitForPodsReady(cs clientset.Interface, namespace, labelSelector string) {
```
qinqon: As said why we don't use the current infra we have at e2e tests ? also the implementation is not right it can have false possitive if pod didn't change

e2etests/scale_tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+			},
+			Spec: v1alpha1.L2VNISpec{
+				VRF:        ptr.To(vrfName),
+				VNI:        uint32(baseVNI + i + 1),
+				HostMaster: newHostMaster(bridgeType),
+			},
+		}
+	}
+	return vnis
+}
```
qinqon, line 245: Do we need to add some random bits so so they don't collide between tests ? I suppose not since running scale test in parallel does not make much sense.

e2etests/scale_tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("l2vni-%03d", i+1),
+				Namespace: namespace,
+			},
+			Spec: v1alpha1.L2VNISpec{
+				VRF:        ptr.To(vrfName),
+				VNI:        uint32(baseL2VNI + i + 1),
+				HostMaster: newHostMaster(bridgeType),
+			},
+		}
```
qinqon, line 278: Does it make sense to replace this with a call to generateL2VNIs somehow refactoring both functions ?

e2etests/scale_tests/scale.go
```diff
@@ -0,0 +1,534 @@
...
+	return l3vnis, l2vnis
+}
+
+// newHostMaster creates a HostMaster with AutoCreate enabled for the specified bridge type.
+func newHostMaster(bridgeType string) *v1alpha1.HostMaster {
+	switch bridgeType {
+	case v1alpha1.LinuxBridge:
+		return &v1alpha1.HostMaster{
+			Type:        v1alpha1.LinuxBridge,
+			LinuxBridge: &v1alpha1.LinuxBridgeConfig{AutoCreate: true},
```
qinqon, line 288: Does it make sense to add scale test with pre-existing bridge ? for example if all the vnis connects to the same ovs bridge we may test how memory/CPU works for all the flows at the same bridge, not sure about that though.

config/all-in-one/crio.yaml
```diff
@@ -1329,9 +1329,6 @@ spec:
             port: webhook-server
             scheme: HTTPS
         resources:
-          limits:
```
qinqon, line 1332: Shouldn't this be a special type of deployment for scale test instead of one size fits all ?

.github/workflows/ci.yaml
```diff
@@ -394,6 +394,22 @@ jobs:
...
+            --output-dir /tmp/kind_logs/plots
+
+      - name: Upload scale test plots
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: scale-test-plots
+          path: /tmp/kind_logs/plots/
+          retention-days: 30
+
```
qinqon: I would integrate this at a Ginkgo Reporter so we can generate them without ginkgo actions, the gitgo action just retrieve the generated lots,

Looks like there is also a popular golang ploting thing https://github.com/gonum/plot in case we don't want to call python from the ginkgo reporter.

e2etests/tests/scale.go
```diff
@@ -96,255 +103,77 @@ var _ = Describe("VNI Scale Tests", Ordered, Label("scale"), func() {
...
+			Entry("50 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 50}),
+			Entry("150 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 150}),
+			Entry("200 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 200}),
+			Entry("250 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 250}),
+			Entry("300 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 300}),
+			Entry("350 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 350}),
+			Entry("400 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 400}),
+			Entry("450 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 450}),
+			Entry("500 VNIs", scaleTestCase{hostMaster: v1alpha1.LinuxBridge, scenario: l2vni, vniCount: 500}),
+		)
```
qinqon: As commented if we introduce the "step" concept we can have just one test per group

e2etests/scale_tests/scale.go
```diff
@@ -22,11 +22,18 @@ import (
...
 
-// scaleTestCase defines the parameters for a single scale test entry.
+type scenarioUnderTest int
+
+const (
+	l2vni = iota
+	l3vni
+	l2l3vni
+)
+
```
qinqon, line 31: nit: Bitmask are not worth here right ? 

```golang

const (
    l2vni    Byte = 1 << iota
    l3vni     
)
scenario = l2vni
...
scenario = l3vni
..
scenario = l2vni | l3vni
```

charts/openperouter/templates/router.yaml
```diff
@@ -177,7 +177,10 @@ spec:
         {{- with .Values.openperouter.logLevel }}
         - --loglevel={{ . }}
         {{- end }}
-        securityContext: 
+        {{- with .Values.openperouter.frr.reloader.vtyshTimeout }}
```
qinqon, line 180: Shoudn't this kind of things implemented at different PRs ? at the end we are implementing something at product at the same scale tests, not sure if that's ok.

I understand is super convenient, so you test the product fixes at the same PR you are creating the tests so they pass when the product get fixed.

internal/frr/vtysh/vtysh.go
```diff
@@ -8,18 +8,21 @@ import (
...
 
-// vtyshTimeout is the maximum time to wait for a vtysh command to complete.
-// Under heavy load (e.g. hundreds of VRFs), FRR daemons can be slow to respond
-// to vtysh IPC, so we use a generous timeout to avoid blocking indefinitely.
-const vtyshTimeout = 10 * time.Second
+const DefaultTimeout = 10 * time.Second
 
 type Cli func(args string) (string, error)
 
+func RunWithTimeout(timeout time.Duration) Cli {
```
qinqon: This is not "RunWithTimeout" this is "NewCLI" or "NewCLIWithTimeout"

Like RunWithTimeout would be 
```golang
func RunWithTimeout(args string, timeout time.Duraiton) (string, error) {
	return RunWithTimeout(timeout)(args)
}
```

cmd/reloader/main.go
```diff
@@ -53,7 +53,8 @@ func main() {
 	flag.StringVar(&args.unixSocket, "unixsocket", "", "Unix socket path to listen on")
 	flag.StringVar(&args.logLevel, "loglevel", "info", "The log level of the process")
 	flag.StringVar(&args.frrConfigPath, "frrconfig", "/etc/frr/frr.conf", "The path the frr configuration is at")
-	flag.DurationVar(&args.vtyshTimeout, "vtysh-timeout", vtysh.DefaultTimeout, "Timeout for vtysh commands used in health checks")
+	flag.DurationVar(&args.vtyshTimeout, "vtysh-timeout", vtysh.DefaultTimeout,
```
qinqon, line 56: super nit: I am not sure this change belongs to this commit

e2etests/pkg/metrics/report.go
```diff
@@ -146,10 +145,6 @@ func (r *ScaleReport) PrintConsole() {
 		r.Environment.NodeCount,
 		r.Environment.RouterPodCount,
 		r.Environment.ControllerPodCount)
-	if !r.Environment.MetricsServerAvailable {
```
qinqon: If we want to make metrics server mandatory why don't we assert them directly like mine comment on the first commit ? `Expect(r.Environment.MetricsServerAvailable).To(BeTruE())`

operator/bindata/deployment/openperouter/values.yaml
```diff
@@ -63,6 +63,7 @@ openperouter:
     resources: {}
     reloader:
       resources: {}
+      # -- Timeout for vtysh commands used in health checks. Increase under heavy VNI load. Defaults to 10s.
```
qinqon, line 66: super nit: not belonging here.

.github/workflows/ci.yaml
```diff
@@ -292,3 +292,115 @@ jobs:
         uses: ./.github/workflows/composite/collectlogs
         with:
           artifact-name: kind-logs-${{ matrix.deployment }}
+
+  scale-tests:
```
qinqon, line 297: I don't see the `run-scale-tests` label filtering here

e2etests/tests/scale.go
```diff
@@ -0,0 +1,359 @@
...
+
+	err := Updater.Update(resources)
+	Expect(err).NotTo(HaveOccurred())
+
+	By("Waiting for pods to be ready")
+	waitForPodsReady(cs, openperouter.Namespace, routerLabelSelector)
+	waitForPodsReady(cs, openperouter.Namespace, controllerLabelSelector)
+
+	By("Collecting scaled metrics")
+	scaledMetrics := collectAllMetrics()
```
qinqon: Is pod's readiness enough signal to know if we can start collecting ? I understand for big scenario it takes times for stuff to be configured, maybe we should check something else (I don't like stabilization time).

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,155 @@
...
+			CPUCores:  cpu,
+			MemoryMB:  mem,
+			Timestamp: time.Now(),
+		})
+	}
+
+	return metrics, nil
+}
+
+func parseCPU(s string) (float64, error) {
```
qinqon: does it merits some unit testing ?

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse CPU value %q: %w", matches[1], err)
+	}
+	if len(matches) > 2 && matches[2] == "m" {
+		return val, nil // Already in millicores
+	}
+	return val * 1000, nil // Convert cores to millicores
+}
+
+func parseMemory(s string) (float64, error) {
```
qinqon: does it merits some unit testing ?

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,155 @@
...
+	CPUCores  float64   `json:"cpu_millicores"` // In millicores (e.g., 250 = 250m)
+	MemoryMB  float64   `json:"memory_mb"`      // In megabytes
+	Timestamp time.Time `json:"timestamp"`
+}
+
+// ScaleMetrics represents metrics for a scale test run.
+type ScaleMetrics struct {
+	VNICount       int           `json:"vni_count"`
+	TestType       string        `json:"test_type"` // "l2vni_only" or "l3vni_with_l2vni"
+	RouterPod      []Pod         `json:"pod_metrics"`
```
qinqon: nit: Shouldn't pod_metrics be router_pod or RouterPod be PodMetrics?

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,155 @@
...
+var (
+	cpuRegex = regexp.MustCompile(`^(\d+)(m)?$`)
+	memRegex = regexp.MustCompile(`^(\d+)(Ki|Mi|Gi)?$`)
+)
+
+// Pod represents CPU and memory metrics for a single pod.
+type Pod struct {
+	PodName   string    `json:"pod_name"`
+	Namespace string    `json:"namespace"`
+	CPUCores  float64   `json:"cpu_millicores"` // In millicores (e.g., 250 = 250m)
```
qinqon: This is not Cores it's milliseconds

e2etests/pkg/metrics/report.go
```diff
@@ -0,0 +1,174 @@
...
+		return fmt.Errorf("failed to marshal report: %w", err)
+	}
+	if err := os.WriteFile(path, data, 0644); err != nil {
+		return fmt.Errorf("failed to write report to %s: %w", path, err)
+	}
+	return nil
+}
+
+// PrintConsole prints a summary of the report to stdout.
+func (r *ScaleReport) PrintConsole() {
```
qinqon: Shoulnd't we use GinkgoWriter insetead of fmt ?

e2etests/pkg/metrics/report.go
```diff
@@ -0,0 +1,174 @@
...
+// ScaleReport contains all collected metrics from scale tests.
+type ScaleReport struct {
+	TestRunTime time.Time        `json:"test_run_time"`
+	Environment EnvironmentInfo  `json:"environment"`
+	DataPoints  []ScaleDataPoint `json:"data_points"`
+}
+
+// EnvironmentInfo captures information about the test environment.
+type EnvironmentInfo struct {
+	KubernetesVersion  string `json:"kubernetes_version,omitempty"`
```
qinqon: This is never used.

e2etests/pkg/metrics/report.go
```diff
@@ -0,0 +1,174 @@
...
+
+// ScaleReport contains all collected metrics from scale tests.
+type ScaleReport struct {
+	TestRunTime time.Time        `json:"test_run_time"`
+	Environment EnvironmentInfo  `json:"environment"`
+	DataPoints  []ScaleDataPoint `json:"data_points"`
+}
+
+// EnvironmentInfo captures information about the test environment.
+type EnvironmentInfo struct {
```
qinqon: Is missing the field that we read at the ploting python script "metrics_server_available"

qinqon (changes requested):
Looks pretty good, but we have to be sure that this is not running at every PR, I think we are missing label filtering.

e2etests/tests/scale.go
maiqueb: tks but I do prefer to have specific items for each test I want to run. For instance, if I want to just execute the test with 300 VNIs, how do I do it using your semantics ? Using mine I know how.

e2etests/tests/scale.go
maiqueb: ditto

e2etests/tests/scale.go
maiqueb: a follow up PR can do that.

e2etests/tests/scale.go
maiqueb: I guess so. What is the problem you're trying to solve here ? What are the advantages ?

e2etests/tests/scale.go
maiqueb: you're commenting on outdated code ... not sure why.

e2etests/tests/scale.go
maiqueb: I'll remove this, good catch.

e2etests/scale_tests/scale.go
maiqueb: if running scale tests in parallel doesn't make sense why would we want to do it, or prepare the test for it ? 

Let's make our lives simple for once.

e2etests/scale_tests/scale.go
maiqueb: IDK. I'd defer that to a follow-up.

e2etests/scale_tests/scale.go
maiqueb: maybe. For now, I don't see a reason to. But we could for sure.

config/all-in-one/crio.yaml
maiqueb: good point.

e2etests/tests/scale.go
maiqueb: I don't want to use step for the reasons explained above.

e2etests/scale_tests/scale.go
maiqueb: what's the advantage ? you save an iota number ? 
I personally feel what we have now is more explicit.

charts/openperouter/templates/router.yaml
maiqueb: I am good with this iteration. 

Do you foresee any issues with it ?

e2etests/pkg/metrics/metrics.go
maiqueb: it can.

internal/frr/vtysh/vtysh.go
maiqueb: OK, I can live with NewCLIWithTimeout / NewCLI

.github/workflows/ci.yaml
maiqueb: The plot thing is up for grabs. Let's resolve https://github.com/openperouter/openperouter/pull/189#discussion_r3111831487 first, then make a call about this one.

e2etests/tests/scale.go
maiqueb: Removed it. This lane only runs the scale tests, thus, imho, there's no need to actually be removing and re-provisioning the underlay.

e2etests/tests/scale.go
qinqon: > tks but I do prefer to have specific items for each test I want to run. For instance, if I want to just execute the test with 300 VNIs, how do I do it using your semantics ? Using mine I know how.

You can still do it if you want a specific test apart of the "dynamic" ones, since step is 0 by default
```golang
Range{
   Max: 300
   Min: 300
}
```

e2etests/tests/scale.go
qinqon: Why do we want our own test description function when we have a Ginkgo mechanism for that ? that's why we use a framework.

e2etests/tests/scale.go
maiqueb: I don't understand your answer, sorry.

e2etests/tests/scale.go
maiqueb: ok, let's do this.

e2etests/tests/scale.go
qinqon: Puff totally, I didn't refresh my clone, let me take another look

e2etests/tests/scale.go
maiqueb: Removed this code.

Since the lane ONLY executes these scale tests - I'd say it doesn't make any sense for this lane to run other tests - we can avoid doing anything. Just ensure the underlay is provisioned, and go forward.

e2etests/scale_tests/scale.go
qinqon: Can you add possible follow ups at the PR description ? or maybe create Issues ?

e2etests/tests/scale.go
maiqueb: Resolving it.

e2etests/tests/scale.go
maiqueb: Done.

e2etests/tests/scale.go
qinqon: I think the best of both worlds is to generate the graph only if we find regressions so we have debug artifacts also I think it should be generated by the ginkgo test suite not by the github action.

.github/workflows/ci.yaml
qinqon: > The plot thing is up for grabs. Let's resolve [#189 (comment)](https://github.com/openperouter/openperouter/pull/189#discussion_r3111831487) first, then make a call about this one.

Added comment there with my opinion.

e2etests/tests/scale.go
qinqon: you can have both things stepped and non stepped tests.

e2etests/scale_tests/scale.go
qinqon: Maybe an improvement since we don't care mutch for performance here, is to use strings instead of integers so you can print them, but this is just an opinion.

charts/openperouter/templates/router.yaml
qinqon: As far as we keep them at separate commits and comment at the PR description that we are also touching the product then we are ok.

maiqueb: OK, then it's a good point. I'll ensure to add a release note.

e2etests/tests/scale.go
qinqon: I mean that you can have the stepper and non stepper version of the test instead of having one test per step 

```golang
// Runs a test with 300 VNIs
Entry("L2VNI only: 300 VNIs, linux-bridge", scaleTestCase{
			hostMaster: v1alpha1.LinuxBridge,
			l2VNI: Range {
                Min: 300
                Max: 300,
		})
```

e2etests/pkg/metrics/metrics.go
qinqon: And will ... ?

e2etests/tests/scale.go
maiqueb: imho that's just complicating it, without any benefit. 

since IDK which test I may want to call on its own, I'd have to duplicate all. I don't see any practical advantage of doing it this way.

e2etests/pkg/metrics/metrics.go
maiqueb: I've removed all these files/structs, and I'm trying out the `gMeasure` structs gomega provides.

e2etests/pkg/metrics/report.go
maiqueb: this is gone from the code base.

e2etests/pkg/metrics/report.go
maiqueb: this is gone right ?

e2etests/pkg/metrics/report.go
maiqueb: I think I've removed these structs

e2etests/tests/scale.go
maiqueb: is it OK now ?

e2etests/tests/scale.go
maiqueb: Done. Confirm this is what you expected @qinqon

config/all-in-one/crio.yaml
maiqueb: Done. Please confirm this meets your expectations @qinqon

e2etests/tests/scale.go
maiqueb: I'd rather have something simple. I haven't yet understood the advantage brought by using the step feature you propose.

e2etests/scale_tests/scale.go
maiqueb: as long as the test case `.String()` method can output it properly, I don't think it matters much.

internal/frr/vtysh/vtysh.go
maiqueb: Done.

e2etests/pkg/metrics/report.go
maiqueb: this was removed.

e2etests/pkg/metrics/report.go
qinqon: I se the field "metrics_server_available" now, you can resolve this thread (I can't)

e2etests/tests/scale.go
qinqon: Nope, I mean the the following 
```golang
func(tc sclateTestCase ) string{
      if tc.l3VNICount == 0 {
          return "L2VNI only: %d VNI, %s", tc.l2VNICount, hostMaster"
      } else {
      ....
    },
},
Entry(nil, scaleTestCase{
			hostMaster: v1alpha1.LinuxBridge,
			l2VNICount: 1,
			l3VNICount: 0,
		}),  
    ....
```

e2etests/tests/scale.go
qinqon: Fair enough, I still see this very redudant but ok.

e2etests/tests/scale.go
qinqon: I don't see this implemented at the PR

e2etests/tests/scale.go
maiqueb: OK, now the entire suite fails in the before all if the metrics server is not available.

e2etests/tests/scale.go
maiqueb: beats me how can this be more readable. That "nil" entry as the first parameter in the `Entry` alone almost burns my eyes. 

But there you go: I think we're using it now.

charts/openperouter/templates/router.yaml
fedepaol: 👍 , as long as separate in commit and documented in release notes if needed it's not an issue imo

config/all-in-one/crio.yaml
fedepaol: not sure it worked as intended. I see the overlay still being applied to the all-in-one manifests, while I guess the intention is to use the scale overlay only for scale tests?

e2etests/pkg/k8s/nad.go
```diff
@@ -17,16 +17,25 @@ import (
 
 var nadClient *nadclientset.Clientset
 
-func init() {
-	var err error
-	config := ctrl.GetConfigOrDie()
+func initNADClient() error {
+	if nadClient != nil {
```
fedepaol, line 21: re: the changed code. It's an init function that does not depend on the environment. IMO we don't need the error handling at all as it either works or does not, there's no point in raising the error and making the surrounding code more complex.

e2etests/pkg/k8sclient/client.go
```diff
@@ -18,6 +18,6 @@ func New() *clientset.Clientset {
 	return res
 }
 
-func RestConfig() *restclient.Config {
-	return ctrl.GetConfigOrDie()
+func RestConfig() (*restclient.Config, error) {
+	return ctrl.GetConfig()
```
fedepaol, line 22: same here. This will affect all the tests so it must succeed

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+	"time"
+)
+
+var (
+	cpuRegex = regexp.MustCompile(`^(\d+)(m)?$`)
+	memRegex = regexp.MustCompile(`^(\d+)(Ki|Mi|Gi)?$`)
+)
+
+// Pod represents CPU and memory metrics for a single pod.
+type Pod struct {
```
fedepaol: PodMetrics? PodInfo?
Calling this Pod seems represent a Pod.

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+		return val, nil
+	case "Gi":
+		return val * 1024, nil
+	default:
+		return val / (1024 * 1024), nil // Bytes to MiB
+	}
+}
+
+// MetricsSummaryResult contains aggregate statistics for a slice of Pod.
+type MetricsSummaryResult struct {
```
fedepaol: metrics is already part of the package name.

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+// Pod represents CPU and memory metrics for a single pod.
+type Pod struct {
+	PodName   string    `json:"pod_name"`
+	Namespace string    `json:"namespace"`
+	CPUMillicores  float64   `json:"cpu_millicores"` // In millicores (e.g., 250 = 250m)
+	MemoryMB  float64   `json:"memory_mb"`      // In megabytes
+	Timestamp time.Time `json:"timestamp"`
+}
+
+// ForPod uses kubectl top to collect metrics for pods matching the label.
```
fedepaol: this seems to be used only internally. Let's make it private. Also, let's move private / util files to the bottom of the file

e2etests/pkg/metrics/metrics.go
fedepaol: Also, this is probably "Aggregated"? It might even make sense to split average and total in two structures

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+			if math.Abs(curr.TotalMem-prev.TotalMem) <= cfg.ToleranceMB {
+				return curr, nil
+			}
+			prev = curr
+		}
+	}
+}
+
+// SummarizeMetrics calculates aggregate statistics for a slice of Pod.
+func SummarizeMetrics(pods []Pod) MetricsSummaryResult {
```
fedepaol: let's call the pods argument podsMetrics

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+		PollInterval: 5 * time.Second,
+		Timeout:      90 * time.Second,
+		ToleranceMB:  1.0,
+	}
+}
+
+// WaitForStableMemory polls kubectl top until two consecutive total-memory
+// readings are within ToleranceMB of each other, ensuring at least one
+// metrics-server refresh has been observed.
+func WaitForStableMemory(kubectl, namespace, labelSelector string, cfg StableMemoryConfig) (MetricsSummaryResult, error) {
```
fedepaol: this does not only wait, it also returns the summary.
Let's call this FetchMetricsSummary or FetchMetricsAggregation

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+	pods, err := ForPod(kubectl, namespace, labelSelector)
```
fedepaol: we can have a local fetchPodsMetrics or something like that as we reuse this block below in the select

e2etests/pkg/k8s/nad.go
maiqueb: oh there is ... 

If you don't surface this error, when you run the test, it'll simply fail silently, and probably return something super misleading to the user. I've spent 2 hours debugging this because I was not surfacing the error.

e2etests/pkg/metrics/metrics.go
maiqueb: well ... it's on "metrics" package ... 

I think I'm just following the guidelines.

e2etests/pkg/metrics/metrics.go
maiqueb: I agree aggregated is a more common term. I'll rename it. I'll also consider splitting avg / total into 2 structs.

e2etests/pkg/metrics/metrics.go
maiqueb: sure.

e2etests/pkg/metrics/metrics.go
maiqueb: it's on the metrics pkg ... calling it pod metrics is frowned upon, according to the repo's guidelines.

e2etests/pkg/metrics/metrics.go
maiqueb: will do; FetchMetricsAggregation it is.

e2etests/pkg/metrics/metrics.go
fedepaol: I am speaking about the argument, which is local to the function

e2etests/pkg/metrics/metrics.go
fedepaol: I think we don't even consume it from outside, so we can make this private (and move it below the public bits) and call it podMetrics or metricsForPod

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,194 @@
...
+// MetricsSummaryResult contains aggregate statistics for a slice of Pod.
+type MetricsSummaryResult struct {
+	TotalCPU float64
+	TotalMem float64
+	AvgCPU   float64
+	AvgMem   float64
+}
+
+// StableMemoryConfig controls the polling behavior of WaitForStableMemory.
+type StableMemoryConfig struct {
```
fedepaol: Let's call this "ConvergenceConfig" or "MemoryConvergenceConfig", as "StableMemoryConfig" is not telling us what it's about imo

maiqueb:
@fedepaol @qinqon regarding the metrics ... 

I am thinking of stop monitoring the CPU consumption (can't get anything decent from it), and only do memory + reconcile times. 

The reconcile times are IMHO a very interesting metric - right now, it seems we are linear. Not amazing, but not bad at all. I think we should stop and re-evaluate whatever we're changing if we ever see we're diverging from that trend. 

Thoughts ?

e2etests/pkg/k8sclient/client.go
maiqueb: as said above: since this is called from an `init` block, if it doesn't return an error, it'll just explode, leaving nothing behind but the error code. 

Very hard to debug. I would prefer to move to this pattern.

e2etests/pkg/metrics/metrics.go
maiqueb: ok. that I can happily do.

e2etests/pkg/metrics/metrics.go
maiqueb: ah right. 

I'll do it.

e2etests/tests/scale.go
maiqueb: Can you look again ?

.github/workflows/ci.yaml
maiqueb: now you should.

e2etests/tests/scale.go
maiqueb: I really don't know how to answer here. I'm afraid I don't fully understand the question. 

If you have concerns share them. If you have suggestions, please suggest. 

What would you do ?

e2etests/pkg/metrics/metrics.go
maiqueb: OK, let's roll w/ MemoryConvergenceConfig

e2etests/pkg/metrics/metrics.go
maiqueb: Added it; done.

e2etests/pkg/metrics/metrics.go
maiqueb: Added it; done.

operator/bindata/deployment/openperouter/values.yaml
maiqueb: I don't follow. Here where ?..

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/pkg/metrics/metrics.go
maiqueb: done

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/pkg/metrics/metrics.go
maiqueb: Done

e2etests/pkg/metrics/metrics.go
maiqueb: ? are you sure that's what kubernetes returns via kubectl top ?

maiqueb (commented):
Reacting to the review. I plan to squash all these new commits into their proper places.

e2etests/pkg/k8s/nad.go
fedepaol: how so if we panic / getConfigOrDie?

e2etests/pkg/k8sclient/client.go
fedepaol: Fair enough, I don't want to hold on this

e2etests/pkg/k8s/nad.go
fedepaol: in any case I don't have objections

e2etests/pkg/metrics/metrics.go
maiqueb: Done. I have used a closure, since we don't need it elsewhere.

operator/bindata/deployment/openperouter/values.yaml
qinqon: humm nothing all good we can close this.

e2etests/tests/scale.go
qinqon: You have refactor, we are good now

.github/workflows/ci.yaml
```diff
@@ -292,3 +293,124 @@ jobs:
...
+
+      - name: Deploy the cluster
+        run: |
+          make deploy-scale
+
+      - name: Install metrics-server
+        env:
+          KUBECONFIG: bin/kubeconfig
+        run: |
+          kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml
```
fedepaol: not blocking, let's see if there's something else to touch in the PR, but would it make sense to have this in a makefile rule?

.github/workflows/ci.yaml
```diff
@@ -292,3 +293,124 @@ jobs:
...
+          sleep 60
+
+      - name: Run scale tests
+        run: |
+          make scale-tests
+
+      - name: Parse scale test report
+        if: ${{ always() }}
+        run: |
+          python3 hack/parse-scale-report.py /tmp/kind_logs/e2etests_suite_scale-report.json \
```
fedepaol: same, it probably makes sense to wrap this in a makefile rule

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,209 @@
...
+		result.TotalCPU += p.cpuMillicores
+		result.TotalMem += p.memoryMB
+	}
+
+	result.AvgCPU = result.TotalCPU / float64(len(podsMetrics))
+	result.AvgMem = result.TotalMem / float64(len(podsMetrics))
+	return result
+}
+
+// --- private helpers and types ---
```
fedepaol: comment not needed

e2etests/tests/scale.go
```diff
@@ -0,0 +1,296 @@
...
+			Entry(nil, scaleTestCase{hostMaster: v1alpha1.OVSBridge, scenario: l2l3vni, vniCount: 500}),
+		)
+	})
+})
+
+func runScaleTest(tc scaleTestCase, experiment *gmeasure.Experiment) {
+	testLabel := CurrentSpecReport().FullText()
+
+	By("Waiting for VNI resources to be fully removed")
+	waitForVNIsGone(Updater.Client())
```
fedepaol: shouldn't this go in aftereach?

e2etests/tests/scale.go
```diff
@@ -0,0 +1,296 @@
...
+		gmeasure.Units("MB"), gmeasure.Precision(2),
+		gmeasure.Annotation(fmt.Sprintf("%d VNIs", tc.vniCount)),
+	)
+	experiment.RecordValue(testLabel+" controller memory",
+		ctrlMem.TotalMem,
+		gmeasure.Units("MB"), gmeasure.Precision(2),
+		gmeasure.Annotation(fmt.Sprintf("%d VNIs", tc.vniCount)),
+	)
+
+	By("Cleaning up VNIs")
```
fedepaol: same, does it make sense to have it in aftereach?

e2etests/scale_tests/scale.go
```diff
@@ -0,0 +1,296 @@
...
+		err = Updater.Update(config.Resources{
+			Underlays: []v1alpha1.Underlay{infra.Underlay},
+		})
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterAll(func() {
+		By("Cleaning up all resources")
+		err := Updater.CleanAll()
+		Expect(err).NotTo(HaveOccurred())
```
fedepaol, line 82: in theory here we should wait for pods rollout. In practice we are always running this in isolation, so it's not super important. 
But now, given the fact that this has little to do with the other tests, I wonder if it'd make sense for it to be in its own suite.

fedepaol (commented):
Few comments, nothing blocking imo

internal/frr/vtysh/vtysh.go
```diff
@@ -3,14 +3,26 @@
...
+func NewCLIWithTimeout(timeout time.Duration) Cli {
+	return func(args string) (string, error) {
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+		out, err := exec.CommandContext(ctx, "/usr/bin/vtysh", "-c", args).CombinedOutput()
+		return string(out), err
+	}
+}
+
+func NewCLI(args string) (string, error) {
```
RamLavi: NewCLI looks like a constructor but it's actually a Cli function itself — the naming is inconsistent with NewCLIWithTimeout which is a constructor. Can we make NewCLI a constructor too?
```Suggestion
func NewCLI() Cli {
    return NewCLIWithTimeout(DefaultTimeout)
}
```

Then the callsite becomes 
```
cli := NewCLI()
out, err := cli("show bgp summary")
```

 and both follow the same pattern.

e2etests/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,145 @@
...
+type Pod struct {
+	PodName   string    `json:"pod_name"`
+	Namespace string    `json:"namespace"`
+	CPUMillicores  float64   `json:"cpu_millicores"` // In millicores (e.g., 250 = 250m)
+	MemoryMB  float64   `json:"memory_mb"`      // In megabytes
+	Timestamp time.Time `json:"timestamp"`
+}
+
+// ForPod uses kubectl top to collect metrics for pods matching the label.
+func ForPod(kubectl, namespace, labelSelector string) ([]Pod, error) {
```
RamLavi: sort of nit:
Could we avoid parsing `kubectl top pods` output here? Even with `--no-headers`, it is still human-readable table output and may be fragile. A more robust option is to query the Metrics API directly with `kubectl get --raw`, which returns structured JSON:

`/apis/metrics.k8s.io/v1beta1/namespaces/<namespace>/pods?labelSelector=<selector>`

That should make this easier and safer to parse.

e2etests/tests/scale.go
```diff
@@ -0,0 +1,301 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type scenarioUnderTest int
+
+const (
+	l2vni = iota
+	l3vni
```
RamLavi: l3vni is not used at all If I am not mistaken

e2etests/tests/scale.go
```diff
@@ -0,0 +1,301 @@
...
+	})
+
+	AfterAll(func() {
+		By("Cleaning up all resources")
+		err := Updater.CleanAll()
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	Describe("L2VNI only - Linux Bridge", func() {
+		DescribeTable("VNI Scale Measurements",
```
RamLavi: not sure it's totally worth it, but consider using a boilerplate helper function that will generate the Entries for all the 4 Describes (with `hostMaster`, `scenario` as input params). 
Your call.

.github/workflows/ci.yaml
maiqueb: Done.

.github/workflows/ci.yaml
maiqueb: Done.

config/scale/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
RamLavi, line 2: commit `e2etests: add scale-specific kustomize layer without resource limits` is missing your sign off.

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

e2etests/tests/scale.go
maiqueb: Done.

e2etests/tests/scale.go
maiqueb: Done.

RamLavi (commented):
PR description stale nit: says "1, 5, 10, 20, 50, and 100 VNIs" but the actual entries are 1, 50, 150, 200, 250, 300, 350, 400, 450, 500.

Just a forward-looking suggestion: I think a good followup issue from this change is to add an AI agent that would **analyze** the data, making it more easy for human eyes, and possibly detect potential problems that might originate from the PR that is now running this scale test. 
Perhaps gemini-code-review could be tweaked to do so, if we give it a valid baseline.

maiqueb:
> PR description stale nit: says "1, 5, 10, 20, 50, and 100 VNIs" but the actual entries are 1, 50, 150, 200, 250, 300, 350, 400, 450, 500.
> 
> Just a forward-looking suggestion: I think a good followup issue from this change is to add an AI agent that would **analyze** the data, making it more easy for human eyes, and possibly detect potential problems that might originate from the PR that is now running this scale test. Perhaps gemini-code-review could be tweaked to do so, if we give it a valid baseline.

Yeah, this aligns with what I have spoken w/ @qinqon out of band.

The rough idea would be to "publish" the scale tests results on each "tagged" version release, and to have that as a rough baseline. We'd then ask CI to run the scale tests on some PRs (by applying the label to it), and, we'd have the agent +1 / -1 it depending on some criteria. 

This "agentic" part would be left as a follow-up. 

Does that make sense to you @RamLavi ? 

Thoughts @fedepaol / @qinqon ?

e2etests/tests/scale.go
maiqueb: I know, this is added so the follow-up that adds scale tests for L3 VNIs only. 

If you prefer, I can drop it from this PR.

e2etests/tests/scale.go
maiqueb: I think this is good as is. We can improve this later on, if we think it would be preferable.

config/scale/kustomization.yaml
maiqueb: good catch! Added it.

qinqon:
> > PR description stale nit: says "1, 5, 10, 20, 50, and 100 VNIs" but the actual entries are 1, 50, 150, 200, 250, 300, 350, 400, 450, 500.
> > Just a forward-looking suggestion: I think a good followup issue from this change is to add an AI agent that would **analyze** the data, making it more easy for human eyes, and possibly detect potential problems that might originate from the PR that is now running this scale test. Perhaps gemini-code-review could be tweaked to do so, if we give it a valid baseline.
> 
> Yeah, this aligns with what I have spoken w/ @qinqon out of band.
> 
> The rough idea would be to "publish" the scale tests results on each "tagged" version release, and to have that as a rough baseline. We'd then ask CI to run the scale tests on some PRs (by applying the label to it), and, we'd have the agent +1 / -1 it depending on some criteria.
> 
> This "agentic" part would be left as a follow-up.
> 
> Does that make sense to you @RamLavi ?
> 
> Thoughts @fedepaol / @qinqon ?

I think a daily periodic job checking threshold is better, then if things goes bad we will have to investigate stuff merged from last day (is not that massive).

maiqueb:
> > > PR description stale nit: says "1, 5, 10, 20, 50, and 100 VNIs" but the actual entries are 1, 50, 150, 200, 250, 300, 350, 400, 450, 500.
> > > Just a forward-looking suggestion: I think a good followup issue from this change is to add an AI agent that would **analyze** the data, making it more easy for human eyes, and possibly detect potential problems that might originate from the PR that is now running this scale test. Perhaps gemini-code-review could be tweaked to do so, if we give it a valid baseline.
> > 
> > 
> > Yeah, this aligns with what I have spoken w/ @qinqon out of band.
> > The rough idea would be to "publish" the scale tests results on each "tagged" version release, and to have that as a rough baseline. We'd then ask CI to run the scale tests on some PRs (by applying the label to it), and, we'd have the agent +1 / -1 it depending on some criteria.
> > This "agentic" part would be left as a follow-up.
> > Does that make sense to you @RamLavi ?
> > Thoughts @fedepaol / @qinqon ?
> 
> I think a daily periodic job checking threshold is better, then if things goes bad we will have to investigate stuff merged from last day (is not that massive).

That can complement the release baseline, which I think is the "figures" we should try to adhere to.

maiqueb (commented):
replying to @fedepaol and @RamLavi reviews. 

As usually, I'm adding reactions to the PR review in separate commits, to get CI feedback. Depending on it, I'll squash them into their proper places in a follow-up push.

e2etests/scale_tests/scale.go
maiqueb: I'm exploring moving the scale tests to a separate suite in the latest push. 

Let's discuss this new test topology after we get CI feedback. Good suggestion !

internal/frr/vtysh/vtysh.go
maiqueb: Done.

RamLavi:
> > PR description stale nit: says "1, 5, 10, 20, 50, and 100 VNIs" but the actual entries are 1, 50, 150, 200, 250, 300, 350, 400, 450, 500.
> > Just a forward-looking suggestion: I think a good followup issue from this change is to add an AI agent that would **analyze** the data, making it more easy for human eyes, and possibly detect potential problems that might originate from the PR that is now running this scale test. Perhaps gemini-code-review could be tweaked to do so, if we give it a valid baseline.
> 
> Yeah, this aligns with what I have spoken w/ @qinqon out of band.
> 
> The rough idea would be to "publish" the scale tests results on each "tagged" version release, and to have that as a rough baseline. We'd then ask CI to run the scale tests on some PRs (by applying the label to it), and, we'd have the agent +1 / -1 it depending on some criteria.
> 
> This "agentic" part would be left as a follow-up.
> 
> Does that make sense to you @RamLavi ?

It does, with the only caveat that checking only on "tagged" version release would be too late, cause we usually do it at the end of time schedules, when everything is already stressed. We can maybe think of a better logic of "if you touch these files, then auto label for scale performance). But worth a deeper thinking..

> 
> Thoughts @fedepaol / @qinqon ?

e2etests/pkg/metrics/metrics.go
maiqueb: Done.

fedepaol:
LGTM

fedepaol:
thanks! Feel free to adjust the commit history and merge

maiqueb:
> thanks! Feel free to adjust the commit history and merge

Thank **you** ! 

Already on it ... I should be done with the squash in a bit . I'll then push, wait for CI, then merge.

maiqueb:
We will ignore the failing lane "CI / feature-needs-docs" [CI / feature-needs-docs (pull_request)](https://github.com/openperouter/openperouter/actions/runs/25377105960/job/74415059484?pr=189) since this "feature" is not really user facing. 

The only part that requires user documentation is the release notes, where we document the new attribute to define the vtysh timeout.

e2etests/tests/scale.go
RamLavi: if you say its use is coming, and it's not currently firing any linter, then I'm not against keeping it

PR #202, opened 2026-01-19
Enable arp_accept to fix slow MAC/IP route advertisement for IPv4

maiqueb:

>

/kind bug

> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

During cross-cluster VM live migration, there was a 20+ second packet loss window because FRR could not advertise MAC/IP routes until a kernel neighbor entry existed. The kernel's default arp_accept=0 setting causes it to ignore Gratuitous ARP packets (sent by libvirt when the migration finishes successfully), thus preventing neighbor entry creation from incoming GARPs.

This change enables arp_accept=1 on both 'all' and 'default' interfaces in the router namespace at startup. This allows the kernel to create neighbor entries from received GARP packets, enabling FRR to immediately advertise MAC/IP routes when a VM migrates.

Testing shows packet loss reduced from 207-210 packets (~21-29 seconds) to just 3-4 packets during live migration.

Fixes: #203

IPv6 will be handled on a different PR, should the issue reproduce for that IP family. 

**Release note**:

```release-note
Enable the arp_accept sysctl in the router network namespaces, to learn MAC mobility events from GARPs.
```

### DIFF
--- a/internal/controller/routerconfiguration/host_config.go
+++ b/internal/controller/routerconfiguration/host_config.go
@@ -52,14 +52,14 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
 		return fmt.Errorf("failed to convert config to host configuration: %w", err)
 	}
 
-	slog.InfoContext(ctx, "ensuring IPv6 forwarding")
-	if err := sysctl.EnsureIPv6Forwarding(config.targetNamespace); err != nil {
-		return fmt.Errorf("failed to ensure IPv6 forwarding: %w", err)
-	}
-
-	slog.InfoContext(ctx, "ensuring arp_accept enabled")
-	if err := sysctl.EnsureArpAccept(config.targetNamespace); err != nil {
-		return fmt.Errorf("failed to ensure arp_accept: %w", err)
+	slog.InfoContext(ctx, "ensuring sysctls")
+	if err := sysctl.Ensure(
+		config.targetNamespace,
+		sysctl.IPv6Forwarding(),
+		sysctl.ArpAcceptAll(),
+		sysctl.ArpAcceptDefault(),
+	); err != nil {
+		return fmt.Errorf("failed to ensure sysctls: %w", err)
 	}
 
 	slog.InfoContext(ctx, "setting up underlay")
--- a/internal/sysctl/sysctl.go
+++ b/internal/sysctl/sysctl.go
@@ -12,96 +12,74 @@ import (
 	"github.com/vishvananda/netns"
 )
 
-// EnsureIPv6Forwarding checks if IPv6 forwarding is enabled in the target namespace
-// and enables it if not already set to 1.
-func EnsureIPv6Forwarding(namespace string) error {
+// Sysctl represents a sysctl setting to be enabled.
+type Sysctl struct {
+	Path        string // The sysctl path under /proc/sys/
+	Description string // Human-readable description for logging
+}
+
+// Ensure enables the given sysctls in the target namespace.
+// Each sysctl is checked and only written if not already set to "1".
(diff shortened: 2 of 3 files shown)


maiqueb:
> @maiqueb since ARp is IPv4, do we know if we have to activate something similar for NDP in IPv6 ?

Opened #205 which tracks this issue for IPv6. A fix is provided in #207 .

Could you review @qinqon ?

internal/hostnetwork/forwarding_test.go
```diff
@@ -74,3 +74,90 @@ var _ = Describe("EnsureIPv6Forwarding", func() {
 		})
 	})
 })
+
+var _ = Describe("EnsureArpAccept", func() {
```
qinqon: This unit test force user to be root to be able to create namespaces, are we ok with that ?

internal/controller/routerconfiguration/host_config.go
```diff
@@ -52,7 +52,10 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
 	}
 
 	slog.InfoContext(ctx, "ensuring IPv6 forwarding")
-	if err := hostnetwork.EnsureIPv6Forwarding(config.targetNamespace); err != nil {
+	if err := hostnetwork.EnsureSysctls(
+		config.targetNamespace,
+		append(hostnetwork.IPv6ForwardingSysctl(), hostnetwork.ArpAcceptSysctls()...)...,
+	); err != nil {
```
qinqon: We are making the systctl functions return multiuple just because we have to configure ARP accept at all and at defualt, just be explicit and make individual ones so we end up with something like 
```golang
EnsureSysctl(netNS, sysctl.IPv6Forwarding, sysctl.ARPAcceptAll, sysctlARPAcceptDefalt)
```

internal/hostnetwork/forwarding.go
```diff
@@ -11,41 +11,72 @@ import (
...
+	}
+}
+
+// ArpAcceptSysctls returns the sysctl definitions for enabling arp_accept.
+// Enabling arp_accept allows the kernel to create neighbor entries from received
+// Gratuitous ARP packets, which is critical for fast EVPN MAC/IP route advertisement
+// during VM migrations.
+func ArpAcceptSysctls() []Sysctl {
+	return []Sysctl{
+		{
```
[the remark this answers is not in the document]
maiqueb: Moving this will require moving the `inNamespace` thing to another package to avoid import loops. Please name it.

internal/sysctl/sysctl.go
```diff
@@ -0,0 +1,85 @@
...
+
+// Sysctl represents a sysctl setting to be enabled.
+type Sysctl struct {
+	Path        string // The sysctl path under /proc/sys/
+	Description string // Human-readable description for logging
+}
+
+// EnsureSysctls enables the given sysctls in the target namespace.
+// Each sysctl is checked and only written if not already set to "1".
+func EnsureSysctls(namespace string, sysctls ...Sysctl) error {
```
fedepaol: if there's no real need to make this generic and accept multiple sysctls at once, I'd remove the duplication by making the implementation private and having an EnsureIPForwardingSysCtl, EnsureArpAcceptAllSysctl and so on, leveraging the private function

maiqueb: I think this goes against a sysctl suggestion by @qinqon .

And, FWIW, it's not only removing the duplication - my major concern is avoiding the `inNamespace` call, which,  would cost tapping into the network namespace once per sysctl, which is - afaiu - expensive.

PR #215, opened 2026-01-28
Integration tests: auto-discover runasroot suites and pass ginkgo.v

qinqon:
/kind bug

Dynamically find test packages with the runasroot build tag instead of
hardcoding them in the Makefile. This revealed that the bridgerefresh
test suite was not being run. Compiled test binary paths are passed as
arguments to integration_tests.sh, which detects ginkgo suites and
adds -ginkgo.v for verbose output.

**Release note**:

```release-note
NONE
```


fedepaol:
> > Not a big fan of autodetection mechanisms. How about using a ginkgo tag and using ginkgo args to skip / select these tests?
> 
> That was my first approach, but issue is that not all the tests are using ginkgo.

but these which require root do right? Otherwise the new function would not work. So in this case we can propagate ginkgo parameters?

PR #217, opened 2026-02-02
enhancement: Per node multus annotation

qinqon:

/kind design

Add enhancement to be able to configure different NAD depending on worker or control plane node role.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/enhancements/per-node-multus-annotation.md
+++ b/enhancements/per-node-multus-annotation.md
@@ -1,0 +1,443 @@
+# Per-Node Multus Network Annotation
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


qinqon:
@fedepaol @maiqueb an alternative to this is to move the multus configuration to Underlay, since we already have a node selector to underlay we can do this there, but it will mean we annotate the damonsetPods after openperouter deploy and that we need to use the multus dynamic controller so interfaces are dynamically attached, what I like about this is that it kind of make sense, since we are already moving the nic, infact I think we can go and totally remove the moving nic feature from openperouter and use the multus CNI that do exactly that using the Underlay possible multus feature.

PR #218, opened 2026-02-04
enhancement: internal iBGP route reflector on control plane nodes

qinqon:
/kind design

Adds the enhancement document for an internal iBGP Route Reflector.

In environments where the ToR cannot distribute EVPN routes, east/west either requires full-mesh iBGP (does not scale) or an internal RR. This enhancement adds the latter so VXLAN data plane goes directly node-to-node.

NONE

**Release note**:

```release-note
NONE
```

### DIFF
--- a/enhancements/route-reflector.md
+++ b/enhancements/route-reflector.md
@@ -1,0 +1,572 @@
+# Enhancement: Internal iBGP Route Reflector
+
+## Summary
+
+In environments where the external network does not support distributing routes between router pods via the fabric or can't be changed to do so, the only option is a full-mesh iBGP topology, which does not scale.
+
+This enhancement adds an internal iBGP Route Reflector for route distribution between router pods:
+
+- **East/West**: iBGP via an internal RR. Data plane goes directly between nodes.
+- **North/South**: eBGP with the ToR for IPv4/IPv6 (unchanged).
+
+No dedicated FRR pods are added. The hostcontroller natively configures the local FRR process as a route reflector when its Underlay CR contains a `routeReflector` section. Users add the RR node IPs as iBGP neighbors in client Underlay CRs to enable east/west — no new CRDs.
+
+## Motivation
+
+- **Cloud environments**: the ToR cannot be used for east/west traffic
+- **Hybrid deployments**: depending on the on-prem ToR for route reflection adds latency and cost to every East/West flow
+- **Full-mesh iBGP**: N*(N-1)/2 sessions, does not scale
+
+### Goals
+
+- Enable East/West without relying on the ToR
+- Keep East/West data plane traffic within the cluster (no hairpin through the ToR)
+- HA via multiple RR nodes (typically control plane nodes, 3 in production)
+- No new CRDs — opt-in by adding a `routeReflector` section to the RR nodes' Underlay CR and adding RR node IPs as iBGP neighbors in client Underlay CRs
+
+### Non-Goals
+
+- Exposing RR peering to external routers
+- Configure no-rib if nodes are pure router reflectors so they are not going to forward traffic, 
+  an [issue](https://github.com/openperouter/openperouter/issues/482) has being open to do follow up if needed 
+
+## Design
+
+The primary use case for the internal RR is environments where the ToR cannot be used for east/west traffic. In cloud deployments, control plane and worker nodes are commonly placed on different subnets. The examples throughout this document use this multi-subnet topology as it is the most complex case and the one the RR is originally designed to address.
+
+### How It Works
+
+When the Underlay CR contains a `routeReflector` section, the hostcontroller generates the RR-specific FRR configuration (`bgp listen range`, `route-reflector-client` peer-group, `bgp cluster-id`) through the existing template and conversion pipeline. All RR nodes sharing the same Underlay CR get the same `clusterID` value (default `10.0.1.1`), which is required so that inter-RR CLUSTER_LIST loop detection works correctly and clients do not receive duplicate routes (RFC 4456 §7).
+
+**Deployment**: Any node with a router pod whose Underlay CR contains a `routeReflector` section becomes an RR. The user is responsible for ensuring the router DaemonSet is scheduled on those nodes (e.g., via tolerations).
+
+RR nodes get their own Underlay CR with a `routeReflector` section.
+
+**Client configuration**: Users configure the openperouter instances to connect to the instances configured as route reflectors. The FRR templates already handle iBGP correctly.
+
+**Inter-RR peering**: Standard iBGP — users list all RR node IPs as `type: internal` neighbors with `address` in the RR Underlay CR. FRR explicit `neighbor` takes precedence over a matching `bgp listen range`, so inter-RR sessions are not treated as route-reflector clients.
+
+**`bgp listen range`**: Users configure iBGP neighbors with `listenRange` (instead of `address`) and `routeReflectorClient` — set either at the session level (`properties`), which applies it to every activated address family, or scoped to a specific family (`addressFamilies[].properties`, e.g. EVPN) — in the RR Underlay CR. The hostcontroller generates a `bgp listen range` stanza for each `listenRange` neighbor, and marks the corresponding peer-group as `route-reflector-client` in the corresponding address families. All ranges are explicit — no auto-derivation — which keeps the hostcontroller simple and supports multi-NIC / multi-subnet environments.
+
+
+
+### API
+
+Two changes to the existing API:
+
+1. A new `routeReflector` field on `UnderlaySpec` holding the shared `clusterID`.
+2. Three new fields on `Neighbor`: `listenRange` for dynamic neighbor acceptance (`bgp listen range`), `properties` for session-level features, and `addressFamilies` for per-address-family activation and features ([PR #316](https://github.com/openperouter/openperouter/pull/316), [PR #341](https://github.com/openperouter/openperouter/pull/341)).
+
+Properties use a discriminated-struct list (`[]NeighborProperty`, `+listType=map +listMapKey=type`) at two tiers: `Neighbor.properties` for session-level settings (e.g. `ebgpMultiHop`) and `NeighborAddressFamily.properties` for per-AF settings (e.g. `routeReflectorClient`). A struct list is required because Kubernetes structural schemas forbid mixed enum-or-struct items, and valued properties like `ebgpMultiHop` need typed sub-fields. The shared `NeighborPropertyType` enum with CEL placement guards keeps one type definition while enforcing valid scope. This enhancement ships `routeReflectorClient` (session or per-AF) and folds `ebgpMultiHop` (session) from [PR #341](https://github.com/openperouter/openperouter/pull/341); the same pattern extends to future properties without breaking changes. `AddressFamilyType` and the `addressFamilies` list structure are owned by [PR #316](https://github.com/openperouter/openperouter/pull/316).
+
+```go
+// UnderlaySpec defines the desired state of Underlay.
+type UnderlaySpec struct {
+	// ...existing fields...
+
+	// routeReflector configures the local FRR process as a BGP route reflector.
+	// When set, the hostcontroller generates bgp cluster-id from clusterID
+	// and derives bgp listen range and route-reflector-client stanzas from
+	// neighbors with listenRange and the routeReflectorClient property.
+	// Omit to run as a standard router without route reflection.
+	// +optional
+	RouteReflector *RouteReflectorConfig `json:"routeReflector,omitempty"`
+}
+
+// RouteReflectorConfig holds BGP Route Reflector parameters (RFC 4456).
+// Its presence on the Underlay enables route reflection on matching nodes.
+type RouteReflectorConfig struct {
+	// clusterID is the BGP cluster-id shared by all RR nodes in the same
+	// cluster (RFC 4456 §7). All RRs serving the same set of clients must
+	// use the same value so that CLUSTER_LIST loop detection prevents
+	// duplicate route reflection. The default (10.0.1.1) is outside the
+	// default routeridcidr range (10.0.0.0/24) to avoid collisions.
+	// +default="10.0.1.1"
+	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="must be a valid IPv4 address"
+	// +optional
+	ClusterID *string `json:"clusterID,omitempty"`
+}
+
+// NeighborPropertyType defines an optional feature on a Neighbor.
+// Properties are set at the session level (Neighbor.Properties) or per address
+// family (NeighborAddressFamily.Properties). CEL rules enforce valid placement.
+// +kubebuilder:validation:Enum=ebgpMultiHop;routeReflectorClient
+type NeighborPropertyType string
+
+// EBGPMultiHopProperties holds parameters for the ebgpMultiHop property.
+type EBGPMultiHopProperties struct {
+	// ttl is the maximum number of hops for the eBGP multihop session.
+	// When omitted, FRR defaults to 255.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=255
+	// +optional
+	TTL *int32 `json:"ttl,omitempty"`
+}
+
+// NeighborProperty is an optional feature applied to a neighbor session or
+// address family. The type field selects the property; typed sub-fields hold
+// parameters for properties that require them.
+// +kubebuilder:validation:XValidation:rule="!has(self.ebgpMultiHop) || self.type == 'ebgpMultiHop'",message="ebgpMultiHop parameters require type ebgpMultiHop"
+type NeighborProperty struct {
+	// type selects the property.
+	// +required
+	Type NeighborPropertyType `json:"type"`
+
+	// ebgpMultiHop holds parameters for the ebgpMultiHop property.
+	// May only be set when type is ebgpMultiHop.
+	// +optional
+	EBGPMultiHop *EBGPMultiHopProperties `json:"ebgpMultiHop,omitempty"`
+}
+
+// AddressFamilyType identifies a BGP address family.
+// Owned by PR #316; listed here for context.
+// +kubebuilder:validation:Enum=ipv4unicast;ipv6unicast;evpn;ipv4vpn;ipv6vpn
+type AddressFamilyType string
+
+// NeighborAddressFamily activates an address family on a neighbor with
+// optional per-AF features.
+// +kubebuilder:validation:XValidation:rule="!has(self.properties) || !self.properties.exists(o, o.type == 'ebgpMultiHop')",message="ebgpMultiHop is session-level only, not per address family"
+type NeighborAddressFamily struct {
+	// type selects the address family.
+	// +required
+	Type AddressFamilyType `json:"type"`
+
+	// properties is the set of optional features for this address family.
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	Properties []NeighborProperty `json:"properties,omitempty"`
+}
+
+// Neighbor represents a BGP Neighbor we want FRR to connect to.
+// +kubebuilder:validation:XValidation:rule="!has(self.listenRange) || !has(self.address)",message="listenRange and address are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!has(self.listenRange) || (has(self.type) && self.type == 'internal')",message="listenRange requires type internal"
+// +kubebuilder:validation:XValidation:rule="!(has(self.properties) && self.properties.exists(o, o.type == 'routeReflectorClient')) && !(has(self.addressFamilies) && self.addressFamilies.exists(af, has(af.properties) && af.properties.exists(o, o.type == 'routeReflectorClient'))) || (has(self.type) && self.type == 'internal')",message="routeReflectorClient requires type internal"
+type Neighbor struct {
+	// ...existing fields (type, address, asn, etc.)...
+
+	// listenRange is a CIDR range for dynamic neighbor acceptance via FRR
+	// bgp listen range. When set, the hostcontroller generates a
+	// "bgp listen range <listenRange> peer-group <name>" stanza instead of
+	// an explicit neighbor statement. Mutually exclusive with address.
+	// +kubebuilder:validation:XValidation:rule="isCIDR(self)",message="must be a valid CIDR"
+	// +optional
+	ListenRange *string `json:"listenRange,omitempty"`
+
+	// properties is the set of optional session-level features for this
+	// neighbor. ebgpMultiHop applies to the BGP session. routeReflectorClient
+	// set here is applied to every address family activated in
+	// addressFamilies; to enable it for specific families only, set it under
+	// addressFamilies[].properties instead.
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	Properties []NeighborProperty `json:"properties,omitempty"`
+
+	// addressFamilies is the list of address families to activate on
+	// this neighbor, each with optional per-AF features.
+	// Owned by PR #316; extended here with per-AF properties.
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	AddressFamilies []NeighborAddressFamily `json:"addressFamilies,omitempty"`
+}
+```
+
+The two-tier properties model in YAML:
+
+```yaml
+neighbors:
+  - type: internal
+    address: 10.0.0.5
+    properties:                          # session-level (applies to the whole BGP session)
+      - type: ebgpMultiHop
+        ebgpMultiHop:
+          ttl: 3
+    addressFamilies:
+      - type: evpn
+        properties:                      # per-AF (applies only to this address family)
+          - type: routeReflectorClient
+      - type: ipv4unicast               # no per-AF properties on this family
+```
+
+### Session Topology
+
+```
+            ToR (eBGP)
+               |
+  +------------+------------+
+  |            |            |
+cp-1 (RR) <--iBGP--> cp-2 (RR)       (explicit full mesh between RRs)
+  ^  bgp listen range       ^
+  |  route-reflector-client  |
+  +----------+--------------+
+             |
+   worker-1 (client)    worker-2 (client)    (connect to all RRs)
+```
+
+- **Clients -> RRs**: clients configure RR IPs as explicit iBGP neighbors via the Underlay CR. RRs accept passively via `bgp listen range` covering all underlay subnets.
+- **RR <-> RR**: standard iBGP neighbors from the RR Underlay CR. Explicit `neighbor` takes precedence over `bgp listen range`.
+- **All nodes -> ToR**: eBGP unchanged.
+- **Path selection**: iBGP (local-preference 100) beats the longer eBGP AS path, so VXLAN goes directly between nodes rather than hairpinning through the ToR.
+- **RR-only nodes**: when RR nodes do not participate in the data plane, they do NOT have a ToR eBGP session — they only do route reflection.
+
+### Example: 4-Node Dual-Stack Cluster
+
+Setup: ASN 64514, ToR at 192.168.11.2 / fd00:11::2 (ASN 64512). Control plane nodes are on a separate subnet (192.168.10.0/24, fd00:10::/64) from workers (192.168.11.0/24, fd00:11::/64), as is typical in cloud environments.
+
+| Node | Role | Underlay NIC IPs | VTEP IP |
+|------|------|-----------------|---------|
+| cp-1 | RR (control plane) | 192.168.10.3/24, fd00:10::3/64 | 100.65.0.0 |
+| cp-2 | RR (control plane) | 192.168.10.4/24, fd00:10::4/64 | 100.65.0.1 |
+| worker-1 | client | 192.168.11.5/24, fd00:11::5/64 | 100.65.0.2 |
+| worker-2 | client | 192.168.11.6/24, fd00:11::6/64 | 100.65.0.3 |
+
+#### Underlay CRs — CP nodes with full routing
+
+When control plane nodes participate in the data plane, they get a full Underlay with ToR. Workers include the control plane RR IPs as iBGP neighbors. The control plane Underlay uses `listenRange` neighbors with `routeReflectorClient` in their EVPN address family for dynamic client acceptance, and `address` neighbors for inter-RR peering — each hostcontroller filters out its own IP:
+
+```yaml
+# Worker Underlay
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-workers
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchExpressions:
+      - key: node-role.kubernetes.io/control-plane
+        operator: DoesNotExist
+  asn: 64514
+  tunnelEndpoint:
+    cidrs:
+      - 100.65.0.0/24
+  nics:
+    - toswitch
+  neighbors:
+    - asn: 64512                       # ToR (eBGP, IPv4)
+      address: 192.168.11.2
+    - asn: 64512                       # ToR (eBGP, IPv6)
+      address: fd00:11::2
+    - type: internal                   # cp-1 RR (iBGP, IPv4)
+      address: 192.168.10.3
+    - type: internal                   # cp-1 RR (iBGP, IPv6)
+      address: fd00:10::3
+    - type: internal                   # cp-2 RR (iBGP, IPv4)
+      address: 192.168.10.4
+    - type: internal                   # cp-2 RR (iBGP, IPv6)
+      address: fd00:10::4
+---
+# Control plane Underlay (full routing + RR)
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-controlplane
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchLabels:
+      node-role.kubernetes.io/control-plane: ""
+  asn: 64514
+  tunnelEndpoint:
+    cidrs:
+      - 100.65.0.0/24
+  nics:
+    - toswitch
+  routeReflector:
+    clusterID: 10.0.1.1                # optional, shown for clarity
+  neighbors:
+    - asn: 64512                       # ToR (eBGP, IPv4)
+      address: 192.168.10.2
+    - asn: 64512                       # ToR (eBGP, IPv6)
+      address: fd00:10::2
+    - type: internal                   # dynamic RR clients (CP subnet, IPv4)
+      listenRange: 192.168.10.0/24
+      addressFamilies:
+        - type: evpn
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # dynamic RR clients (worker subnet, IPv4)
+      listenRange: 192.168.11.0/24
+      addressFamilies:
+        - type: evpn
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # dynamic RR clients (CP subnet, IPv6)
+      listenRange: fd00:10::/64
+      addressFamilies:
+        - type: evpn
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # dynamic RR clients (worker subnet, IPv6)
+      listenRange: fd00:11::/64
+      addressFamilies:
+        - type: evpn
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # cp-1 (inter-RR, filtered on cp-1)
+      address: 192.168.10.3
+    - type: internal                   # cp-1 (inter-RR, IPv6, filtered on cp-1)
+      address: fd00:10::3
+    - type: internal                   # cp-2 (inter-RR, filtered on cp-2)
+      address: 192.168.10.4
+    - type: internal                   # cp-2 (inter-RR, IPv6, filtered on cp-2)
+      address: fd00:10::4
+```
+
+#### Underlay CRs — CP nodes as RR only
+
+When control plane nodes only do route reflection (no data plane), workers get the ToR + RR neighbors. Control plane nodes get an Underlay with NIC, ASN, `listenRange` neighbors with `routeReflectorClient` in their EVPN address family, and inter-RR `address` neighbors (no ToR, no `tunnelEndpoint`):
+
+```yaml
+# Worker Underlay
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-workers
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchExpressions:
+      - key: node-role.kubernetes.io/control-plane
+        operator: DoesNotExist
+  asn: 64514
+  tunnelEndpoint:
+    cidrs:
+      - 100.65.0.0/24
+  nics:
+    - toswitch
+  neighbors:
+    - asn: 64512                       # ToR (eBGP, IPv4)
+      address: 192.168.11.2
+    - asn: 64512                       # ToR (eBGP, IPv6)
+      address: fd00:11::2
+    - type: internal                   # cp-1 RR (iBGP, IPv4)
+      address: 192.168.10.3
+    - type: internal                   # cp-1 RR (iBGP, IPv6)
+      address: fd00:10::3
+    - type: internal                   # cp-2 RR (iBGP, IPv4)
+      address: 192.168.10.4
+    - type: internal                   # cp-2 RR (iBGP, IPv6)
+      address: fd00:10::4
+---
+# Control plane Underlay (RR only, no data plane)
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-controlplane
+  namespace: openperouter-system
+spec:
+  nodeSelector:
+    matchLabels:
+      node-role.kubernetes.io/control-plane: ""
+  asn: 64514
+  nics:
+    - toswitch
+  routeReflector:
+    clusterID: 10.0.1.1
+  neighbors:
+    - type: internal                   # dynamic RR clients (CP subnet, IPv4)
+      listenRange: 192.168.10.0/24
+      addressFamilies:
+        - type: evpn
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # dynamic RR clients (worker subnet, IPv4)
+      listenRange: 192.168.11.0/24
+      addressFamilies:
+        - type: evpn
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # dynamic RR clients (CP subnet, IPv6)
+      listenRange: fd00:10::/64
+      addressFamilies:
+        - type: evpn
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # dynamic RR clients (worker subnet, IPv6)
+      listenRange: fd00:11::/64
+      addressFamilies:
+        - type: evpn
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # cp-1 (inter-RR, filtered on cp-1)
+      address: 192.168.10.3
+    - type: internal                   # cp-1 (inter-RR, IPv6, filtered on cp-1)
+      address: fd00:10::3
+    - type: internal                   # cp-2 (inter-RR, filtered on cp-2)
+      address: 192.168.10.4
+    - type: internal                   # cp-2 (inter-RR, IPv6, filtered on cp-2)
+      address: fd00:10::4
+```
+
+#### Key FRR Additions on RR Nodes
+
+The hostcontroller generates the following RR-specific FRR stanzas through the template and conversion pipeline. These are added on top of the normal router configuration (or standalone for RR-only nodes):
+
+**Cluster ID** — all RR nodes in the same cluster must share the same cluster-id so that CLUSTER_LIST loop detection works and clients do not receive duplicate reflected routes (RFC 4456 §7). The value comes from `routeReflector.clusterID` (default `10.0.1.1`):
+
+```
+bgp cluster-id 10.0.1.1
+```
+
+**Peer-group name generation** — FRR requires a peer-group for `bgp listen range`. The hostcontroller generates one peer-group per `listenRange` neighbor, deriving the name by sanitizing the CIDR: replace every `.`, `:`, and `/` with `-`, then prefix with `PG-`. The substitution is per character, so an IPv6 `::` becomes `--` and the trailing `/` adds one more `-` (e.g. `fd00:10::/64` → `PG-fd00-10---64`); because each separator maps to its own `-`, distinct CIDRs always produce distinct peer-group names. Examples:
+
+| CIDR | Peer-group name |
+|------|----------------|
+| `192.168.10.0/24` | `PG-192-168-10-0-24` |
+| `192.168.11.0/24` | `PG-192-168-11-0-24` |
+| `fd00:10::/64` | `PG-fd00-10---64` |
+| `fd00:11::/64` | `PG-fd00-11---64` |
+
+**Dynamic client acceptance** — for each `listenRange` neighbor, the hostcontroller generates a peer-group with a `bgp listen range` stanza. Each CIDR gets its own peer-group so that per-neighbor settings (options, timers, etc.) remain independent:
+
+```
+neighbor PG-192-168-10-0-24 peer-group
+neighbor PG-192-168-10-0-24 remote-as internal
+bgp listen range 192.168.10.0/24 peer-group PG-192-168-10-0-24
+
+neighbor PG-192-168-11-0-24 peer-group
+neighbor PG-192-168-11-0-24 remote-as internal
+bgp listen range 192.168.11.0/24 peer-group PG-192-168-11-0-24
+
+neighbor PG-fd00-10---64 peer-group
+neighbor PG-fd00-10---64 remote-as internal
+bgp listen range fd00:10::/64 peer-group PG-fd00-10---64
+
+neighbor PG-fd00-11---64 peer-group
+neighbor PG-fd00-11---64 remote-as internal
+bgp listen range fd00:11::/64 peer-group PG-fd00-11---64
+```
+
+**Route reflection** — because the `listenRange` neighbors have `routeReflectorClient` in their EVPN address family properties, the hostcontroller marks each peer-group as `route-reflector-client` in the corresponding address families (setting `routeReflectorClient` at the session level instead would fan it out to every activated address family):
+
+```
+address-family l2vpn evpn
+  neighbor PG-192-168-10-0-24 activate
+  neighbor PG-192-168-10-0-24 route-reflector-client
+  neighbor PG-192-168-11-0-24 activate
+  neighbor PG-192-168-11-0-24 route-reflector-client
+  neighbor PG-fd00-10---64 activate
+  neighbor PG-fd00-10---64 route-reflector-client
+  neighbor PG-fd00-11---64 activate
+  neighbor PG-fd00-11---64 route-reflector-client
+exit-address-family
+```
+
+#### Client Nodes
+
+Client nodes use standard iBGP neighbors pointing at the RR IPs (configured via the Underlay CR). No RR-specific FRR stanzas are needed on clients.
+
+#### BGP sessions at runtime — CP node as RR + normal router (cp-1)
+
+```
+L2VPN EVPN Summary (BGP router identifier 10.0.0.1, local AS 64514):
+
+Neighbor        V    AS   Up/Down  State/PfxRcd  Desc
+*192.168.11.5   4  64514  00:25:56           5   <- dynamic client (worker-1, IPv4)
+*192.168.11.6   4  64514  00:25:56           2   <- dynamic client (worker-2, IPv4)
+*fd00:11::5     4  64514  00:25:56           5   <- dynamic client (worker-1, IPv6)
+*fd00:11::6     4  64514  00:25:56           2   <- dynamic client (worker-2, IPv6)
+192.168.10.2    4  64512  00:25:56           9   <- ToR (eBGP, IPv4)
+fd00:10::2      4  64512  00:25:56           9   <- ToR (eBGP, IPv6)
+192.168.10.4    4  64514  00:25:56           9   <- other RR cp-2 (explicit, IPv4)
+fd00:10::4      4  64514  00:25:56           9   <- other RR cp-2 (explicit, IPv6)
+
+* = dynamic neighbor accepted via bgp listen range
+```
+
+#### BGP sessions at runtime — CP node as RR only (cp-1)
+
+```
+L2VPN EVPN Summary (BGP router identifier 10.0.0.1, local AS 64514):
+
+Neighbor        V    AS   Up/Down  State/PfxRcd  Desc
+*192.168.11.5   4  64514  00:25:56           5   <- dynamic client (worker-1, IPv4)
+*192.168.11.6   4  64514  00:25:56           2   <- dynamic client (worker-2, IPv4)
+*fd00:11::5     4  64514  00:25:56           5   <- dynamic client (worker-1, IPv6)
+*fd00:11::6     4  64514  00:25:56           2   <- dynamic client (worker-2, IPv6)
+192.168.10.4    4  64514  00:25:56           9   <- other RR cp-2 (explicit, IPv4)
+fd00:10::4      4  64514  00:25:56           9   <- other RR cp-2 (explicit, IPv6)
+
+* = dynamic neighbor accepted via bgp listen range
+```
+
+#### BGP sessions at runtime — client node (worker-1)
+
+```
+L2VPN EVPN Summary (BGP router identifier 10.0.0.2, local AS 64514):
+
+Neighbor        V    AS   Up/Down  State/PfxRcd  Desc
+192.168.11.2    4  64512  00:25:56           9   <- ToR (eBGP, IPv4)
+fd00:11::2      4  64512  00:25:56           9   <- ToR (eBGP, IPv6)
+192.168.10.3    4  64514  00:25:56           9   <- RR cp-1 (IPv4)
+fd00:10::3      4  64514  00:25:56           9   <- RR cp-1 (IPv6)
+192.168.10.4    4  64514  00:25:56           9   <- RR cp-2 (IPv4)
+fd00:10::4      4  64514  00:25:56           9   <- RR cp-2 (IPv6)
+```
+
+### Controller Reconcile Logic
+
+**Hostcontroller** (existing, extended with RR logic):
+
+- Watches: Nodes, Underlays, L2VNIs, L3VNIs, Pods (router)
+- When the Underlay CR contains a `routeReflector` section, the hostcontroller additionally:
+  - Filters out `address` neighbors matching its own IP (self-filtering)
+  - Generates `bgp cluster-id` from `routeReflector.clusterID` (default `10.0.1.1`)
+  - Collects neighbors with `listenRange` + `routeReflectorClient` in their address family properties → generates one peer-group with a `bgp listen range` per CIDR and `route-reflector-client` in the corresponding AFs
+  - All through the existing FRR template and conversion pipeline
+- Processes `address` neighbors as before (including inter-RR iBGP neighbors and client iBGP neighbors from client Underlay CRs)
+
+### Failure and Recovery
+
+RR nodes are typically stable (e.g., control plane nodes) and do not get rescheduled. DaemonSets restart pods on the same node.
+
+**RR node goes down** (e.g., cp-1 fails):
+
+1. **Client nodes**: iBGP sessions to cp-1 time out (based on hold-time, or faster with BFD). Sessions to cp-2 (surviving RR) stay up — **no traffic disruption**.
+2. **cp-2 RR**: iBGP session to cp-1 drops. cp-2 continues reflecting routes from all clients.
+3. **When cp-1 recovers**: DaemonSet recreates the router pod, FRR restarts, sessions re-establish.
+
+**Router pod crash on RR node** (node is fine):
+
+1. DaemonSet restarts the pod. The hostcontroller regenerates the RR configuration on startup.
+2. Brief BGP session flap during restart.
+
+Convergence is bounded by BGP hold-time (default 180s, or less with BFD) for session failover, and pod restart time for recovery.
+
+### Systemd Mode
+
+No special handling is needed. All RR logic lives in the Underlay spec, and the static configuration files use the same `spec` schema as the Kubernetes CRs. The hostcontroller processes the `routeReflector` section through the same conversion and template pipeline regardless of whether the Underlay comes from a static file or the API server.
+
+### Testing
+
+**Unit tests** (`internal/frr/`, `internal/conversion/`):
+
+- FRR template rendering: RR stanzas (`bgp cluster-id`, `bgp listen range`, `route-reflector-client`) rendered correctly from `listenRange` + `routeReflectorClient` neighbors (golden file tests)
+- Conversion: an Underlay with a `routeReflector` section generates the `bgp cluster-id` and RR peer-group stanzas; an Underlay without it generates none
+- Neighbors with `listenRange` produce `bgp listen range` stanzas, neighbors with `address` produce explicit `neighbor` statements
+
+**E2e tests** (`e2etests/tests/`, Ginkgo suite against a 4-node clab cluster):
+
+- **BGP sessions**: all iBGP sessions to RR nodes established, dynamic clients accepted via `bgp listen range` (`*` prefix in vtysh output)
+- **Dual stack listen range**: verify that `bgp listen range` covers all `listenRange` neighbors with `routeReflectorClient` in their EVPN address family to support dual stack scenarios
+- **Route reflection**: routes on client nodes show `*>i` (iBGP best path from RR), not the eBGP path from the ToR
+- **East/West data plane**: ping between workloads on different client nodes, VXLAN packets go directly node-to-node (zero packets via ToR IP)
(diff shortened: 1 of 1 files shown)


enhancements/route-reflector.md
```diff
@@ -0,0 +1,673 @@
+# Enhancement: BGP Route Reflector
+
+## Summary
+
+Add an optional internal BGP Route Reflector (RR) component to OpenPERouter to enable EVPN route distribution between router pods. The RR is deployed as a Deployment using the cluster's default network (CNI), configured via Helm values. Router pods automatically discover and peer with RR pods when the Underlay is configured with `routeReflector.type: Internal`.
+
```
maiqueb: I think you're mixing the what with the how and where.

The enhancement is saying in the summary how the solution will be implemented, which is, IMHO, wrong.

It should first highlight the problem it is solving, then present alternatives, and argue for one of them.

maiqueb (changes requested):

maiqueb (changes requested):
Thanks for this proposal. 

Some questions, thoughts, and suggestions.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,673 @@
...
+            - containerPort: 179
+              name: bgp
+              protocol: TCP
+          securityContext:
+            capabilities:
+              add:
+                - NET_ADMIN
+                - NET_RAW
+                - NET_BIND_SERVICE
+                - SYS_ADMIN
```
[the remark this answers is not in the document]
maiqueb: Good catch bot. 
This was done; resolving.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,652 @@
...
+```
+
+**The Route Reflector solves this by:**
+1. Handling EVPN route distribution between router pods within the cluster
+2. Router pods peer with cloud router for external IPv4/IPv6 routes (eBGP)
+3. Keeping EVPN control plane internal while maintaining external connectivity
+
+### Use Case: Hybrid Cloud Cost and Efficiency
+
+In hybrid deployments connecting on-premises infrastructure with cloud, depending on the on-prem ToR for route reflection is costly and inefficient:
```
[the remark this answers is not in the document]
maiqueb: Thanks for educating me. I now understand the use case better.

maiqueb (commented):
Some more comments.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Design
+
+### How It Works
+
+Each hostcontroller reads the underlay NIC IPs (IPv4 and/or IPv6, with mask) from inside the router pod network namespace and writes them as a JSON array to the `openperouter.io/underlay-ips` annotation on the router pod (e.g. `["192.168.11.3/24","fd00::3/64"]`).
+
+> **Note:** This annotation is a temporary discovery mechanism. It is expected to be replaced by a per-node Underlay status subresource.
+
+The RR controller is a 2-replica Deployment. Each replica lands on a different node (anti-affinity) that runs a router pod (affinity). It reads the local router pod's `underlay-ips` annotation and creates a `RawFRRConfig` with `bgp listen range` (one per address family, derived by zeroing the host bits) and `route-reflector-client`.
```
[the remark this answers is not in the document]
qinqon: > this is under the assumption that all the nodes are on the same subnet, which might not always be the case. 

In GCP for example control plane and workers has different network/subnets, but I understand that's not a problem for RR since RR is for the L2VNI and L3VNI routes right ? maybe a problem for passthrough?

> Also, there is the assumption that all the nodes have one single ip for the underlay.

qinqon:
> > > We need automation to look for the RR ip (what we want us to peer with) and on the subnet the clients are going to come from. This relies on the fact that we have one IP per node, and that all the nodes are on the same subnet.
> > > It will work on cloud providers but will conflict in real datacenters, where nodes might have multiple nics, each nic might have multiple ips, and different nodes might land on different racks.
> > > Additionally, we are going soon (#279) to allow having multiple nics for the underlay, which makes things even more complicated.
> > 
> > 
> > I don't see that as a problem, we just discover the subnets and RR IPs and compose the config, also do we really care about interfaces ?
> 
> We don't care about interfaces in general, but this means we'll have more nics (and thus, more ips and more subnets to take into account, not just one)
> 
> > > I would try to implement something simpler (even if less reliable) such as always running the route reflectors on master nodes and asking the users to configure the session as they would on an external router.
> > 
> > 
> > That was my first approach, but we have a knob at openperouter installation where we can disable running at control planes, so customers with that will break, but maybe if that knob is set we can still run FRRs just with the router reflector config at ocntrol planes ? at the end only control plane traffic if going to go over there not data plane traffic.
> 
> I like this idea, it's simpler and makes sense as in that scenario the RR is critical

@fedepaol so I am going to reword the enhancement for the following:
- Run RR on control plane nodes, so we simplify node selection and restarts and the like.
- If openperouter has router pods also running at control plane nodes we re-use the FRR there and just add some extra config 
- If openperouter has no router pods for control plane (it uses the knob to disable it at control plane) it start a new router pod but will just do the RR config.
- The user will need to manually configure underlay to point to RR control plane node IPs as neighbor to be able to do east/west, we we don't want to do east/west over the ToR (cloud envs)

fedepaol (commented):
reviewed until a certain point, will stop as the following bits make assumptions on that

enhancements/route-reflector.md
```diff
@@ -0,0 +1,416 @@
...
+
+## Design
+
+The primary use case for the internal RR is cloud environments where the ToR does not support EVPN. In cloud deployments, control plane and worker nodes are commonly placed on different subnets. The examples throughout this document use this multi-subnet topology as it is the most complex case and the one the RR is originally designed to address.
+
+### How It Works
+
+The RR controller is a DaemonSet pinned to control plane nodes (`nodeSelector: node-role.kubernetes.io/control-plane`, toleration for the `node-role.kubernetes.io/control-plane:NoSchedule` taint). It runs on every control plane node.
+
+**Two deployment modes based on `runOnMaster`**:
```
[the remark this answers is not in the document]
qinqon: > I think we should have only one deployment mode "enableRR" or something like that that assumes we are running on masters.

@fedepaol more than modes is an explanation of what the router pod do related to RR depending on runOnMaster.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,416 @@
...
+
+- **`runOnMaster=true` (default)**: Router pods already run on control plane nodes as part of the regular router DaemonSet. The RR controller DaemonSet is lightweight — it only creates a `RawFRRConfig` for its local node, adding `bgp listen range`, `route-reflector-client` peer-group, and `bgp cluster-id` to the existing FRR process. No additional router pods are needed.
```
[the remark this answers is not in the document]
qinqon: > we shouldn't use rawconfig at all. The controller should know if it needs to act as RR or not.

Ok, I will make hostcontroller discover if it's RR or not and add the needed bits to the FRR config.

> Also, this shouldn't prevent interacting with the TOR either.

What do you mean ? at hybrid cloud for east/west we should not interact with the with the on prem TOR at all since it may be at the other side of the world.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,416 @@
...
+
+- **`runOnMaster=false`**: No regular router pods run on control plane nodes. An additional "RR router" DaemonSet is deployed on control plane nodes, using the same pod spec as the regular router DaemonSet (FRR container + hostcontroller/reloader). The RR controller creates `RawFRRConfig` as in the `runOnMaster=true` case. The control plane nodes get a separate Underlay CR with `nodeSelector` targeting control plane nodes, containing only the NIC and ASN (no ToR eBGP neighbor needed, since these nodes only do route reflection — no VXLAN data plane participation).
+
+**Client configuration**: Users manually add the control plane node underlay IPs as neighbors in their Underlay CR with the same ASN (iBGP). The existing FRR template renders these as iBGP neighbors — `neighbor X.X.X.X remote-as <same-ASN>`. The `allowas-in` directive is emitted only for eBGP neighbors (neighbor ASN != local ASN), so iBGP neighbors are handled correctly.
```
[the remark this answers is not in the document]
qinqon: Ahh so this means we will not need to implement the iBGP logic at this and it will be already in place right ? 

I will make this enhancement depend on that:
- https://github.com/openperouter/openperouter/pull/260

enhancements/route-reflector.md
```diff
@@ -0,0 +1,416 @@
...
+
+> **Note:** The FRR templates (`underlay_evpn.tmpl`, `neighboripfamily.tmpl`) must be updated to conditionally emit `allowas-in` only when the neighbor ASN differs from the local ASN. Currently they emit it unconditionally.
+
+**Inter-RR peering**: Each RR controller discovers other RR controller pods on control plane nodes and adds explicit iBGP neighbors for each peer RR in the `RawFRRConfig`. This ensures a full mesh between all RRs. FRR explicit `neighbor` takes precedence over a matching `bgp listen range`, so inter-RR sessions use explicit peering.
```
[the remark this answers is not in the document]
qinqon: > I'd rather configure the same underlay for them too (and maybe skip if I am peering with myself). Would that make sense?

ahh so you mean that if we have three control plane nodes we put at the control plane underlay we puth the three of them, then hostcontroller ignore the neighbor that is itself, so users just need one underlay and not three ? that's it right ? ok, let's do that explicit.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+
+- **[PR #260](https://github.com/openperouter/openperouter/pull/260) — iBGP support**: Adds the `PeerASN` type with `"internal"`/`"external"` modes, conditional `allowas-in` (eBGP only) and `next-hop-self force` (iBGP only) in FRR templates, and removes validation that blocked same-ASN neighbors. This enhancement builds on top of PR #260's iBGP support for client-to-RR and inter-RR peering.
+
+## Design
+
+The primary use case for the internal RR is environments where the ToR cannot be used for east/west traffic. In cloud deployments, control plane and worker nodes are commonly placed on different subnets. The examples throughout this document use this multi-subnet topology as it is the most complex case and the one the RR is originally designed to address.
+
+### How It Works
+
+When the Underlay CR contains a `routeReflector` section, the hostcontroller generates the RR-specific FRR configuration (`bgp listen range`, `route-reflector-client` peer-group, `bgp cluster-id`, inter-RR peers) through the existing template and conversion pipeline.
```
fedepaol, line 41: would it make sense to have it in a new CR instead of a section of underlay? I am asking because with the introduction of srv6, I had the feeling that we (I) put too many things in the underlay already _just because it's our "singleton"_

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+
+**Deployment**: Any node with a router pod whose Underlay CR contains a `routeReflector` section becomes an RR. The user is responsible for ensuring the router DaemonSet is scheduled on those nodes (e.g., via tolerations). No separate DaemonSet is ever created.
+
+RR nodes get their own Underlay CR with a `routeReflector` section. Its contents depend on whether the RR nodes also participate in the data plane (ToR + full routing) or only do route reflection (NIC + ASN only, no ToR).
```
fedepaol, line 45: I wouldn't expect to have different contents, but I'd expect a different behavior behind the scenes.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+
+**Client configuration**: Users manually add the RR node underlay IPs as `type: internal` neighbors in the client Underlay CR (PR #260). The FRR templates already handle iBGP correctly — `allowas-in` is emitted only for eBGP neighbors and `next-hop-self force` is emitted for iBGP neighbors.
```
fedepaol, line 47: I'd rephrase with "users configure the openperouter instances to connect to the instances configured as route reflectors"

fedepaol: Also, I am not quite sure next hop self force is good when we act as RR

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+- **Path selection**: iBGP (local-preference 100) beats the longer eBGP AS path, so VXLAN goes directly between nodes rather than hairpinning through the ToR.
+- **RR-only nodes**: when RR nodes do not participate in the data plane, they do NOT have a ToR eBGP session — they only do route reflection.
+
+### Example: 4-Node Dual-Stack Cluster
+
+Setup: ASN 64514, ToR at 192.168.11.2 / fd00:11::2 (ASN 64512). Control plane nodes are on a separate subnet (192.168.10.0/24, fd00:10::/64) from workers (192.168.11.0/24, fd00:11::/64), as is typical in cloud environments.
+
+| Node | Role | Underlay NIC IPs | VTEP IP |
+|------|------|-----------------|---------|
+| cp-1 | RR (control plane) | 192.168.10.3/24, fd00:10::3/64 | 100.65.0.0 |
```
fedepaol, line 80: Let's focus on single stack unless we think it's super relevant. FWIW, we will allow advertising v4 routes over v6 links in https://github.com/openperouter/openperouter/pull/286

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+| cp-2 | RR (control plane) | 192.168.10.4/24, fd00:10::4/64 | 100.65.0.1 |
+| worker-1 | client | 192.168.11.5/24, fd00:11::5/64 | 100.65.0.2 |
+| worker-2 | client | 192.168.11.6/24, fd00:11::6/64 | 100.65.0.3 |
+
+#### Underlay CRs — CP nodes with full routing
+
+When control plane nodes participate in the data plane (e.g. `runOnMaster=true`), they get a full Underlay with ToR. Workers include the control plane RR IPs as iBGP neighbors. The control plane Underlay lists all CP node IPs as `type: internal` neighbors for inter-RR peering — each hostcontroller filters out its own IP:
```
fedepaol, line 87: we said it doesn't matter where RR runs, so probably the runOnMaster part is a leftover?

fedepaol: What I think it makes sense is to suggest using masters as RRs (and documenting how to do so, as you are doing here) because of the special treatment masters get

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+      address: 192.168.10.3
+    - type: internal                   # cp-1 (iBGP, IPv6, filtered on cp-1)
+      address: fd00:10::3
+    - type: internal                   # cp-2 (iBGP, IPv4, filtered on cp-2)
+      address: 192.168.10.4
+    - type: internal                   # cp-2 (iBGP, IPv6, filtered on cp-2)
+      address: fd00:10::4
+```
+
+#### Resulting frr.conf — CP node as RR + normal router (cp-1)
```
fedepaol, line 221: not sure I would discuss the resulting frr here. Maybe we can add a few snippets on what we are going to add and why, but the whole config is a bit too much imo and we'll certainly drift when doing the real implementation

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+1. **Client nodes**: iBGP sessions to cp-1 time out (based on hold-time, or faster with BFD). Sessions to cp-2 (surviving RR) stay up — **no traffic disruption**.
+2. **cp-2 RR**: inter-RR session to cp-1 drops. cp-2 continues reflecting routes from all clients.
+3. **When cp-1 recovers**: DaemonSet recreates the router pod, FRR restarts, sessions re-establish.
+
+**Router pod crash on RR node** (node is fine):
+
+1. DaemonSet restarts the pod. The hostcontroller regenerates the RR configuration on startup.
+2. Brief BGP session flap during restart.
+
+Convergence is bounded by BGP hold-time (default 180s, or less with BFD) for session failover, and pod restart time for recovery.
```
fedepaol, line 402: Let's add a sectoin about how this is goign to work on systemdmode (fine, I guess as all the logic is in the underlay or whatever CRD we'll add)

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+- **BGP sessions**: all iBGP sessions to RR nodes established, dynamic clients accepted via `bgp listen range` (`*` prefix in vtysh output)
+- **Multi-subnet listen range**: verify that `bgp listen range` covers all configured `listenRanges`
+- **Route reflection**: routes on client nodes show `*>i` (iBGP best path from RR), not the eBGP path from the ToR
+- **East/West data plane**: ping between workloads on different client nodes, VXLAN packets go directly node-to-node (zero packets via ToR IP)
+- **Failover**: delete one router pod on an RR node -> verify DaemonSet recreates it, surviving RR keeps sessions up, clients reconverge
+- **RR-only nodes**: verify Underlay with `routeReflector` and no ToR produces RR-only FRR config, route reflection works
+- **RR disabled**: without `routeReflector` section in Underlay, system behaves identically to baseline (no RR stanzas in FRR config)
+- **ToR VXLAN block**: drop VXLAN (UDP 4789) forwarding on the ToR (`iptables -I FORWARD -i toswitch -p udp --dport 4789 -j DROP`) to prove East/West traffic never hairpins through the ToR — pings between workloads must still succeed via direct node-to-node VXLAN
```
fedepaol, line 422: do we really want to test this?

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+
+**CI lane**: run the full existing e2e suite with route reflection configured on a 4-node cluster. The RR should be transparent — all existing tests must pass without modification.
```
fedepaol, line 424: this means we'll have to change the clab env right?

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+
+- FRR template rendering: RR stanzas (`bgp listen range`, `route-reflector-client`, `bgp cluster-id`, inter-RR peers) rendered correctly (golden file tests)
+- iBGP `allowas-in` / `next-hop-self force` handling already covered by PR #260 tests
+
+**E2e tests** (`e2etests/tests/`, Ginkgo suite against a 4-node clab cluster):
+
+- **RR configuration lifecycle**: add `routeReflector` section to RR Underlay -> verify RR stanzas present in FRR config on matching nodes
+- **Underlay iBGP neighbors**: verify that adding RR node IPs as `type: internal` neighbors in client Underlay CR produces correct FRR config
+- **BGP sessions**: all iBGP sessions to RR nodes established, dynamic clients accepted via `bgp listen range` (`*` prefix in vtysh output)
+- **Multi-subnet listen range**: verify that `bgp listen range` covers all configured `listenRanges`
```
fedepaol, line 416: is this something we want to e2e?

enhancements/route-reflector.md
```diff
@@ -0,0 +1,438 @@
...
+- **Route reflection**: routes on client nodes show `*>i` (iBGP best path from RR), not the eBGP path from the ToR
+- **East/West data plane**: ping between workloads on different client nodes, VXLAN packets go directly node-to-node (zero packets via ToR IP)
+- **Failover**: delete one router pod on an RR node -> verify DaemonSet recreates it, surviving RR keeps sessions up, clients reconverge
+- **RR-only nodes**: verify Underlay with `routeReflector` and no ToR produces RR-only FRR config, route reflection works
+- **RR disabled**: without `routeReflector` section in Underlay, system behaves identically to baseline (no RR stanzas in FRR config)
```
fedepaol, line 421: this means  running the current tests I thikn? If so, the item is not necessary

enhancements/route-reflector.md
qinqon: > would it make sense to have it in a new CR instead of a section of underlay? I am asking because with the introduction of srv6, I had the feeling that we (I) put too many things in the underlay already _just because it's our "singleton"_

I am not sure it merits a different CRD since we are just going to configure some subnets on it :-/, also most of the time RR is not going to be configured so maintaining a CRD for just 5% of cases (cloud envs) not sure is it worth it, but this is just my 2 cents, implementation wise is more or less we just another watcher to the same controller and look for everything as we do today.

enhancements/route-reflector.md
qinqon: > I wouldn't expect to have different contents, but I'd expect a different behavior behind the scenes.

What all means is that in the scenario where user don't want to run router pods on the workers but do want router reflector, since we have decide to to run router reflector on master it has to allow to run router pods on all the nodes but has different Underlays, one for control plane with just RR bits and one fork workers with full things.

enhancements/route-reflector.md
qinqon: > Also, I am not quite sure next hop self force is good when we act as RR

Right, I will remove it

enhancements/route-reflector.md
qinqon: > we said it doesn't matter where RR runs, so probably the runOnMaster part is a leftover?

Yep in fact the runOnMaster=true is the default value, I will remove this.

enhancements/route-reflector.md
qinqon: > not sure I would discuss the resulting frr here. Maybe we can add a few snippets on what we are going to add and why, but the whole config is a bit too much imo and we'll certainly drift when doing the real implementation.

It helped me with the PoC, but you are right that we just need snippets with the new stuff, let me address that on the whole document.

enhancements/route-reflector.md
qinqon: > do we really want to test this?

It was interesting, but it adds too much complexity, if we already verify at unit test that the generated configuration do not forward to the ToR then we are covered.

qinqon: > do we really want to test this?

It was interesting, but it adds too much complexity, if we already verify at unit test that the generated configuration do not forward to the ToR then we are covered, I will remove this line

enhancements/route-reflector.md
qinqon: > this means we'll have to change the clab env right?

Nope, now that RR is an Underlay thing, is about a new e2e with different Underlay so we don't have to change anything, the current clab topology should be fine to test RR, I will remove this line.

enhancements/route-reflector.md
qinqon: > is this something we want to e2e?

This is more or less for dual stack, I think we need dual stack for GCP

enhancements/route-reflector.md
fedepaol: yes but as I wrote already, what I am not sure I like is polluting the underlay CR with more and more information. Implementatoin wise it might even be the same controller, just listening to one more CR.

thoughts @maiqueb ?

maiqueb: I actually agree *conceptually* with @fedepaol . 

This needs to make sense holistically. 

IMHO, we need to understand first if we'll add srv6 support in the same underlay CRD vs adding a dedicated one. 

Then, we need to understand if route reflector makes sense in non BGP world. I honestly do not know.

enhancements/route-reflector.md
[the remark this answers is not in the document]
fedepaol: right, maybe we'll have to adjust some bits but the heavylifting is gonna happen in that PR

enhancements/route-reflector.md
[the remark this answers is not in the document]
fedepaol: yep, that was my suggestion

enhancements/route-reflector.md
qinqon: I think we need dualstack for GCP

enhancements/route-reflector.md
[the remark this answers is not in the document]
qinqon: I have adapt the enhancement to depend on that.

enhancements/route-reflector.md
qinqon: Added a section although there is nothing special about it since it's just underlay config

enhancements/route-reflector.md
qinqon: @maiqueb @fedepaol I don't know if new CRD will compliate a little the systemd mode, since it means we have to convert a new CRD or something ? not sure though

also @maiqueb what do you mean with `no BGP world` ?

fedepaol: srv6 is still bgp (meaning, the reacheability information is still spread around as bgp routes)
re: systemd mode, shouldn't be a big deal mio

maiqueb: > srv6 is still bgp (meaning, the reacheability information is still spread around as bgp routes)

I was just browsing the srv6 enhancement without much context, and saw it provides ISIS connections. Hence, I wrongly assumed it would be using ISIS as control plane protocol. My bad.

enhancements/route-reflector.md
fedepaol: makes sense

fedepaol: maybe mention dual stack explicitly?

enhancements/route-reflector.md
fedepaol: isis is just a different way to advertise ips around the network. It can be used to advertise the reacheability of vteps too, while advertising the relation between routes and vtep via regular bgp

fedepaol: @maiqueb @qinqon another option is to land this as part of the underlay, we land srv6 (soon or later) and then we stop, step 1 mt back and think about the api?

enhancements/route-reflector.md
fedepaol: yes but the route reflector section is the same either if a node is RR only or if it's RR + regular router.

fedepaol: so just to clarify, when you write "Its contents" what do you refer to? The content of the underlay or of the RR section? Because I understood the latter and I was probably wrong

enhancements/route-reflector.md
fedepaol: makes sense

enhancements/route-reflector.md
qinqon: > @maiqueb @qinqon another option is to land this as part of the underlay, we land srv6 (soon or later) and then we stop, step 1 mt back and think about the api?

@fedepaol what do you mean ?

fedepaol: what I wrote here. Here we are discussing if it makes sense or not to have the RR section part of the underlay or not. I believe that having the full picture will make things clearer, so I don't want to hold anything with the discussion considering the api can be changed.

enhancements/route-reflector.md
qinqon: I am going to remove that phase for RR only underlays we also need the nic and ASN, not just the routeReflector field.

enhancements/route-reflector.md
qinqon: Ahh so you propose that we put all (rr and srv6) together at the underlay, so we don't stop this enhacement with "to be or not to be CRD", then we review the API as a whole for example at the enhancement related to API improvements ?

fedepaol: right

enhancements/route-reflector.md
```diff
@@ -0,0 +1,373 @@
...
+
+## Design
+
+The primary use case for the internal RR is environments where the ToR cannot be used for east/west traffic. In cloud deployments, control plane and worker nodes are commonly placed on different subnets. The examples throughout this document use this multi-subnet topology as it is the most complex case and the one the RR is originally designed to address.
+
+### How It Works
+
+When the Underlay CR contains a `routeReflector` section, the hostcontroller generates the RR-specific FRR configuration (`bgp listen range`, `route-reflector-client` peer-group, `bgp cluster-id`, inter-RR peers) through the existing template and conversion pipeline.
+
+**Deployment**: Any node with a router pod whose Underlay CR contains a `routeReflector` section becomes an RR. The user is responsible for ensuring the router DaemonSet is scheduled on those nodes (e.g., via tolerations). No separate DaemonSet is ever created.
```
fedepaol: I don't think we need the "No separate Daemonset is ever created" part

enhancements/route-reflector.md
```diff
@@ -0,0 +1,373 @@
...
+
+RR nodes get their own Underlay CR with a `routeReflector` section.
+
+**Client configuration**: Users configure the openperouter instances to connect to the instances configured as route reflectors. The FRR templates already handle iBGP correctly — `allowas-in` is emitted only for eBGP neighbors.
+
+**Inter-RR peering**: Users list all RR node IPs as `type: internal` neighbors in the RR Underlay CR. The hostcontroller on each RR node filters out its own IP and configures the remaining entries as explicit iBGP neighbors. This ensures a full mesh between all RRs. FRR explicit `neighbor` takes precedence over a matching `bgp listen range`, so inter-RR sessions use explicit peering.
```
fedepaol: can't we use the regular neighbor section for this?

enhancements/route-reflector.md
```diff
@@ -0,0 +1,373 @@
...
+
+**`bgp listen range`**: Users explicitly configure all subnets from which iBGP clients may connect via `listenRanges` in the RR Underlay CR. No auto-derivation — all ranges are explicit, which keeps the hostcontroller simple and supports multi-NIC / multi-subnet environments.
```
fedepaol: I think we should also make it work if a user wants to specify a neighbor directly. The same can probably said for allowing listen range to TORs.

In general, I'd split the listen range functionality from the client being a route reflector client (we can probably refine the neighbor data structure)

enhancements/route-reflector.md
```diff
@@ -0,0 +1,373 @@
...
+```
+
+#### Key FRR Additions on RR Nodes
+
+The hostcontroller generates the following RR-specific FRR stanzas through the template and conversion pipeline. These are added on top of the normal router configuration (or standalone for RR-only nodes):
+
+**Cluster ID** — identifies the RR cluster for loop prevention (RFC 4456):
+
+```
+bgp cluster-id 10.0.0.1
```
fedepaol: how would we generate it?

enhancements/route-reflector.md
qinqon: Ok, then I will mark it as resolved, since we are not going to create new CRDs here.

enhancements/route-reflector.md
qinqon: > can't we use the regular neighbor section for this?

Do you mean for listen ? the rest is done using the normal neighbors section.

enhancements/route-reflector.md
qinqon: Maybe we can have all at the neighbor with something like this  ?
```yaml
# RR Underlay (control plane nodes)
spec:
  asn: 64514
  neighbors:
    - asn: 64512                         # ToR (eBGP, explicit)
      address: 192.168.10.2
    - type: internal                     # Dynamic RR clients (listen range)
      cidr: 192.168.10.0/24
      routeReflectorClient: true
    - type: internal                     # Dynamic RR clients (listen range)
      cidr: 192.168.11.0/24
      routeReflectorClient: true
    - type: internal                     # Other RR (explicit, NOT a client)
      address: 192.168.10.4
```

enhancements/route-reflector.md
qinqon: I think we can remove it, FRR defaults to router id and since we are not going to have a massive number of RR nodes having one cluster per node is not that bad and fix the loop issue.

What do you think ?

enhancements/route-reflector.md
fedepaol: either that or we can do what frr does and have a section with neighbor aliases if we want, which is orthogonal to this particular feature (although useless in this context).

Something like 

```
peergroups:
    - name: foo
      asn: 123
      listenRange: / ip:
```

this will be useful to specify which neighbor we want to reflect routes too as we'll have a name for them

enhancements/route-reflector.md
```diff
@@ -0,0 +1,392 @@
...
+type RouteReflectorConfig struct {
+	// listenRanges is the list of CIDRs from which iBGP clients are accepted
+	// via FRR bgp listen range. Each entry maps to a
+	// "bgp listen range <cidr> peer-group CLIENTS" stanza.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=64
+	// +kubebuilder:validation:items:XValidation:rule="isCIDR(self)",message="must be a valid CIDR"
+	// +required
+	// +listType=atomic
+	ListenRanges []string `json:"listenRanges,omitempty"`
```
fedepaol: if we follow my suggestion above, this would be a list of who's a route reflector client. I would keep the door open to specify for which address family, in case we want to reflect only evpn routes and not srv6 or vice versa.

enhancements/route-reflector.md
fedepaol: yep, if it defaults to routerid we can remove it. We are generating routerid from node index, so shoudln't be an issue

enhancements/route-reflector.md
qinqon: What about adding `name` field and neighbor and reference that under `routerReflector` also being able to specify for witch families ? 

```yaml
# RR Underlay (control plane nodes)
spec:
  asn: 64514
  routerReflector:
    clients:
    - peerGroup: RR_CLIENT_1
      families: 
       - ipv4
       - ipv6
       - evpn
    - peerGroup:  RR_CLIENT_2 # For future SRv6
      families: 
      - ipv4vpn
      - ipv6vpn
    
  neighbors:
    - asn: 64512                         # ToR (eBGP, explicit)
      address: 192.168.10.2
    - type: internal                     # Dynamic RR clients (listen range)
      cidr: 192.168.10.0/24
      peerGroup: RR_CLIENTS_1
    - type: internal                     # Dynamic RR clients (listen range)
      cidr: 192.168.11.0/24
      peerGroup: RR_CLIENTS_2
    - type: internal                     # Other RR (explicit, NOT a client)
      address: 192.168.10.4
```

enhancements/route-reflector.md
qinqon: I have add a proposal to above message, tell me if it fits better.

enhancements/route-reflector.md
fedepaol: that's what I was suggesting (kind of). Now I am wondering if it'd make sense to specify what family a neighbor is rr client for instead of having a routeReflector section. 
I think that maps better to what we have in the real configuration

fedepaol: which is what you proposed in https://github.com/openperouter/openperouter/pull/218#discussion_r3303493303 :sweat_smile:

qinqon: > which is what you proposed in [#218 (comment)](https://github.com/openperouter/openperouter/pull/218#discussion_r3303493303) 😅

Well there is one FRR conf requirement related to listen ranges and is the fact that it need to be associated to a peer group name, so if we don't make that explicit openperouter will need generate one for us.

Feels to me that ingroducing peerGroup concept may help with other features in the future, although it adds a new level of indirection, anyhow I am ok either way.

qinqon:
Added back the cluster-id again and also add the taking into accoun the "no-rib" option https://github.com/openperouter/openperouter/compare/ed0cdbdd87b321804568cf7414208b994ee092e8..46b89a2e3b06a934e88513314a28a8e8cb01b62d

fedepaol:
> Added back the cluster-id again and also add the taking into accoun the "no-rib" option https://github.com/openperouter/openperouter/compare/ed0cdbdd87b321804568cf7414208b994ee092e8..46b89a2e3b06a934e88513314a28a8e8cb01b62d

why the cluster-id back?

qinqon:
> > Added back the cluster-id again and also add the taking into accoun the "no-rib" option https://github.com/openperouter/openperouter/compare/ed0cdbdd87b321804568cf7414208b994ee092e8..46b89a2e3b06a934e88513314a28a8e8cb01b62d
> 
> 
> 
> why the cluster-id back?

Without it routes get duplicated per RR node

fedepaol:
> > > Added back the cluster-id again and also add the taking into accoun the "no-rib" option https://github.com/openperouter/openperouter/compare/ed0cdbdd87b321804568cf7414208b994ee092e8..46b89a2e3b06a934e88513314a28a8e8cb01b62d
> > 
> > 
> > why the cluster-id back?
> 
> Without it routes get duplicated per RR node

but you said it defaults to routerid? Isn't that the case?

qinqon:
> > > > Added back the cluster-id again and also add the taking into accoun the "no-rib" option https://github.com/openperouter/openperouter/compare/ed0cdbdd87b321804568cf7414208b994ee092e8..46b89a2e3b06a934e88513314a28a8e8cb01b62d
> > > 
> > > 
> > > why the cluster-id back?
> > 
> > 
> > Without it routes get duplicated per RR node
> 
> but you said it defaults to routerid? Isn't that the case?

So non RR client advertise a route to RR-1 and RR-2
Then RR-1 also send it to RR-2 
RR-2 see is from different cluster so it do not drop it marking it for clusters (router id 1 and router id 2)
Another non RR client will get the route twiche per cluster id since they are different per RR node.

I suppose we want to prevent this, like traffic forwarding would be the same but it will take more resources, cpu and memory.

fedepaol:
> > > > > Added back the cluster-id again and also add the taking into accoun the "no-rib" option https://github.com/openperouter/openperouter/compare/ed0cdbdd87b321804568cf7414208b994ee092e8..46b89a2e3b06a934e88513314a28a8e8cb01b62d
> > > > 
> > > > 
> > > > why the cluster-id back?
> > > 
> > > 
> > > Without it routes get duplicated per RR node
> > 
> > 
> > but you said it defaults to routerid? Isn't that the case?
> 
> So non RR client advertise a route to RR-1 and RR-2 Then RR-1 also send it to RR-2 RR-2 see is from different cluster so it do not drop it marking it for clusters (router id 1 and router id 2) Another non RR client will get the route twiche per cluster id since they are different per RR node.
> 
> I suppose we want to prevent this, like traffic forwarding would be the same but it will take more resources, cpu and memory.

Ah, TIL! Yes, it makes sense.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,413 @@
...
+
+**Client configuration**: Users configure the openperouter instances to connect to the instances configured as route reflectors. The FRR templates already handle iBGP correctly.
+
+**Inter-RR peering**: Standard iBGP — users list all RR node IPs as `type: internal` neighbors in the RR Underlay CR. FRR explicit `neighbor` takes precedence over a matching `bgp listen range`, so inter-RR sessions are not treated as route-reflector clients.
+
+**`bgp listen range`**: Users explicitly configure all subnets from which iBGP clients may connect via `listenRanges` in the RR Underlay CR. No auto-derivation — all ranges are explicit, which keeps the hostcontroller simple and supports multi-NIC / multi-subnet environments.
+
+**`bgp no-rib`**: When the Underlay CR has a `routeReflector` section but no `evpn` section (RR-only mode), the hostcontroller emits `bgp no-rib` in the FRR configuration. This prevents BGP routes from being installed into the kernel RIB, since the node only reflects routes and does not forward traffic. When the Underlay has both `routeReflector` and `evpn` (RR + data plane), `bgp no-rib` is NOT emitted — the node needs kernel routes for VXLAN forwarding.
```
fedepaol: no evpn nor passthrough nor srv6* , meaning the node is a pure RR.
I'd be careful here, maybe we can optimize later on? Or will those routes be harmful?

enhancements/route-reflector.md
fedepaol: in frr being a rr client is a property of neighbor, so the original version feel more natural to me. 

We can still add an alias to refer neighbors by name, both for this scenario and for future ones.

qinqon: Ok going back to 
```yaml
# RR Underlay (control plane nodes)
spec:
  asn: 64514
  neighbors:
    - asn: 64512                         # ToR (eBGP, explicit)
      address: 192.168.10.2
    - type: internal                     # Dynamic RR clients (listen range)
      cidr: 192.168.10.0/24
      routeReflectorClient: true
    - type: internal                     # Dynamic RR clients (listen range)
      cidr: 192.168.11.0/24
      routeReflectorClient: true
    - type: internal                     # Other RR (explicit, NOT a client)
      address: 192.168.10.4
```

enhancements/route-reflector.md
qinqon: It just an optimization since it does not install routes at the kernel, we can create an issue for this as follow up ? or just add a not at the enhancement, whatever you see is better.

enhancements/route-reflector.md
qinqon: @fedepaol do we care about address families ? I see on SRv6 enhancment we are going to incorportea addressFamilies field, maybe we have to integrate with it ?

enhancements/route-reflector.md
fedepaol: yep let's create an issue and keep this out for now

enhancements/route-reflector.md
```diff
@@ -0,0 +1,485 @@
...
+
+	// features is the set of optional features to enable on this neighbor.
+	// EBGPMultiHop: the peer is multiple hops away (RFC 4271).
+	// RouteReflectorClient: the neighbor is a route-reflector client
+	//   (RFC 4456). Applied to all address families activated on this
+	//   neighbor. Per-AF control may be added in a future version
+	//   (see PR #316 review discussion).
+	// +optional
+	// +listType=set
+	Features []NeighborFeature `json:"features,omitempty"`
```
fedepaol: let's call this "property"

enhancements/route-reflector.md
```diff
@@ -0,0 +1,485 @@
...
+	// +kubebuilder:validation:XValidation:rule="isIP(self)",message="must be a valid IPv4 address"
+	// +optional
+	ClusterID *string `json:"clusterID,omitempty"`
+}
+
+// NeighborFeature defines an optional boolean feature on a Neighbor.
+// Values are applied at the session level or to all activated address
+// families depending on the feature.
+// +kubebuilder:validation:Enum=EBGPMultiHop;RouteReflectorClient
+type NeighborFeature string
```
fedepaol: I'd use a struct. Also, a neighbor can be rr clietn only for a specific address family, and we'll add address family we enable neighbors to, not sure what's the best way to express this. '
We can simplify by saying that a neighbor is a rr client for all teh families it's enabled to, but having a struct will allow us to change that.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,485 @@
...
+type Neighbor struct {
+	// ...existing fields (type, address, asn, etc.)...
+
+	// cidr is a CIDR range for dynamic neighbor acceptance via FRR
+	// bgp listen range. When set, the hostcontroller generates a
+	// "bgp listen range <cidr> peer-group <name>" stanza instead of
+	// an explicit neighbor statement. Mutually exclusive with address.
+	// +kubebuilder:validation:XValidation:rule="isCIDR(self)",message="must be a valid CIDR"
+	// +optional
+	CIDR *string `json:"cidr,omitempty"`
```
fedepaol: I'd call this "listenRange", no need to reinvent

enhancements/route-reflector.md
qinqon: > I'd use a struct. Also, a neighbor can be rr clietn only for a specific address family, and we'll add address family we enable neighbors to, not sure what's the best way to express this. ' We can simplify by saying that a neighbor is a rr client for all teh families it's enabled to, but having a struct will allow us to change that.

From frr conf perspective this features are per address family/neighbor so we need to easily do the following without repeat ourselves:
- we want a feature/option for all the familied specified at neighbor
- we want a feature option only for a specific family subset

So with this in mind and taking into account SRv6 enhancement maybe we can do the following

```yaml
- type: Internal
  addressFamilies:
    - type: evpn
      options: [routeReflectorClient, ebgpMultiHop]   # per-AF
    - type: ipv4unicast
  options: [ebgpMultiHop]                              # global neighbor-level
```
```yaml
- type: Internal
  cidr: 192.168.10.0/24
  addressFamilies:
    - type: evpn
      options: [routeReflectorClient]
```

fedepaol: yep I like it!

qinqon: Let's also make options a struct so we can add field to them too in the future.
```yaml
neighbors:
  - type: internal
    address: fd00:10::3
    options:                         # GLOBAL / session-level (whole neighbor)
      - type: ebgpMultiHop
        ebgpMultiHop:
          ttl: 3
    addressFamilies:
      - type: evpn
        options:                     # PER-AF (this address family only)
          - type: routeReflectorClient
      - type: ipv6unicast
        options:
          - type: nextHopSelf
```

Well we can implement this like the following so we don't need to make them structs, but having struct is more flexible and also more CEL friendly, I will go with structs:

```yaml
neighbors:
  - type: internal
    address: fd00:10::3
    options: [ebgpMultiHop]                        # GLOBAL / session-level (whole neighbor)
    ebgpMultiHop:
      ttl: 3
    addressFamilies:
      - type: evpn
        options: [routeReflectorClient]                    # PER-AF (this address family only)
      - type: ipv6unicast
        options: [nextHopSelf]
```

fedepaol: @andreaskaris mind having a look here?

andreaskaris: ~~I'd go with addressFamily.name, and not with .type, but I've got nothing to back that up.~~
@qinqon seems to have a stronger opinion for type (and arguments) than I have for name, so type it is: https://github.com/openperouter/openperouter/pull/316#discussion_r3360578683

Technically, the type field combines the AFI and SAFI, right?  So IPv4 + unicast, IPv4+VPN, EVPN (l2vpn + evpn), but I think that's o.k. and I'd go with the same approach of combining them into one name.
addressfamilies are enforced to be unique (not repeat) per neighbor, right? So only one evpn, only one ipv4unicast, only one ipv4vpn, etc ..

--------------------

I prefer the struct approach, this here looks redundant and confusing:

```
    options: [ebgpMultiHop]                        # GLOBAL / session-level (whole neighbor)
    ebgpMultiHop:
      ttl: 3
```

andreaskaris: Questions I'd have with the struct options:
a) when and why is it and option, and when is it an actual field?

Why would e.g.:
```
neighbor PEER ttl-security hops NUMBER
```

become an option, but why is remote-as a field of its own?
```
neighbor PEER remote-as ASN
```

Of course that distinction makes sense, I'm just curious what the criteria is for promoting one to an actual field vs an option

b) I also think in the case of the example, I'd do:
```
   options:                        
      - type: ebgpMultiHop
      - type: ttlSecurity
        hops: 3
```
where ebgpMultiHop and ttlSecurityHops are mutually exclusive. I.e., instead of inventing our own structure, let's mimic FRR's interface as closely as possible with the options. It's IMO less confusing to admins to stick to FRR syntax.

c) Another question then is: how to deal with FRR configurations that are more complex - I guess we deal with them on a case by case basis when / if the need arises?

```
neighbor PEER send-community <both|all|extended|standard|large>
neighbor PEER send-community extended rpki
```

```
   options:                        
      - type: sendCommunity
        communityType: all
        extended: "rpki"
```



And this here:
```
neighbor <A.B.C.D|X:X::X:X|WORD> path-attribute discard (1-255)...
neighbor <A.B.C.D|X:X::X:X|WORD> path-attribute treat-as-withdraw (1-255)
```

```
   options:                        
      - type: pathAttribute
        discard: 200
        treat-as-withdraw: 220
```
?

d) Or could we find some automatic conversion mechanism that always works, and then users can just pass through the options to the neighbors? Alternatively, do we need/want a raw/passthrough option so that admins can configure the options they need that aren't covered by the API?

qinqon: > Questions I'd have with the struct options: a) when and why is it and option, and when is it an actual field?
> 
> Why would e.g.:
> 
> ```
> neighbor PEER ttl-security hops NUMBER
> ```
> 
> become an option, but why is remote-as a field of its own?
> 
> ```
> neighbor PEER remote-as ASN
> ```
> 
> Of course that distinction makes sense, I'm just curious what the criteria is for promoting one to an actual field vs an option
> 

Good question!

the `neighbor.asn` and `neighbor.type` are kind of basic configuration mandatory config users will need to configure, opions goes into frr massive knobs optional fields.

> b) I also think in the case of the example, I'd do:
> 
> ```
>    options:                        
>       - type: ebgpMultiHop
>       - type: ttlSecurity
>         hops: 3
> ```
> 
> where ebgpMultiHop and ttlSecurityHops are mutually exclusive. I.e., instead of inventing our own structure, let's mimic FRR's interface as closely as possible with the options. It's IMO less confusing to admins to stick to FRR syntax.
> 

We will implement a CEL to make them mutually exclussive, so something like this is illegal:
```yaml
options:
  - type: ebgpMultiHop
    ebgpMultiHop:
      ttl: 2
  - type: ttlSecurity
    ttlSecurity:
      hops: 3
```

So we will have either 
```yam
options:
  - type: ebgpMultiHop
    ebgpMultiHop:
      ttl: 2
```

Or 
```yaml
options:
  - type: ttlSecurity
    ttlSecurity:
      hops: 3
```

Also I understand `ttlSecurity` are different than `ebgpMultiHop` in the sense that the first one has "hops" intermediate field and the seconda one directly configure TTL (there is no "ttl" field at frr.conf) I think this is a small deviation that is fine.

> c) Another question then is: how to deal with FRR configurations that are more complex - I guess we deal with them on a case by case basis when / if the need arises?
> 

We will do case by case.

> ```
> neighbor PEER send-community <both|all|extended|standard|large>
> neighbor PEER send-community extended rpki
> ```
> 
> ```
>    options:                        
>       - type: sendCommunity
>         communityType: all
>         extended: "rpki"
> ```
> 

That would be something like 
```yaml
options:                        
  - type: sendCommunity
    sendCommunity:
       type: all
       extended: "rpki"
```

Or if just extended
```yaml
options:                        
  - type: sendCommunity
    sendCommunity:
       type: extended
       extended: "rpki"
```

Also for cases like "extended, large" or "standard, large" is better to make type a plural list 
```yaml
options:                        
  - type: sendCommunity
    sendCommunity:
       types: ["extended", "large"]
       extended: "rpki"
```

And then maybe this means that "all" and "both" ar just syntatic suggar and we can replace them with 

**all:**
```yaml
options:                        
  - type: sendCommunity
    sendCommunity:
       types: ["extended", "large", "standard"]
       extended: "rpki"
```

**both:**
```yaml
options:                        
  - type: sendCommunity
    sendCommunity:
       types: ["extended", "standard"]
       extended: "rpki"
```


So I don't think the current approach limit what we will decide, also looks like you **must** configure send-community only per address family, meaning that we cannot share options between per session and per AF, or we do but we validate.

> And this here:
> 
> ```
> neighbor <A.B.C.D|X:X::X:X|WORD> path-attribute discard (1-255)...
> neighbor <A.B.C.D|X:X::X:X|WORD> path-attribute treat-as-withdraw (1-255)
> ```
> 
> ```
>    options:                        
>       - type: pathAttribute
>         discard: 200
>         treat-as-withdraw: 220
> ```
> 
> ?
> 

This one should be something like 
```yaml
options:                        
- type: pathAttribute
  pathAttribute
    discard: 200
    treat-as-withdraw: 220
```

> d) Or could we find some automatic conversion mechanism that always works, and then users can just pass through the options to the neighbors? Alternatively, do we need/want a raw/passthrough option so that admins can configure the options they need that aren't covered by the API?

well that's why we have the Raw config thing right ? maybe for those cases we should has productify that ?

fedepaol: I would distinguish between required fields (ip, asn or equivalent) and extra fields, which can become options (I still think properties is a nicer name).

I would try to be pragmatic here, having a struct seems to leave the door open and if not we'll deal with it.
Allowing users to inject raw configs is not something we want to enable from the API, if not using the unsupported rawconfig to cover gaps until we enable the proper API

qinqon: > I would distinguish between required fields (ip, asn or equivalent) and extra fields, which can become options (I still think properties is a nicer name).
> 

I like `properties` better too, will change to it.

> I would try to be pragmatic here, having a struct seems to leave the door open and if not we'll deal with it. Allowing users to inject raw configs is not something we want to enable from the API, if not using the unsupported rawconfig to cover gaps until we enable the proper API

At the end adding `properties` directly to neighbros will be for cases were it cannot be decuded from the non underlay CRDs like L3VNI, L2VNI and friends.

fedepaol: > > I would distinguish between required fields (ip, asn or equivalent) and extra fields, which can become options (I still think properties is a nicer name).
> 
> I like `properties` better too, will change to it.
> 
> > I would try to be pragmatic here, having a struct seems to leave the door open and if not we'll deal with it. Allowing users to inject raw configs is not something we want to enable from the API, if not using the unsupported rawconfig to cover gaps until we enable the proper API
> 
> At the end adding `properties` directly to neighbros will be for cases were it cannot be decuded from the non underlay CRDs like L3VNI, L2VNI and friends.

It's more that those properties are related to the bgp session imo. The ones you mentioned are related to overlays

fedepaol:
LGTM

fedepaol (approved):

enhancements/route-reflector.md
qinqon: Issue added, I will reference it on the enhancement non-goals to make follow up at implementation if needed:
- https://github.com/openperouter/openperouter/issues/482

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+- **Cloud environments**: the ToR cannot be used for east/west traffic
+- **Hybrid deployments**: depending on the on-prem ToR for route reflection adds latency and cost to every East/West flow
+- **Full-mesh iBGP**: N*(N-1)/2 sessions, does not scale
+
+### Goals
+
+- Enable East/West without relying on the ToR
+- Keep East/West data plane traffic within the cluster (no hairpin through the ToR)
+- HA via multiple RR nodes (typically control plane nodes, 3 in production)
+- No new CRDs — opt-in by adding a `routeReflector` section to the RR nodes' Underlay CR and adding RR node IPs as iBGP neighbors in client Underlay CRs
```
maiqueb, line 25: could you remind me why did we add this goal ? 

trying to understand why we'd want to bloat / increase responsibility of the underlay CR vs introducing a dedicated, RR CR.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+- Configure no-rib if nodes are pure router reflectors so they are not going to forward traffic, 
+  an [issue](https://github.com/openperouter/openperouter/issues/482) has being open to do follow up if needed 
+
+## Design
+
+The primary use case for the internal RR is environments where the ToR cannot be used for east/west traffic. In cloud deployments, control plane and worker nodes are commonly placed on different subnets. The examples throughout this document use this multi-subnet topology as it is the most complex case and the one the RR is originally designed to address.
+
+### How It Works
+
+When the Underlay CR contains a `routeReflector` section, the hostcontroller generates the RR-specific FRR configuration (`bgp listen range`, `route-reflector-client` peer-group, `bgp cluster-id`) through the existing template and conversion pipeline. All RR nodes sharing the same Underlay CR get the same `clusterID` value (default `10.0.1.1`), which is required so that inter-RR CLUSTER_LIST loop detection works correctly and clients do not receive duplicate routes (RFC 4456 §7).
```
maiqueb, line 39: so let's say I want to deploy 2 different RRs on a subset of workers. 

... I unfortunately forget to specify the clusterID on any of them. I've shot myself in the foot, didn't I ? 
Trying to understand if we can't somehow throw the underlay name into the mix so we get a better default. Maybe I'm overthinking. 

thoughts @qinqon ?

enhancements/route-reflector.md
qinqon: We decided this on this thread already https://github.com/openperouter/openperouter/pull/218#discussion_r3166663240

I think gist is that "Let's do rr and srv6 at underlay now, then at the api enhancement we see as a whole"

maiqueb: perfect, thank you.

enhancements/route-reflector.md
qinqon: so cluster-id is either IPv4 dotted quad or a 32-bit integer, also underlay name is not a good fit, maybe you have different underlays because you need to do specific stuff per node (like different nic names) but you want to share the router reflector config 

The only thing that we can do here is to propertly document clusterID, we can add example of the scenario you mention, since this is customer deciding his topology.

maiqueb: OK.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+	// +listMapKey=type
+	Properties []NeighborProperty `json:"properties,omitempty"`
+
+	// addressFamilies is the list of address families to activate on
+	// this neighbor, each with optional per-AF features.
+	// Owned by PR #316; extended here with per-AF properties.
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	AddressFamilies []NeighborAddressFamily `json:"addressFamilies,omitempty"`
```
maiqueb, line 172: how do these 2 play along ? like, I see in the CEL that routeReflectorClient is per-address-family only, not session-level . 

Why / when would we use the session-level route reflector client property ?

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+| Node | Role | Underlay NIC IPs | VTEP IP |
+|------|------|-----------------|---------|
+| cp-1 | RR (control plane) | 192.168.10.3/24, fd00:10::3/64 | 100.65.0.0 |
+| cp-2 | RR (control plane) | 192.168.10.4/24, fd00:10::4/64 | 100.65.0.1 |
+| worker-1 | client | 192.168.11.5/24, fd00:11::5/64 | 100.65.0.2 |
+| worker-2 | client | 192.168.11.6/24, fd00:11::6/64 | 100.65.0.3 |
+
+#### Underlay CRs — CP nodes with full routing
+
+When control plane nodes participate in the data plane, they get a full Underlay with ToR. Workers include the control plane RR IPs as iBGP neighbors. The control plane Underlay uses `listenRange` neighbors with `routeReflectorClient` in their EVPN address family for dynamic client acceptance, and `address` neighbors for inter-RR peering — each hostcontroller filters out its own IP:
```
maiqueb, line 227: ~~I don't understand this part, sorry.~~

~~who runs the RRs ? I initially thought it would be the control plane nodes, but it can't be , since it seems they have the `routeReflectorClient` property set.~~

**EDIT:** OK now I got it - I was mislead by my understanding of the `routeReflectorClient` property.

enhancements/route-reflector.md
qinqon: if user configure it at neighbor level, we will configure at all the generated AFs

maiqueb: but the CEL explicitly says: 

```
// +kubebuilder:validation:XValidation:rule="!has(self.properties) || !self.properties.exists(o, o.type == 'routeReflectorClient')",message="routeReflectorClient is per-address-family only, not session-level"
```

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+
+RR nodes get their own Underlay CR with a `routeReflector` section.
+
+**Client configuration**: Users configure the openperouter instances to connect to the instances configured as route reflectors. The FRR templates already handle iBGP correctly.
+
+**Inter-RR peering**: Standard iBGP — users list all RR node IPs as `type: internal` neighbors with `address` in the RR Underlay CR. FRR explicit `neighbor` takes precedence over a matching `bgp listen range`, so inter-RR sessions are not treated as route-reflector clients.
+
+**`bgp listen range`**: Users configure iBGP neighbors with `listenRange` (instead of `address`) and `routeReflectorClient` in their EVPN address family properties in the RR Underlay CR. The hostcontroller generates a `bgp listen range` stanza for each `listenRange` neighbor, and marks the corresponding peer-group as `route-reflector-client` in the corresponding address families. All ranges are explicit — no auto-derivation — which keeps the hostcontroller simple and supports multi-NIC / multi-subnet environments.
```
maiqueb: I originally thought `routeReflectorClient` was something you'd define on a node that *wants to connect* to a route reflector. 

It might be me, but I think this property name is a bit misleading. 

FWIW, is there a use case for setting `listenRange` but NOT set `routeReflectorClient` ? can we run just with `listenRange` ?

enhancements/route-reflector.md
qinqon: Humm that's wrong, either we remove that CEL or we make specific Types one for bgp session and others per AF (some of them will be shared).

If we remove the CEL we make users life easier and routerReflectorClient option get added to all the families that get configured

If we add different properties types then user will have to explicity put the families where he wants RR.

enhancements/route-reflector.md
fedepaol: > I originally thought `routeReflectorClient` was something you'd define on a node that _wants to connect_ to a route reflector.
> 
> It might be me, but I think this property name is a bit misleading.
> 

That's frr terminology, so as much as it sounds odd I'd stick with it

> FWIW, is there a use case for setting `listenRange` but NOT set `routeReflectorClient` ? can we run just with `listenRange` ?

Yep, it's a convenient way not to define many peers living on the same subnet

maiqueb: > > I originally thought `routeReflectorClient` was something you'd define on a node that _wants to connect_ to a route reflector.
> > It might be me, but I think this property name is a bit misleading.
> 
> That's frr terminology, so as much as it sounds odd I'd stick with it

I do understand that motivation, but, we have a chance of providing a better, simpler API than FRR. 
Otherwise, we'd just accept FRR config on a CR, and throw that at it, right ?... 

Just a thought, I can live with it.

> 
> > FWIW, is there a use case for setting `listenRange` but NOT set `routeReflectorClient` ? can we run just with `listenRange` ?
> 
> Yep, it's a convenient way not to define many peers living on the same subnet

OK, makes sense then to have separate knobs.

enhancements/route-reflector.md
maiqueb: I vote for removing that CEL expression then. Let's go for simpler UX.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+          properties:
+            - type: routeReflectorClient
+    - type: internal                   # cp-1 (inter-RR, filtered on cp-1)
+      address: 192.168.10.3
+    - type: internal                   # cp-1 (inter-RR, IPv6, filtered on cp-1)
+      address: fd00:10::3
+    - type: internal                   # cp-2 (inter-RR, filtered on cp-2)
+      address: 192.168.10.4
+    - type: internal                   # cp-2 (inter-RR, IPv6, filtered on cp-2)
+      address: fd00:10::4
```
maiqueb, line 315: remind me ... why do we have to establish this mesh in the control plane ?

enhancements/route-reflector.md
fedepaol: > > > I originally thought `routeReflectorClient` was something you'd define on a node that _wants to connect_ to a route reflector.
> > > It might be me, but I think this property name is a bit misleading.
> > 
> > 
> > That's frr terminology, so as much as it sounds odd I'd stick with it
> 
> I do understand that motivation, but, we have a chance of providing a better, simpler API than FRR. Otherwise, we'd just accept FRR config on a CR, and throw that at it, right ?...
> 

There are many reasons why I think the current version is better:

- It's about making users find what they already know.
I think openperouter will be configured by the same people that are used to deal with real routers. And in real routers (other than cisco), you define `route-reflector-clients`.

- It's about having a straightforward conversion between the api and the frr configuration
Literally, the current implementation is a no brainer. If we define what are the rr clients on the client and moving that configuration on the RR would require a translation gymnastic that can add more complexity (ie a controller today ignores configurations that are not selecting the node it runs on). Also also, this would limit us in being able to define RR clients among peers running on the same node.

enhancements/route-reflector.md
qinqon: for  `listenRange` to work one of the parts should specify the listener IP as `address`, in our case is not clear how would be the listenRange who would be the specific, like which one of the control plane nodes will be the listenRange ? 

Since this is just for control-plane nodes we are going to have usually just three node, so this is doable.

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+
+The hostcontroller generates the following RR-specific FRR stanzas through the template and conversion pipeline. These are added on top of the normal router configuration (or standalone for RR-only nodes):
+
+**Cluster ID** — all RR nodes in the same cluster must share the same cluster-id so that CLUSTER_LIST loop detection works and clients do not receive duplicate reflected routes (RFC 4456 §7). The value comes from `routeReflector.clusterID` (default `10.0.1.1`):
+
+```
+bgp cluster-id 10.0.1.1
+```
+
+**Peer-group name generation** — FRR requires a peer-group for `bgp listen range`. The hostcontroller generates one peer-group per `listenRange` neighbor, deriving the name by sanitizing the CIDR: replace `.`, `:`, and `/` with `-`, then prefix with `PG-`. Examples:
```
maiqueb: the examples below do not match for IPv6 - the examples should either have "---" before the subnet mask, OR, you would replace "::" (two colons) with one "-" (dash).

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+### Testing
+
+**Unit tests** (`internal/frr/`, `internal/conversion/`):
+
+- FRR template rendering: RR stanzas (`bgp cluster-id`, `bgp listen range`, `route-reflector-client`) rendered correctly from `listenRange` + `routeReflectorClient` neighbors (golden file tests)
+- Neighbors with `listenRange` produce `bgp listen range` stanzas, neighbors with `address` produce explicit `neighbor` statements
+
+**E2e tests** (`e2etests/tests/`, Ginkgo suite against a 4-node clab cluster):
+
+- **RR configuration lifecycle**: add `routeReflector` section to RR Underlay -> verify RR stanzas present in FRR config on matching nodes
```
maiqueb: isn't this a unit test ?..

enhancements/route-reflector.md
```diff
@@ -0,0 +1,572 @@
...
+- **Underlay iBGP neighbors**: verify that adding RR node IPs as `type: internal` neighbors in client Underlay CR produces correct FRR config
```
maiqueb: likewise. this sounds like a unit test.

enhancements/route-reflector.md
maiqueb: that plus the enhancement states that we're limited to having a SINGLE cluster-id in it. 

... I guess it says that in the name ...

maiqueb (approved):

PR #219, opened 2026-02-05
vet: Make cel expressions vet compatible

qinqon:
/kind bug

The go fmt command is breaking the cel expression by changing "" into
single double quote unsupported char, this change replace the expressions to use
propertly double quote.

Also replacing xvalidation with XValidation since the former end up with
cel expressions not embedded at the generated manifests](api: Make cel expressions fmt compatible and use proper tag)

**Release note**:

```release-note
Generate linux bridge and ovs bridge manifests with CEL expressions
```


maiqueb (approved):
Thanks for this PR. 

Really nice catch.

PR #226, opened 2026-02-10
React to static configuration file changes

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

The configuration provided as files is read only from the beginning. Here we allow users to be able to modify it.

**Release note**:

```release-note
Handle changes in the static configuration files when running in host mode.
```


internal/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,163 @@
...
+	return nil
+}
+
+// watchLoop runs in background goroutine to process file system events
+func (fw *FileWatcher) watchLoop(ctx context.Context) {
+	defer func() {
+		if err := fw.watcher.Close(); err != nil {
+			fw.logger.Error("error closing watcher", "error", err)
+		}
+		fw.logger.Info("file watcher stopped")
```
qinqon: Pending `debounceTimer` is not stopped on shutdown. If the context is cancelled while a timer is pending, the callback fires from an orphaned goroutine. Add to the defer:
```go
if fw.debounceTimer != nil {
    fw.debounceTimer.Stop()
}
```

internal/filewatcher/filewatcher_test.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	testFile := filepath.Join(tmpDir, "debounce-test.yaml")
+
+	// Make rapid successive changes (5 changes within 200ms)
+	for i := 0; i < 5; i++ {
+		err = os.WriteFile(testFile, []byte("content-"+string(rune(i))), 0644)
+		if err != nil {
+			t.Fatalf("failed to write test file iteration %d: %v", i, err)
+		}
```
qinqon: `string(rune(i))` for i=0..4 produces control characters (NUL, SOH, STX...), not printable text. Use `fmt.Sprintf("content-%d", i)` instead.

fedepaol: done

PR #242, opened 2026-02-26
fix(vrf): add unreachable routes to prevent VRF escape

### LINKED ISSUE #227: Add unreachable route to avoid VRF leak
As recommended by the kernel documentation https://docs.kernel.org/networking/vrf.html, a default unreachable route must be added to the VRF, otherwise if there is not match in the VRF routing table, it will fallback on following rules (so the main routing table for instance).

It can be reproduced very easily from your own laptop:

**Prerequisites**
Having a default route in your main routing table; for instance 
```
~ > ip route
default via 192.168.1.1 dev wlp0s20f3 proto dhcp src 192.168.1.20 metric 600
```

**Steps to reproduce**
1.  Create a VRF
```
sudo ip link add test type vrf table 100
sudo ip link set test up
sudo ip address add 192.168.1.20/32 dev test
```
a new rule must have appeared in `ip rule` to make traffic going through a VRF device lookup the right table
```
~ > ip rule
0:	from all lookup local
1000:	from all lookup [l3mdev-table] <- new rule
32766:	from all lookup main
32767:	from all lookup default
 ```
2. Run a ping in the VRF
```
~ > ping -I test 8.8.8.8
PING 8.8.8.8 (8.8.8.8) from 192.168.1.20 test: 56(84) bytes of data.
```
Obviously the traffic is not coming back to the VRF, but we can see that the traffic is going out on my wireless interface (which exists in my main routing table, not my `test` vrf)
```
~ > sudo tcpdump -ni any host 8.8.8.8
tcpdump: data link type LINUX_SLL2
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on any, link-type LINUX_SLL2 (Linux cooked v2), snapshot length 262144 bytes
18:16:36.822338 wlp0s20f3 Out IP 192.168.1.20 > 8.8.8.8: ICMP echo request, id 37662, seq 5, length 64
18:16:36.834788 wlp0s20f3 In  IP 8.8.8.8 > 192.168.1.20: ICMP echo reply, id 37662, seq 5, length 64
```

That's something visible as well with tools such as https://github.com/cilium/pwru

**Solutions**
- Add a default unreachable route as recommended by the kernel's doc
```ip route add table 100 unreachable default metric 4278198272```
or
- Add another rule to prevent lookup to next tables for traffic coming from a VRF (found this solution from this [blogpost](https://stbuehler.de/blog/article/2020/02/29/using_vrf__virtual_routing_and_forwarding__on_linux.html))
```
2000:   from all lookup [l3mdev-table] unreachable
```

### PR DESCRIPTION
**Pull request by fdomain**
This commit ensures that unreachable default routes are added to each VRF, preventing packets from escaping to the main routing table when no match is found.

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

**What this PR does / why we need it**: Fixes #227 

**Release note**:

```release-note
Fix: add unreachable routes to prevent VRF escape
```

### DIFF
--- a/internal/hostnetwork/vrf.go
+++ b/internal/hostnetwork/vrf.go
@@ -6,9 +6,24 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"net"
 
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 )
+
+// Route priority (aka metric) value that must be set on default unreachable routes.
+// This value is the one recommended by the Linux kernel documentation, must be higher than any other route priority used in the system, to ensure that it is only used when no other route matches.
+// See https://docs.kernel.org/networking/vrf.html for more details.
+const UnreachableRoutePriority = 4278198272
+
+var defaultUnreachableRoutes = []struct {
+	family int
+	dst    *net.IPNet
+}{
+	{family: netlink.FAMILY_V4, dst: &net.IPNet{IP: net.IPv4zero, Mask: net.CIDRMask(0, 32)}},
+	{family: netlink.FAMILY_V6, dst: &net.IPNet{IP: net.IPv6zero, Mask: net.CIDRMask(0, 128)}},
+}
 
 // setupVRF creates a new VRF and sets it up.
 func setupVRF(name string) (*netlink.Vrf, error) {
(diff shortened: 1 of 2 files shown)


maiqueb (approved):
Tks for this PR.

Let's wait for @fedepaol 's review as well.

internal/hostnetwork/vrf.go
```diff
@@ -22,9 +29,64 @@ func setupVRF(name string) (*netlink.Vrf, error) {
...
+
+	for _, r := range []familyRoute{
+		{family: netlink.FAMILY_V4, dst: &net.IPNet{IP: net.IPv4zero, Mask: net.CIDRMask(0, 32)}},
+		{family: netlink.FAMILY_V6, dst: &net.IPNet{IP: net.IPv6zero, Mask: net.CIDRMask(0, 128)}},
+	} {
+		found, err := hasUnreachableDefaultRoute(int(vrf.Table), r.family)
+		if err != nil {
+			return fmt.Errorf("failed to get unreachable default route from VRF %s (table %d): %w", vrf.Name, vrf.Table, err)
+		}
+		if !found {
```
fedepaol: you can avoid one level of nesting and return if found here

internal/hostnetwork/vrf.go
```diff
@@ -22,9 +29,64 @@ func setupVRF(name string) (*netlink.Vrf, error) {
 		return nil, fmt.Errorf("could not set link up for VRF %s: %v", name, err)
 	}
 
+	err = ensureUnreachableDefaultRoutes(vrf)
```
fedepaol: if err := ...

internal/hostnetwork/vrf.go
fdomain: indeed, modified

maiqueb (approved):
Tks.

Some e2e tests failed, I wonder if those are flakes. 
Something we need to address ASAP @fedepaol .

fedepaol:
Thanks!

andreaskaris:
This PR is incompatible with and breaks SRv6 setups. Also, doesn't VRF strict mode (/proc/sys/net/vrf/strict_mode = 1) do exactly the same? https://lwn.net/Articles/823946/
strict_mode = 1 is a prerequisite for SRv6

Edit: strict_mode does not affect this behavior. Be it as is may, the unreachable route breaks SRv6 - I'll check on the SRv6 PR what's needed to get this working, e.g. I'll check if the ip rule approach works, instead

PR #260, opened 2026-03-09
Allow "external" and "internal" for BGP peers and allow numeric iBGP

andreaskaris:

/kind feature

Fixes: https://github.com/openperouter/openperouter/issues/267

    Allow "external" and "internal" for BGP peers and allow numeric iBGP
    
    Add support for BGP `remote-as external` and `remote as internal` and
    allow numeric internal BGP sessions.
    For underlay, this enables more flexible BGP peering scenarios where
    the exact remote ASN is unknown respectively it simplifies the L3VPN
    configuration.
    
    This commit introduces a new PeerASN type that can represent either a
    numeric ASN or "external"/"internal" peer type. The API types are
    updated to accept configs without an (Host)ASN but with (Host)Type
    representing external/internal peers.
    
    The PeerASN abstraction provides proper validation and string
    representation for FRR config generation, distinguishing between
    numeric ASNs and the special "external"/"internal" type.

**Release note**:

```release-note
Add support for BGP `remote-as external`, `remote-as internal` and for iBGP with ASNs.
For underlay, this enables more flexible BGP peering scenarios where
the exact remote ASN is unknown respectively it simplifies the L3VPN
configuration.

Potentially breaking API change due to removal of unused field Neighbor.HostASN.
```


internal/frr/config.go
```diff
@@ -75,7 +76,7 @@ type BFDProfile struct {
 
 type NeighborConfig struct {
 	Name          string
-	ASN           uint32
+	ASN           asn.PeerASN
```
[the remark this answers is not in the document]
andreaskaris: I moved PeerASN into the frr package, but didn't change anything else, so far. I'll ping you tomorrow to ask for further clarification - thanks!

andreaskaris:
@fedepaol

fedepaol (commented):
I have one doubt, would it makes sense to validate how routes are echanged at least in one simple scenario with ibgp ? I am afraid we might hit allowas in somewhere

fedepaol (commented):
still need to go through e2es

api/v1alpha1/neighbor.go
```diff
@@ -19,6 +19,10 @@ type Neighbor struct {
 	// +optional
 	Type string `json:"type,omitempty"`
 
+	// NextHopSelf forces the next-hop to be set to this router, for eBGP and iBGP alike.
+	// +optional
+	NextHopSelf bool `json:"nexthopself,omitempty"`
```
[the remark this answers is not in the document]
andreaskaris: done in next push

e2etests/pkg/infra/underlay.go
```diff
@@ -8,7 +8,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var Underlay = v1alpha1.Underlay{
+var underlay = v1alpha1.Underlay{
```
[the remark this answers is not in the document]
andreaskaris: we talked about this in our 1:1, I'll drop that commit

api/v1alpha1/hostsession.go
```diff
@@ -12,11 +12,12 @@ type HostSession struct {
...
 
-	// ASN is the expected AS number for a BGP speaking component running in
-	// the default network namespace. If not set, the ASN field is going to be used.
+	// HostASN is the expected AS number for a BGP speaking component running in
+	// the default network namespace. If not set, or if set to 0, the remote-as is going to be
+	// set to "external".
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=4294967295
-	// +required
+	// +optional
```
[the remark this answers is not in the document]
fedepaol: Sorry if I get back on this. I think being explicit is better in this case, as it'd mimic what users are used to configure on real routers.

So I'd keep either the type OR the neighbor asn as mandatory

e2etests/pkg/infra/testdata/leafkind.tmpl
```diff
@@ -28,7 +28,11 @@ router bgp 64512
  neighbor 192.168.1.4 remote-as 64612
 
 {{- range .Neighbors }}
+{{- if $.IBGP }}
```
fedepaol: I would be explicit and provide the asn from outside instead of the ibgp flag

e2etests/pkg/infra/testdata/leafkind.tmpl
```diff
@@ -42,6 +46,9 @@ router bgp 64512
   neighbor 192.168.1.4 activate
 {{- range .Neighbors }}
   neighbor {{ . }} activate
+{{- if $.IBGP }}
```
fedepaol: here we can have a flag that tells to set next-hop-self (in general, I'd have the flag describe the behavior we want and not the cause)

e2etests/tests/evpn_routes.go
```diff
@@ -420,6 +420,231 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		routers.Dump(ginkgo.GinkgoWriter)
+
+		By("setting iBGP next-hop-self force on leaf kind")
+		nodes, err = k8s.GetNodes(cs)
+		Expect(err).NotTo(HaveOccurred())
+		ibgpForLeafKind(nodes)
+
+		underlay := *infra.Underlay.DeepCopy()
```
fedepaol: given it's used only here, we can have a local iBGPunderlay instead of cloning hte global one. It's going to be easier to reference and we'll be able to see immediately what we are configuring

e2etests/tests/evpn_routes.go
```diff
@@ -420,6 +420,231 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
...
+			By("Creating the test pod")
+			testPod, err = k8s.CreateAgnhostPod(cs, "test-pod", testNamespace)
+			Expect(err).NotTo(HaveOccurred())
+
+			podNode, err = cs.CoreV1().Nodes().Get(context.Background(), testPod.Spec.NodeName, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			nodeSelector := k8s.NodeSelectorForPod(testPod)
+
+			advertisePodToVNI := func(pod *corev1.Pod, vni v1alpha1.L3VNI) []frrk8sapi.FRRConfiguration {
```
fedepaol: we can extract this in a shared function

e2etests/tests/evpn_routes.go
```diff
@@ -420,6 +420,231 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
...
+			By("setting redistribute connected on leaves")
+			redistributeConnectedForLeaf(infra.LeafAConfig)
+			redistributeConnectedForLeaf(infra.LeafBConfig)
+
+			By("Creating the test namespace")
+			_, err := k8s.CreateNamespace(cs, testNamespace)
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Creating the test pod")
+			testPod, err = k8s.CreateAgnhostPod(cs, "test-pod", testNamespace)
```
fedepaol: testPod is never removed I think? Also, name it after the test somehow

e2etests/pkg/infra/testdata/leafkind.tmpl
andreaskaris: done

e2etests/pkg/infra/testdata/leafkind.tmpl
andreaskaris: done

e2etests/tests/evpn_routes.go
andreaskaris: done

e2etests/tests/evpn_routes.go
andreaskaris: named it "test-pod-ibgp", and the namespace "test-namespace-ibgp"

no, it's not explicitly deleted, but the AfterAll deletes the namespace which will remove the pod

btw, this is a copy/paste from the original test in, with some test cases removed, so it's the same there

e2etests/tests/evpn_routes.go
andreaskaris: done

e2etests/tests/evpn_routes.go
fedepaol: ah right, I forgot about it :sweat_smile:

fedepaol:
I think you need to regenerate the golden files

e2etests/tests/evpn_routes.go
```diff
@@ -420,6 +402,246 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
...
+				hostClientIP, err := extractClientIP(res)
+				Expect(err).NotTo(HaveOccurred())
+
+				if hostClientIP != externalHostIP {
+					return fmt.Errorf("curl from %s to %s:8090 returned %s, expected %s", hostName, podIP, clientIP, externalHostIP)
+				}
+				return nil
+			}, 5*time.Minute, 5*time.Second).ShouldNot(HaveOccurred())
+		},
+			Entry("vni red host A ipv4", vniRed, "hostA_red", infra.HostARedIPv4, ipfamily.IPv4),
```
fedepaol: if we have one single entry, no need for table test

internal/frr/templates/neighboripfamily.tmpl
```diff
@@ -1,16 +1,24 @@
...
-{{- if activateNeighborFor "ipv4" .IPFamily }}
+{{- if activateNeighborFor "ipv4" .neighbor.IPFamily }}
   address-family ipv4 unicast
-    neighbor {{.Addr}} activate
-    neighbor {{.Addr}} allowas-in
+    neighbor {{.neighbor.Addr}} activate
+{{- if isEBGP .routerASN .neighbor.ASN }}
+    neighbor {{.neighbor.Addr}} allowas-in
+{{- else }}
+    neighbor {{.neighbor.Addr}} next-hop-self force
```
fedepaol, line 9: this might require some rework when we introduce RR

internal/conversion/frr_conversion.go
```diff
@@ -44,9 +44,14 @@ func APItoFRR(config ApiConfigData, nodeIndex int, logLevel string) (frr.Config,
...
+		neighName, err := neighborName(n)
 		if err != nil {
-			return frr.Config{}, fmt.Errorf("failed to translate underlay neighbor %s to frr, err: %w", neighborName(n), err)
+			return frr.Config{}, fmt.Errorf("could not generate neighbor name, err: %w", err)
+		}
+
+		frrNeigh, err := neighborToFRR(n, neighName)
+		if err != nil {
+			return frr.Config{}, fmt.Errorf("failed to translate underlay neighbor %s to frr, err: %w", neighName, err)
 		}
```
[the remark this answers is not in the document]
andreaskaris: halucination?

andreaskaris:
@fedepaol I think I addressed all of your concerns

andreaskaris:
Rebased today to upstream due to "conflict" (git just couldn't handle 8fdb76a18c89169e440ebca39afdd72c75d61668 in frr_conversion_test.go, even though it wasn't really a conflict)

internal/conversion/frr_conversion.go
```diff
@@ -55,9 +55,14 @@ func APItoFRR(config ApiConfigData, nodeIndex int, logLevel string) (frr.Config,
 	underlayNeighbors := []frr.NeighborConfig{}
 	bfdProfiles := []frr.BFDProfile{}
 	for _, n := range underlay.Spec.Neighbors {
-		frrNeigh, err := neighborToFRR(n)
+		neighName, err := neighborName(n)
```
fedepaol: super nit: let's neighborToFRR generate neighName (we are setting it internally anyhow)

fedepaol:
one last minute nit, lgtm otherwise

internal/conversion/frr_conversion.go
andreaskaris: done

fedepaol:
LGTM

e2etests/pkg/infra/underlay.go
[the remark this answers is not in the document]
maiqueb: so, what's the actual action here - other than dropping the commit ? 

I think we all agree that sharing objects between E2E tests is *generally speaking* not a good idea, and accessing the Underlay without a deep copy means sharing the same object. Question is, what are we going to do about it ?

internal/conversion/frr_conversion.go
```diff
@@ -57,7 +57,7 @@ func APItoFRR(config ApiConfigData, nodeIndex int, logLevel string) (frr.Config,
 	for _, n := range underlay.Spec.Neighbors {
 		frrNeigh, err := neighborToFRR(n)
 		if err != nil {
-			return frr.Config{}, fmt.Errorf("failed to translate underlay neighbor %s to frr, err: %w", neighborName(n), err)
+			return frr.Config{}, fmt.Errorf("failed to translate underlay neighbor to frr, err: %w", err)
```
maiqueb, line 60: you've dropped logging the ASN here because you might not have it at this point in time right ? 

IIUC, if we leave it as was, it will either:
- show the ASN + IP - as it is doing now, when the admin configured the neighbor info with a numeric ASN
- show "external@<ip>" or "internal@<ip>" when the admin configured the host type

I think the log is better as was before.

internal/conversion/frr_conversion.go
```diff
@@ -225,13 +240,10 @@ func l3vniToFRR(vni v1alpha1.L3VNI, routerID string, underlayASN uint32, nodeInd
 }
 
 // createVNIConfig creates a VNI configuration for a specific IP family
-func createVNIConfig(vni v1alpha1.L3VNI, hostIP net.IP, mask net.IPMask, routerID string) frr.L3VNIConfig {
+func createVNIConfig(vni v1alpha1.L3VNI, hostIP net.IP, mask net.IPMask, routerID string, hostASN frr.PeerASN) frr.L3VNIConfig {
```
maiqueb: this information - `hostASN` is already present in the VNI. 

IMHO, this is a symptom surfacing of some clunkiness in the function's signature.

e2etests/pkg/infra/underlay.go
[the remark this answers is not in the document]
fedepaol: in my original intention, the Underlay here is treated as a constant, a global test fixture that we can use- but go doesn't allow us to use constants while allows global objects.  Many languages offer this option trusting the implementers to do the right thing and imo it's a legit trade off.

Because of that, accessing it from multiple test is not terrible - as long as we don't mutate it. The cost of making it more robust is not worth the effort.

But if you think it's better to make the test implementation more robust, we can revert to having a getter or having a local instance of the underlay (or clone it).

internal/conversion/frr_conversion_test.go
```diff
@@ -61,7 +61,140 @@ func TestAPItoFRR(t *testing.T) {
...
+			nodeIndex: 0,
+			underlays: []v1alpha1.Underlay{
+				{
+					Spec: v1alpha1.UnderlaySpec{
+						ASN: 65000,
+						EVPN: &v1alpha1.EVPNConfig{
+							VTEPCIDR: "192.168.1.0/24",
+						},
+						RouterIDCIDR: "10.0.0.0/24",
+						Neighbors:    []v1alpha1.Neighbor{{Address: "192.168.1.1", ASN: 0, Type: "external"}},
```
maiqueb: nit: we could omit the ASN, right ? 

in this scenario, since this value is not in the acceptable range for the neighbor from the API, and we have no other explanation I find this explicit definition confusing.

internal/conversion/frr_conversion_test.go
```diff
@@ -61,7 +61,140 @@ func TestAPItoFRR(t *testing.T) {
...
+					Spec: v1alpha1.UnderlaySpec{
+						ASN: 65000,
+						EVPN: &v1alpha1.EVPNConfig{
+							VTEPCIDR: "192.168.1.0/24",
+						},
+						RouterIDCIDR: "10.0.0.0/24",
+						Neighbors: []v1alpha1.Neighbor{
+							{
+								Address: "192.168.1.1",
+								ASN:     0, Type: "internal",
```
maiqueb: likewise; can't we omit the ASN here ? to make it more compliant w/ the rules defined in the API.

internal/frr/asn.go
```diff
@@ -0,0 +1,74 @@
...
+const (
+	typeNumber asnType = iota
+	typeExternal
+	typeInternal
+)
+
+// NewPeerASN creates a PeerASN from the provided ASN number and type string.
+// If the provided number is larger than 0, it creates a numeric ASN.
+// Otherwise, it creates a type-based ASN with valid values "external", "internal" or "" (alias for "external").
+func NewPeerASN(number uint32, t string) (PeerASN, error) {
```
maiqueb, line 26: I understand this utility, but I get the feeling its a bit wonky, exacerbated by the fact you now pass redundant (but convenient ...) info in the updated `createVNIConfig` function signature.

What about we just make the choice in the caller ? i.e. understand if we are in the numeric AS path, or in the "auto" path (internal vs external).

internal/frr/asn_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+func TestPeerASN(t *testing.T) {
+	tcs := []struct {
+		name             string
+		number           uint32
+		asnType          string
+		expectedString   string
+		referenceASN     uint32
+		expectedExternal bool
+		expectedEqual    bool
```
maiqueb: can we rename this attribute ? I don't understand what it does tbh.

internal/frr/asn_test.go
```diff
@@ -0,0 +1,93 @@
...
+			name:             "numbered (internal ignored)",
+			number:           10,
+			asnType:          "internal",
+			expectedString:   "10",
+			referenceASN:     11,
+			expectedExternal: true,
+			expectedEqual:    false,
+		},
+		{
+			name:             "unnumbered external",
```
maiqueb: I'd take this a step further and indicate this is the default.

internal/frr/asn_test.go
```diff
@@ -0,0 +1,93 @@
...
+			number:           0,
+			asnType:          "",
+			expectedString:   "external",
+			referenceASN:     11,
+			expectedExternal: true,
+			expectedEqual:    false,
+		},
```
maiqueb, line 166: I think this is a duplicate of the previous test.

internal/frr/asn_test.go
```diff
@@ -0,0 +1,93 @@
...
+		{
+			name:             "unnumbered internal",
+			number:           0,
+			asnType:          "internal",
+			expectedString:   "internal",
+			referenceASN:     11,
+			expectedExternal: false,
+			expectedEqual:    false,
+		},
```
maiqueb, line 173: I don't follow this scenario, which means I don't fully understand what the `expectedEqual` knob does.

internal/frr/frr_test.go
```diff
@@ -65,6 +65,92 @@ func TestBasic(t *testing.T) {
...
+			},
+		},
+		VNIs: []L3VNIConfig{
+			{
+				VRF:      "red",
+				ASN:      64512,
+				VNI:      100,
+				RouterID: "10.0.0.1",
+				LocalNeighbor: &NeighborConfig{
+					ASN:      mustNewPeerASNFromNumber(64512),
```
maiqueb, line 227: you could have defined this as "internal", right ? it would have been the same afaiu

e2etests/tests/evpn_routes.go
```diff
@@ -420,6 +402,243 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
...
+		err = Updater.Update(config.Resources{
+			Underlays: []v1alpha1.Underlay{
+				underlay,
+			},
+		})
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	AfterAll(func() {
+		resetLeafKindConfig(nodes)
```
maiqueb, line 541: do we make changes to leafkind in this block's associated `BeforeAll` ?

e2etests/tests/evpn_routes.go
```diff
@@ -420,6 +402,243 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
...
+
+	Context("testing e2e integration between a pod and the blue / red hosts", func() {
+		const testNamespace = "test-namespace-ibgp"
+		var testPod *corev1.Pod
+		var podNode *corev1.Node
+
+		BeforeAll(func() {
+			By("setting redistribute connected on leaves")
+			redistributeConnectedForLeaf(infra.LeafAConfig)
+			redistributeConnectedForLeaf(infra.LeafBConfig)
```
maiqueb: where are these reset ? maybe consider using a `DeferCleanup` ? ...

e2etests/tests/evpn_routes.go
```diff
@@ -420,6 +402,243 @@ var _ = Describe("Routes between bgp and the fabric", Ordered, func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(infra.LeafAConfig.RemovePrefixes()).To(Succeed())
+			Expect(infra.LeafBConfig.RemovePrefixes()).To(Succeed())
+		})
+
+		AfterEach(func() {
+			dumpIfFails(cs)
+		})
+
+		It("should be able to reach the hosts from the test pod and vice versa vni red hat A ipv4", func() {
```
maiqueb: the test name is a bit funky, isn't it ? why do you have "red hat" in it ?

e2etests/tests/sessions.go
```diff
@@ -132,6 +133,209 @@ var _ = Describe("Router Host configuration", Ordered, func() {
...
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("establishes a session with the host and then removes it when deleting the vni", func() {
+			frrConfig, err := frrk8s.ConfigFromHostSession(*vni.Spec.HostSession, vni.Name, func(config *frrk8sapi.FRRConfiguration) {
+				for j := range config.Spec.BGP.Routers {
+					config.Spec.BGP.Routers[j].ASN = 64515
+				}
```
maiqueb, line 167: why do you have to manually set the ASN ? how do you know what would be the right ASN to set ? 

I'm aware we don't have a status sub-resource - maybe this is good motivation to having a non-node centric status sub-resource `FRRConfiguration` object.

maiqueb (commented):
some comments. 

I think the function to generate an ASN is a bit funky. Everything else, LGTM.

andreaskaris:
rebase - addressing comments next

internal/conversion/frr_conversion.go
andreaskaris: Hmm .. yes and no: I dropped the `%s` there because I moved the neighborName() into neighborToFRR. The information is still there, just that it's inside neighborToFRR():

```
		return nil, fmt.Errorf("neighbor at address %s: could not parse ASN configuration, err: %w", n.Address, err)
(...)
		return nil, fmt.Errorf("neighbor %s: failed to find IP family for address %s, %w", neighName, n.Address, err)
(...)
		return nil, fmt.Errorf("neighbor %s: invalid timers, err: %w", neighName, err)
(...)
			return nil, fmt.Errorf("neighbor %s: invalid connecttime %v: %w", neighName, n.ConnectTime.Duration, err)
```

I think it's fine doing that in neighborToFRR because it's a helper anyway and is only called by APItoFRR

internal/conversion/frr_conversion.go
andreaskaris: I think the entire createVNIConfig helper is clunky and isn't really needed -> see next push

internal/conversion/frr_conversion_test.go
andreaskaris: yes. For some reason I found it cleaner with `ASN: 0` but I agree in hindsight it's confusing -> removed

internal/conversion/frr_conversion_test.go
andreaskaris: removed

internal/frr/asn.go
andreaskaris: see above, I got rid of the `createVNIConfig and I think that part is looking better now

the peerASN also gives us a String() representation, an Equal() method and an IsExternalTo method which replaces another (to me) strange call somewhere in the old templating

how strong is your objection to this? I'm just asking before I put some significant thinking into this

internal/frr/asn_test.go
andreaskaris: done check at next push

internal/frr/asn_test.go
andreaskaris: it's just checking that the Equal() method works, so that in unit test testing for equality is correct:

```
// Equal determines if the current and the provided PeerASN are the same.
// This method is needed for cmp.Equal comparisons in unit tests.
func (pa PeerASN) Equal(pb PeerASN) bool {
	return pa.asType == pb.asType && pa.number == pb.number
}
```

internal/frr/asn_test.go
andreaskaris: done thanks

internal/frr/asn_test.go
andreaskaris: done

internal/frr/frr_test.go
andreaskaris: The end behavior should be the same. But the generated FRR configuration is different:

one sets 
```
router bgp 64512
  peer x.x.x.x remote-as 64512 
```

the other one sets
```
router bgp 64512
  peer x.x.x.x remote-as internal
```

I'm mimicking the FRR behavior; I didn't really ask myself why FRR has `remote-as internal` (because it's actually a 1 to 1 mapping .. whereas `remote-as external` does introduce more flexibility)

maiqueb: all good. I originally asked because I thought using `internal` would make the test more explicit, but, I don't think it matters much.

e2etests/tests/evpn_routes.go
andreaskaris: yes, in the nested beforeall ... I copy pasted without sanitizing; I'll drop that level of nesting -> next push

e2etests/tests/evpn_routes.go
andreaskaris: In:
```
		Expect(infra.LeafAConfig.RemovePrefixes()).To(Succeed())
		Expect(infra.LeafBConfig.RemovePrefixes()).To(Succeed())
```
-- I know, that pattern is super confusing, but it's a copy/paste of what already existed

andreaskaris: Note: I don't want to address this here, but wdyt about creating a low prio issue to address this naming confusion wrt RemovePrefixes()? Ideally, it should prob be named Reset()

maiqueb: that'd be perfect.

e2etests/tests/evpn_routes.go
andreaskaris: me parto ...  I probably typed something in the wrong window

e2etests/tests/sessions.go
andreaskaris: I don't think that a status resource or anything else would help here. I'm just reusing `ConfigFromHostSession` . `ConfigFromHostSession` is a convenience method that returns an FRR config that mirrors the provided hostsession, so that FRR-K8s and OpenPERouter can peer.

For this to work, the FRR-K8s ASN is set to hostsession.HostASN and the FRR-K8s Neighbors[0].ASN is set to hostsession.ASN  (--> the values are "flipped").

The other tests (see `Context("with a l3 vni", func() {`) rely on the fact that these 2 are set for the HostSession:
```
					ASN:     64514,
					HostASN: 64515,
```
... and can thus be flipped and set in the FRRConfig.

In this test, however,  `HostASN` is empty and instead `HostType` = `external`:
```
					ASN:     64514,
					HostASN: 0,
                                        HostType: "external",
```

 Therefore, I need to manually provide `64515` for `Neighbors[0].ASN` as it would otherwise set it to `0`.

maiqueb: My "idea" for the status sub-resource was simple - it would report which ASN it connected to when using internal/external BGP connections. 

I think your explanation's solid FWIW.

e2etests/tests/evpn_routes.go
andreaskaris: https://github.com/openperouter/openperouter/pull/327

andreaskaris:
@maiqueb  Can you have a look at the 2 fixup commits (which I'll squash) and give a thumbs up / down for these changes and if you want me to change something else? Also see my comments above. Thanks!

internal/frr/asn_test.go
maiqueb: ok, I was getting confused with the fact the same table is used to test 2 attributes. 

Maybe worth splitting these ?... meaning, test the `Equal` method outside the table. That'd be explicit.

maiqueb:
> @maiqueb Can you have a look at the 2 fixup commits (which I'll squash) and give a thumbs up / down for these changes and if you want me to change something else? Also see my comments above. Thanks!

They look good - you can squash those.

internal/frr/asn_test.go
andreaskaris: done

e2etests/tests/sessions.go
andreaskaris: btw, I'm not against having status sub-resources, I honestly think we lack those and it'd be good to have them (when looking at srv6, I had that thought as well, that it might be nice to have a status sub resource)

internal/frr/asn_test.go
andreaskaris: see https://github.com/openperouter/openperouter/compare/04cc16a3b1b349ac84e753966ff6af87c610208b..75aa619864d5b989db621279ff3237e1db9242d2

andreaskaris:
@maiqueb @fedepaol Let me know if you are o.k. with the changes / lgtm; I hope I addressed everything (either via changes or comments)

maiqueb (approved):
tks

PR #290, opened 2026-03-27
Limit the sudo commands to where they are really needed

fedepaol:

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

Limiting the need for running sudo makes it easier to work on the project / removes the requirement to have agents running sudo.

**Release note**:

```release-note
NONE
```


maiqueb (changes requested):
Plus I think you forgot to add your signoff to most commits (if not all).

fedepaol:
> Plus I think you forgot to add your signoff to most commits (if not all).

which means, we need to put the DCO check in place too

fedepaol:
@maiqueb removed the bridge creation part as I didn't find a nice way to make it work.
I think even without that there's value in the changes provided by this PR, but I am more than open to close it if you think otherwise

maiqueb:
> @maiqueb removed the bridge creation part as I didn't find a nice way to make it work. I think even without that there's value in the changes provided by this PR, but I am more than open to close it if you think otherwise

I think I am OK using that trick (attaching a dummy to the bridge) if it has a decent comment next to it explaining why we're doing that shenanigan. 

FWIW, LGTM as is. Maybe the add dummy to bridge can be done in a follow up PR ?

maiqueb (approved):
lgtm.

PR #304, opened 2026-04-09
Set veth MTU based on underlay NIC for VXLan entry points

### LINKED ISSUE #291: Set the MTU for vxlan entry points
VXLan encapsulation steals 50 bytes from the payload. We should automatically set the MTU of the veths to the one of the underlay nic minus 50.
At the same time we should check the behavior when the bridge already exists, and if this needs to be documented

qinqon:

/kind feature

VXLan encapsulation adds 50 bytes of overhead to each frame. Without
adjusting the MTU of the veth pairs used as VXLan entry points, frames
at the underlay MTU limit will be fragmented or dropped.

This PR automatically sets the MTU of both veth legs (host and namespace
side) to the underlay NIC MTU minus 50 bytes for both L2 and L3 VNI
setups. The underlay MTU is resolved after the underlay is configured by
looking up the interface with the special marker address. When the
underlay MTU cannot be determined (e.g. Multus-managed underlay), the
configuration is skipped gracefully.

Fixes #291

The MTU setting is best-effort: when the underlay interface cannot be
identified (e.g. Multus-managed underlay without the marker address),
the veth MTU is left at the default. This avoids breaking existing
setups where the underlay is managed externally.

**Release note**:

```release-note
Automatically set veth MTU to underlay NIC MTU minus 50 bytes to
account for VXLan encapsulation overhead, preventing frame
fragmentation at MTU boundaries.
```

### DIFF
--- a/internal/hostnetwork/link_properties.go
+++ b/internal/hostnetwork/link_properties.go
@@ -121,6 +121,21 @@ func linkSetUp(l netlink.Link) error {
 	return netlink.LinkSetUp(l)
 }
 
+// linkSetMTU sets the MTU on the link only if it differs from the desired value.
+// This avoids unnecessary RTM_NEWLINK events that can cause FRR to flush neighbor entries.
+func linkSetMTU(l netlink.Link, mtu int) error {
+	currentLink, err := netlink.LinkByIndex(l.Attrs().Index)
+	if err != nil {
+		return fmt.Errorf("linkSetMTU: failed to get link by index: %w", err)
+	}
+
+	if currentLink.Attrs().MTU == mtu {
+		return nil
+	}
+
+	return netlink.LinkSetMTU(l, mtu)
+}
+
 // linkSetMaster sets the master of a link only if it's not already set to the desired master.
 // This avoids unnecessary RTM_NEWLINK events that can cause FRR to flush neighbor entries.
 func linkSetMaster(link, master netlink.Link) error {
--- a/internal/hostnetwork/underlay.go
+++ b/internal/hostnetwork/underlay.go
@@ -133,6 +133,7 @@ func moveUnderlayInterface(ctx context.Context, underlayInterface string, ns net
 		if err := netlink.LinkSetUp(underlay); err != nil {
 			return fmt.Errorf("could not set link up for VRF %s: %v", underlay.Attrs().Name, err)
 		}
+
 		return nil
 	}); err != nil {
 		return err
@@ -163,33 +164,66 @@ func HasUnderlayInterface(namespace string) (bool, error) {
 // findInterfaceWithIP retrieves the interface assigned to the given ip
 // in the given network ns.
 func findInterfaceWithIP(ns netns.NsHandle, ip string) (string, error) {
-	res := ""
+	linkName := ""
 	err := netnamespace.In(ns, func() error {
-		links, err := netlink.LinkList()
+		l, err := findLinkWithIP(ip)
 		if err != nil {
-			return fmt.Errorf("failed to list links: %w", err)
+			return err
 		}
-		for _, l := range links {
-			addr, _ := netlink.AddrList(l, netlink.FAMILY_ALL)
-			slog.Debug("find underlay", "checking link", l.Attrs().Name, "addresses", addr)
-			hasIP, err := interfaceHasIP(l, ip)
-			if err != nil {
-				return err
-			}
-			if hasIP {
-				res = l.Attrs().Name
-				return nil
-			}
+		if l == nil {
+			return nil
 		}
+		linkName = l.Attrs().Name
 		return nil
 	})
 	if err != nil {
 		return "", err
 	}
-	if res != "" {
-		slog.Debug("returning found has ip", "res", res)
-		return res, nil
+	return linkName, nil
+}
+
+// findUnderlayMTU retrieve the underlay interface MTU, it finds the underlay
+// interface by looking with a link with the underlay special IP
+func findUnderlayMTU(ns netns.NsHandle) (int, error) {
+	underlayMTU := 0
+	err := netnamespace.In(ns, func() error {
+		underlayLink, err := findLinkWithIP(underlayInterfaceSpecialAddr)
+		if err != nil {
+			return err
+		}
+		if underlayLink == nil {
+			slog.Info("no underlay link found when finding MTU", "underlay address", underlayInterfaceSpecialAddr)
+
+			return nil
+		}
+		underlayMTU = underlayLink.Attrs().MTU
+		return nil
+
+	})
+	if err != nil {
+		return 0, err
+	}
+	return underlayMTU, nil
+}
+
+// findLinkWithIP retrieves the link assigned to the given ip.
+func findLinkWithIP(ip string) (netlink.Link, error) {
+	links, err := netlink.LinkList()
+	if err != nil {
+		return nil, fmt.Errorf("failed to list links: %w", err)
+	}
+	for _, l := range links {
+		addr, _ := netlink.AddrList(l, netlink.FAMILY_ALL)
+		slog.Debug("find underlay", "checking link", l.Attrs().Name, "addresses", addr)
+		hasIP, err := interfaceHasIP(l, ip)
+		if err != nil {
+			return nil, err
+		}
+		if hasIP {
+			slog.Debug("returning found has ip", "res", l)
+			return l, nil
+		}
 	}
 	slog.Debug("returning not found")
-	return "", nil
+	return nil, nil
 }
--- a/internal/hostnetwork/vni.go
+++ b/internal/hostnetwork/vni.go
@@ -18,6 +18,16 @@ import (
 	"github.com/vishvananda/netns"
 )
 
+const (
+	// VXLanOverhead is the number of bytes added by VXLan encapsulation.
+	VXLanOverhead = 50
+
+	// MinVethMTU is the minimum MTU we will set on the veth.
+	// 1280 is the IPv6 minimum MTU (RFC 8200); the kernel will reject
+	// or disable IPv6 on the link below this.
+	MinVethMTU = 1280
+)
+
 type VNIParams struct {
 	VRF           string `json:"vrf"`
 	TargetNS      string `json:"targetns"`
@@ -105,16 +115,32 @@ func SetupL3VNI(ctx context.Context, params L3VNIParams) error {
 	if errors.As(err, &netlink.LinkNotFoundError{}) {
 		return fmt.Errorf("SetupL3VNI: host veth %s does not exist, cannot setup L3 VNI", vethNames.HostSide)
 	}
+	if err != nil {
+		return fmt.Errorf("SetupL3VNI: failed to get host veth %s: %w", vethNames.HostSide, err)
+	}
 
 	err = assignIPsToInterface(hostVeth, params.HostVeth.HostIPv4, params.HostVeth.HostIPv6)
 	if err != nil {
 		return fmt.Errorf("failed to assign IPs to host veth: %w", err)
 	}
 
+	underlayMTU, err := findUnderlayMTU(ns)
+	if err != nil {
+		return fmt.Errorf("could not find underlay MTU: %w", err)
+	}
+
+	if err := setVethMTUForVXLAN(hostVeth, underlayMTU); err != nil {
+		return fmt.Errorf("SetupL3VNI: failed to set MTU on host veth %s: %w", vethNames.HostSide, err)
+	}
+
 	if err := netnamespace.In(ns, func() error {
 		peVeth, err := netlink.LinkByName(vethNames.NamespaceSide)
 		if err != nil {
 			return fmt.Errorf("could not find peer veth %s in namespace %s: %w", vethNames.NamespaceSide, params.TargetNS, err)
+		}
+
+		if err := setVethMTUForVXLAN(peVeth, underlayMTU); err != nil {
+			return fmt.Errorf("failed to set MTU on pe veth %s: %w", vethNames.NamespaceSide, err)
 		}
 
 		vrf, err := netlink.LinkByName(params.VRF)
@@ -136,6 +162,7 @@ func SetupL3VNI(ctx context.Context, params L3VNIParams) error {
 	}); err != nil {
 		return err
 	}
+
 	return nil
 }
 
@@ -174,6 +201,15 @@ func SetupL2VNI(ctx context.Context, params L2VNIParams) error {
 	}
 	slog.Info("SetupL2VNI: found host veth", "name", vethNames.HostSide, "index", hostVeth.Attrs().Index)
 
+	underlayMTU, err := findUnderlayMTU(ns)
+	if err != nil {
+		return fmt.Errorf("could not find underlay MTU: %w", err)
+	}
+
+	if err := setVethMTUForVXLAN(hostVeth, underlayMTU); err != nil {
+		return fmt.Errorf("SetupL2VNI: failed to set MTU on host veth %s: %w", vethNames.HostSide, err)
+	}
+
 	if params.HostMaster != nil {
 		if err := setupHostMaster(ctx, params, hostVeth); err != nil {
 			return err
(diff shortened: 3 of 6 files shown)


internal/hostnetwork/underlay.go
```diff
@@ -140,6 +140,42 @@ func moveUnderlayInterface(ctx context.Context, underlayInterface string, ns net
 	return nil
 }
 
+// UnderlayMTU returns the MTU of the underlay interface in the given namespace.
+// The underlay interface is identified by the special marker address.
+func UnderlayMTU(namespace string) (int, error) {
```
fedepaol: can't we store the mtu somewhere when we move it instead of looking afterwards? Note this should be compatible with multus mode.

internal/hostnetwork/vni.go
```diff
@@ -105,6 +109,13 @@ func SetupL3VNI(ctx context.Context, params L3VNIParams) error {
 	if errors.As(err, &netlink.LinkNotFoundError{}) {
 		return fmt.Errorf("SetupL3VNI: host veth %s does not exist, cannot setup L3 VNI", vethNames.HostSide)
 	}
+	if err != nil {
+		return fmt.Errorf("SetupL3VNI: failed to get host veth %s: %w", vethNames.HostSide, err)
+	}
+
+	if err := setVethMTUForVXLAN(hostVeth, params.UnderlayMTU); err != nil {
```
fedepaol, line 116: is the mtu required for both sides of the mtu or setting it on the host side is enough?

internal/hostnetwork/underlay.go
qinqon: Good point. Refactored `SetupUnderlay` to return `(int, error)` — the MTU is now captured directly from the link we already have in hand during `moveUnderlayInterface`, both in the first-move and idempotent paths. Removed the separate `UnderlayMTU()` function entirely.

In Multus mode (`UnderlayInterface=""`), `moveUnderlayInterface` is not called, so `SetupUnderlay` returns MTU 0 and the veth MTU configuration is skipped gracefully (same behavior as before since `setVethMTUForVXLAN` is a no-op when `underlayMTU <= VXLanOverhead`).

internal/hostnetwork/vni.go
qinqon: Yes, both sides need the MTU set. Each side of a veth pair has its own independent MTU — setting it on one side does not propagate to the peer. The MTU controls the maximum frame size the local IP stack will generate on that interface.

If we only set the host side, FRR on the namespace side would still see the default MTU (e.g. 1500) and could generate frames at that size. Those frames would traverse the veth fine, but after VXLan encapsulation (adding 50 bytes) they'd exceed the underlay MTU and get fragmented or dropped.

internal/controller/routerconfiguration/host_config.go
```diff
@@ -72,9 +72,18 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
 	}
 
 	slog.InfoContext(ctx, "setting up underlay")
-	if err := hostnetwork.SetupUnderlay(ctx, hostConfig.Underlay); err != nil {
+	underlayMTU, err := hostnetwork.SetupUnderlay(ctx, hostConfig.Underlay)
```
fedepaol, line 75: do you think it is clean to return the mtu from setupunderlay?

qinqon: Clearly no, I will refactor this, I am taking over this PR, is no longer a "good-for-ai" thing.

fedepaol: please note that the direct and maybe crude comment was for the bot, not for you :-)

qinqon: @fedepaol bot is no more, btw is ok to open the netns again to retrieve MTU or do we prefer for SetupUnderlay to return a Underlay struct to with configured underlay lik interface, mtu or whatever we need ?  this way we can rationalize the number of times we open the netns

internal/hostnetwork/link_properties.go
```diff
@@ -121,6 +121,21 @@ func linkSetUp(l netlink.Link) error {
...
 
+// linkSetMTU sets the MTU on the link only if it differs from the desired value.
+// This avoids unnecessary RTM_NEWLINK events that can cause FRR to flush neighbor entries.
+func linkSetMTU(l netlink.Link, mtu int) error {
+	currentLink, err := netlink.LinkByIndex(l.Attrs().Index)
+	if err != nil {
+		return fmt.Errorf("linkSetMTU: failed to get link by index: %w", err)
+	}
+
+	if currentLink.Attrs().MTU == mtu {
```
maiqueb, line 132: nice, tks for not changing anything in the link unless needed.

maiqueb:
@qinqon you need to fix the unit tests.

qinqon:
CI check `feature-needs-docs` failed on commit b5d54fd but appears unrelated to this PR's changes.

The `feature-needs-docs` check is a CI policy gate that enforces PRs labeled `kind/feature` to include documentation updates under the `website/` directory. It fails when a PR carries the `kind/feature` label but does not contain any changes to documentation files under `website/` (excluding auto-generated `api-reference.md`). This is a labeling/process requirement, not a code or test failure — it cannot be resolved by modifying source code. The fix is to either add the required documentation under `website/` or change the PR's `/kind` label to something other than `feature` if the change does not warrant user-facing documentation.

internal/controller/routerconfiguration/host_config.go
```diff
@@ -75,6 +75,19 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
 	if err := hostnetwork.SetupUnderlay(ctx, hostConfig.Underlay); err != nil {
 		return fmt.Errorf("failed to setup underlay: %w", err)
 	}
+
+	underlayMTU, err := hostnetwork.UnderlayMTU(config.targetNamespace)
+	if err != nil {
+		slog.InfoContext(ctx, "could not determine underlay MTU, skipping veth MTU configuration", "error", err)
+	}
```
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist can we resolve this comment ? Was it addressed in the latest push ?

internal/hostnetwork/underlay.go
```diff
@@ -140,6 +140,41 @@ func moveUnderlayInterface(ctx context.Context, underlayInterface string, ns net
...
+	ns, err := netns.GetFromPath(namespace)
+	if err != nil {
+		return 0, fmt.Errorf("UnderlayMTU: failed to find network namespace %s: %w", namespace, err)
+	}
+	defer func() {
+		if err := ns.Close(); err != nil {
+			slog.Error("failed to close namespace", "namespace", namespace, "error", err)
+		}
+	}()
```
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist can we resolve this comment ? Was it addressed in the latest push ?

I don't see any changes from what you had before - if we fail closing the namespace, we only log. Apologies if I missed something.

internal/controller/routerconfiguration/host_config.go
maiqueb: wondering what'll happen now. We all agree returning the MTU from the `SetupUnderlay` fn is clunky since it make the function essentially do two things. 

Are you still trying to improve this ? Is the PR ready to be reviewed ? @qinqon

qinqon: @maiqueb I did have an improvement but it means open netns more, that's why I did the previous question, originaly this was done like this to reduce the number of netns openning.

maiqueb: can't you make the MTU part of the configuration you want to set, and rely on the user's intention ? i.e. throw an error if the actual MTU is different than the MTU you want, and return nil otherwise ? 

That'd be cleaner IMHO, while allowing the netns to be accessed a single time. 
I think that's what @fedepaol is asking for in https://github.com/openperouter/openperouter/pull/304#discussion_r3065543982 .

qinqon: Do you mean making making MTU part of the configuration ? I don't think that's Fede comment, fede comment is about storing somewhere the retrieved MTU and use it so we don't have to look for it after.

maiqueb: if that "somewhere" was the config, isn't it the same ? 

but all in all, I'm OK as long as the code is readable and the net ns is opened once. 
**EDIT:** imho you're making the current code worse since you're now adding another `namespace.In` block, that will access the net namespace twice when the interface needs to be moved.

qinqon: SO I have refactor the PR (human refactor) so now SetupUnderlay returns the underlay link and also inside of it the "setupUnderlay" function also configure the special underlay address and set the link up so we don't need to open the network namespace again, hope this is good, like returning the underlay link with "SetupUnderlay" is not bad pattern, just returning the MTU was kind of ad-hoc

internal/hostnetwork/underlay.go
```diff
@@ -96,48 +101,25 @@ func ensureLoopback(ctx context.Context, ns netns.NsHandle, vtepIP string) error
 }
 
 // moveUnderlayInterface moves the interface to be used for the underlay connectivity in
-// the given namespace.
-func moveUnderlayInterface(ctx context.Context, underlayInterface string, ns netns.NsHandle) error {
+// the given namespace. It returns the MTU of the underlay interface.
```
fedepaol: comment is wrong here

internal/hostnetwork/underlay.go
```diff
@@ -31,36 +31,41 @@ type UnderlayEVPNParams struct {
 	VtepIP string `json:"vtep_ip"`
 }
 
-func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
+// SetupUnderlay configures the underlay interface in the target namespace.
+// It returns the MTU of the underlay interface (0 when in Multus mode where
+// no interface is explicitly moved).
+func SetupUnderlay(ctx context.Context, params UnderlayParams) (netlink.Link, error) {
```
fedepaol: comment is wrong. Also, I don't think making the link float here is worth the save in terms of extra search

internal/hostnetwork/link_properties.go
```diff
@@ -159,49 +174,51 @@ func setNeighSuppression(link netlink.Link) error {
 }
 
 // moveInterfaceToNamespace takes the given interface and moves it into the given namespace.
-func moveInterfaceToNamespace(ctx context.Context, intf string, ns netns.NsHandle) error {
+func moveInterfaceToNamespace(ctx context.Context, intf string, extraAddresses []string, ns netns.NsHandle) (netlink.Link, error) {
```
fedepaol: this should just move and not take care of assigning addresses

internal/hostnetwork/underlay.go
fedepaol: sorry but the end result seems much more complicated than the original form. Would it make sense to keep a cache of nic - mtu (either lazy loaded or fed when we setup the underlay)? We won't have thousands of interfaces and by doing this we wouldn't need to float up the link across multiple layers.

Thoughts?

internal/hostnetwork/link_properties.go
qinqon: I did it to reduce the number of times we have to open the network namespace, I am starting to entertnain the idea that maybe we neez a lazy open artifact that store the open network namespace during the pod lifetime ? this open/close feels unnecesarry.

fedepaol: imo we should optimize for correctness and maintainability first, and optimize only when it's necessary.

Re the "always open first", we can discuss and experiment. One thing for sure is that it will add complexity in tracking the lifecycle of the target ns (vs always having control of the scope), as you don't want to keep open a namespace that was deleted.

qinqon: > imo we should optimize for correctness and maintainability first, and optimize only when it's necessary.
> 
> Re the "always open first", we can discuss and experiment. One thing for sure is that it will add complexity in tracking the lifecycle of the target ns (vs always having control of the scope), as you don't want to keep open a namespace that was deleted.

make sense the only benefict I see is that with that we will not pollute the functions with an extra network namespace handler, but ok.

At the end my main concern for this PR is if we should reduce the number of times we do `namespace.In` or it's ok to just access it again to read MTU later on.

qinqon:
> sorry but the end result seems much more complicated than the original form. Would it make sense to keep a cache of nic - mtu (either lazy loaded or fed when we setup the underlay)? We won't have thousands of interfaces and by doing this we wouldn't need to float up the link across multiple layers.
>
> Thoughts?

I have go back to retrieve the MTU after SetupUnderlay and pass it to VNI configuration

website/content/docs/configuration/evpn.md
```diff
@@ -198,11 +199,20 @@ When you create or update VNI configurations, OpenPERouter automatically:
...
-3. **Enslaves the veth**: the veth is connected to the bridge corresponding to the l2 domain
-4. **Optionally creates a bridge on the host**: if hostmaster.autocreate is set to `true`
-5. **Optionally connects the host veth to the bridge on the host**: if hostmaster.autocreate is set to `true` or name
+3. **Adjusts Veth MTU**: Sets the MTU on both veth legs to account for VXLan encapsulation overhead (see [Automatic Veth MTU Adjustment](#automatic-veth-mtu-adjustment))
+4. **Enslaves the veth**: the veth is connected to the bridge corresponding to the l2 domain
+5. **Optionally creates a bridge on the host**: if hostmaster.autocreate is set to `true`
+6. **Optionally connects the host veth to the bridge on the host**: if hostmaster.autocreate is set to `true` or name
 is set
 
+## Automatic Veth MTU Adjustment
```
fedepaol: I think we are adding too much narrative here for something this simple. A small note is enough imo

internal/hostnetwork/vni.go
```diff
@@ -174,42 +189,57 @@ func SetupL2VNI(ctx context.Context, params L2VNIParams) error {
...
+		return setupL2VNINamespaceSide(params, vethNames)
+	}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// setupL2VNINamespaceSide configures the namespace side of an L2 VNI, including
+// veth MTU, bridge master assignment, gateway IPs, and fixed MAC address.
+func setupL2VNINamespaceSide(params L2VNIParams, vethNames VethNames) error {
```
fedepaol: let's call this "RouterSide"

internal/controller/routerconfiguration/host_config.go
```diff
@@ -75,6 +75,11 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
 	if err := hostnetwork.SetupUnderlay(ctx, hostConfig.Underlay); err != nil {
 		return fmt.Errorf("failed to setup underlay: %w", err)
 	}
+
+	if err := configureMTUs(&hostConfig); err != nil {
```
fedepaol: hostConfig is an artifact of the conversion here, and we hijacking it by adding the mtu which is a netlink concept.

If we go with the LRU approach, I'd rather provide the underlay to SetupL3/L2VNI and do the logic from there. If not, let's discuss an alternative.

internal/hostnetwork/underlay.go
```diff
@@ -69,6 +69,33 @@ func (e UnderlayExistsError) Error() string {
 	return string(e)
 }
 
+func UnderlayMTU(params UnderlayParams) (int, error) {
+	if params.UnderlayInterface == "" {
```
fedepaol: this won't count the multus scenario. Is that because we need to rethink it?

internal/hostnetwork/underlay.go
```diff
@@ -69,6 +69,33 @@ func (e UnderlayExistsError) Error() string {
...
+		return 0, nil
+	}
+	ns, err := netns.GetFromPath(params.TargetNS)
```
fedepaol: does it make sense to have some LRU logic in place here?

internal/hostnetwork/vni.go
qinqon: > let's call this "RouterSide"

the VethNames struct call it NamespaceSide too, that's why this functions follwo that parttern, should I rename both ? (btw this was done to make linter happy related to cyclomatic complexity).

qinqon:
Rebased commit 338a122 on main and pushed.

internal/controller/routerconfiguration/host_config.go
qinqon: I have  manual refactor the PR, now we just find the underlay interface by looking fo rthe underlay special IP when we configure the veths and we take the MTUs from there and configur veths mtus, downside is that we access the network namespace one more time so we can make the code cleaner.

internal/hostnetwork/underlay.go
qinqon: We can do a follow up for multus (if we ever needed) if the code do not find the underlay interface it just do a no-op.

internal/hostnetwork/underlay.go
qinqon: ditto, reading underlay MTU before configuring veth, no need to store MTU anymore.

internal/hostnetwork/vni.go
```diff
@@ -624,6 +662,19 @@ func hostMaster(vni int, m HostMaster) (netlink.Link, error) {
...
+func setVethMTUForVXLAN(link netlink.Link, underlayMTU int) error {
+	if underlayMTU == 0 {
+		slog.Debug("No underlay found")
+		return nil
+	}
+	if underlayMTU > VXLanOverhead {
+		return linkSetMTU(link, underlayMTU-VXLanOverhead)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
fedepaol: not sure this is reasonable, I would say in that case nothing would work.

fedepaol:
I only have one doubt about the gemini comment, lgtm otherwise

internal/hostnetwork/vni.go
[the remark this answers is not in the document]
qinqon: Right but is the kind of things where you waste a week looking for the issue, maybe is not bad to add this here ?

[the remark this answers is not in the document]
fedepaol: yep agree, at least a log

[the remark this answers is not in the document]
fedepaol: also, we'll have to reconsider this behavior when we add support for multiple underlay nics, but I'll merge this first and think about it in the other pr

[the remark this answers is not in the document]
qinqon: So for ipv6 looks like the limit is way bigger, I am going to use the the ipv6 limite and return nil + log, if we see this is a problem we can fix at follow up, but let's play safe here.

fedepaol:
Thanks!

PR #307, opened 2026-04-09
Multiple nics and neighbors

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

**Release note**:

```release-note
Allow moving multiple nics in the OpenPERouter network namespace and allow setting multiple neighbors for the underlay.
```


internal/conversion/validate_underlay.go
```diff
@@ -32,18 +32,48 @@ func ValidateUnderlays(underlays []v1alpha1.Underlay) error {
 	if len(underlays) > 1 {
 		return fmt.Errorf("can't have more than one underlay per node")
 	}
-	return validateUnderlay(&underlays[0])
+	for _, underlay := range underlays {
+		if err := validateUnderlay(underlay); err != nil {
+			return err
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
fedepaol: fixed

internal/hostnetwork/underlay.go
```diff
@@ -98,24 +101,9 @@ func ensureLoopback(ctx context.Context, ns netns.NsHandle, vtepIP string) error
...
-		return UnderlayExistsError(fmt.Sprintf("existing underlay found: %s, new is %s", currentUnderlayInterface, underlayInterface))
-	}
-
-	err = moveInterfaceToNamespace(ctx, underlayInterface, ns)
+	// Check if this specific interface is already in the target namespace.
+	// With multiple underlay interfaces, each is moved independently.
+	err := moveInterfaceToNamespace(ctx, underlayInterface, ns)
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
fedepaol: it's done by removing the ns

internal/hostnetwork/underlay.go
```diff
@@ -44,8 +44,8 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
 		}
 	}()
 
-	if params.UnderlayInterface != "" {
-		if err := moveUnderlayInterface(ctx, params.UnderlayInterface, ns); err != nil {
+	for _, underlayInterface := range params.UnderlayInterfaces {
+		if err := moveUnderlayInterface(ctx, underlayInterface, ns); err != nil {
 			return err
 		}
 	}
```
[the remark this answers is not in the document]
fedepaol: we kill the pod and interfaces get back to the ns

internal/conversion/validate_underlay.go
```diff
@@ -32,18 +32,48 @@ func ValidateUnderlays(underlays []v1alpha1.Underlay) error {
...
+	// Validate at least one neighbor is specified
+	if len(underlay.Spec.Neighbors) == 0 {
+		return fmt.Errorf("underlay %s must have at least one neighbor configured", underlay.Name)
+	}
+
+	// Validate at least one nic is specified
+	if len(underlay.Spec.Nics) == 0 {
+		return fmt.Errorf("underlay %s must have at least one nic configured", underlay.Name)
+	}
+
```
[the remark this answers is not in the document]
fedepaol: removed

clab/singlecluster/leafkind1/frr.conf
```diff
@@ -22,25 +19,26 @@ debug bfd zebra
 
 router bgp 64512
  no bgp ebgp-requires-policy
+ ! Allow BGP to advertise routes not yet present in the local RIB,
```
fedepaol: comment not required

clab/singlecluster/leafkind1/frr.conf
```diff
@@ -22,25 +19,26 @@ debug bfd zebra
...
  no bgp default ipv4-unicast
 
  neighbor 192.168.1.4 remote-as 64612
-
- neighbor kind-nodes peer-group
- neighbor kind-nodes remote-as 64514
-
- bgp listen range 192.168.11.0/24 peer-group kind-nodes
+ neighbor 192.168.11.3 remote-as 64514
+ neighbor 192.168.11.4 remote-as 64514
```
fedepaol: we can use peer-group as the original version

e2etests/tests/hotapply.go
```diff
@@ -0,0 +1,138 @@
...
+
+		By("Verifying router pods were NOT restarted")
+		Consistently(func() error {
+			currentRouters, err := openperouter.Get(cs, HostMode)
+			if err != nil {
+				return err
+			}
+			// DaemonsetRolled returns nil when pods HAVE been rolled.
+			// We expect it to return an error (pods NOT rolled = same names).
+			if err := openperouter.DaemonsetRolled(initialRouters, currentRouters); err == nil {
```
fedepaol: we can flip this check to check the error Should HaveOccurred

website/content/docs/configuration/multi-underlay.md
```diff
@@ -0,0 +1,452 @@
+---
```
fedepaol: we can remove this file

fedepaol:
@maiqueb I think regardless of the CI result this can be reviewed

api/v1alpha1/underlay_types.go
maiqueb, line 1: I don't get this commit msg. The behavior was already there. You're just updating the comment.

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -56,6 +56,7 @@ type PERouterReconciler struct {
 
 	// TriggerChan receives events from FileWatcher (in host mode)
 	TriggerChan chan event.GenericEvent
+
```
maiqueb: not needed right ?

internal/conversion/host_conversion.go
```diff
@@ -34,11 +34,10 @@ func APItoHostConfig(nodeIndex int, targetNS string, underlayFromMultus bool, ap
...
 
 	res.Underlay = hostnetwork.UnderlayParams{
-		TargetNS: targetNS,
-	}
-	if len(underlay.Spec.Nics) > 0 {
-		res.Underlay.UnderlayInterface = underlay.Spec.Nics[0]
+		TargetNS:           targetNS,
+		UnderlayInterfaces: make([]string, len(underlay.Spec.Nics)),
 	}
+	copy(res.Underlay.UnderlayInterfaces, underlay.Spec.Nics)
```
maiqueb, line 41: I somehow (might be me & my vices ...) consider the previous iteration more readable ...

How about something like:
```go
		TargetNS:           targetNS,
    }
	if len(underlay.Spec.Nics) > 0 {
            res.Underlay.UnderlayInterfaces = underlay.Spec.Nics
    }
```

internal/conversion/host_conversion_test.go
```diff
@@ -70,7 +70,45 @@ func TestAPItoHostConfig(t *testing.T) {
...
+			wantL2VNIParams: []hostnetwork.L2VNIParams{},
+			wantPassthrough: nil,
+			wantErr:         false,
+		},
+		{
+			name:      "two underlay interfaces",
+			nodeIndex: 0,
+			targetNS:  "namespace",
+			underlays: []v1alpha1.Underlay{
+				{Spec: v1alpha1.UnderlaySpec{Nics: []string{"eth0", "eth1"}, EVPN: &v1alpha1.EVPNConfig{VTEPCIDR: "10.0.0.0/24"}}},
```
maiqueb: let's have one per line. I'm not sure what's wrong with me, but this is super hard to read.

```suggestion
				{
				        Spec: v1alpha1.UnderlaySpec{
				                Nics: []string{"eth0", "eth1"},
				                EVPN: &v1alpha1.EVPNConfig{
				                        VTEPCIDR: "10.0.0.0/24",
				                },
				        },
				},
```

internal/conversion/host_conversion_test.go
```diff
@@ -70,7 +70,45 @@ func TestAPItoHostConfig(t *testing.T) {
...
+			},
+			vnis: []v1alpha1.L3VNI{
+				{Spec: v1alpha1.L3VNISpec{VRF: "red", HostSession: &v1alpha1.HostSession{LocalCIDR: v1alpha1.LocalCIDRConfig{IPv4: "10.1.0.0/24"}}, VNI: 100, VXLanPort: 4789}},
```
maiqueb: ditto. I can't see anything here. 

Which reminds me, we should have a code formatter enabled for the project right ?

internal/conversion/validate_underlay.go
```diff
@@ -32,33 +32,43 @@ func ValidateUnderlays(underlays []v1alpha1.Underlay) error {
 	if len(underlays) > 1 {
 		return fmt.Errorf("can't have more than one underlay per node")
 	}
-	return validateUnderlay(&underlays[0])
+	return validateUnderlay(underlays[0])
```
maiqueb, line 35: so ... this function is only invoked from the webhook path, is that it ? 

I was expecting the len underlay == 1 limitation to be removed, given the commit msg: 
> Convert underlay setup to handle multiple interfaces with CIDR-based marker IPs for tracking.

If I'm right, would you please clarify in the commit msg that the webhook path will be updated in a follow up PR ?

internal/conversion/validate_underlay.go
```diff
@@ -80,3 +90,22 @@ func validateUnderlayEVPN(underlay *v1alpha1.Underlay) error {
...
+func neighborAddressesOf(neighbors []v1alpha1.Neighbor) []string {
+	res := make([]string, len(neighbors))
+	for i, n := range neighbors {
+		res[i] = n.Address
+	}
+	return res
+}
+
+func checkDuplicates(items []string) (string, bool) {
+	seen := make(map[string]bool, len(items))
```
maiqueb: do we actually need the boolean for anything ? couldn't it be a `map[string]struct{}` instead, which I think is more memory efficient ?

internal/conversion/validate_underlay_test.go
```diff
@@ -144,8 +192,15 @@ func TestValidateUnderlay(t *testing.T) {
...
 					ASN:  65001,
+					Neighbors: []v1alpha1.Neighbor{
+						{
+							ASN:     65002,
+							Address: "192.168.1.1",
+						},
+					},
 				},
 			},
+			wantErr: false,
```
maiqueb, line 165: this is totally unrelated, but I agree consistency is good. Let's keep it.

api/v1alpha1/underlay_types.go
```diff
@@ -39,7 +39,10 @@ type UnderlaySpec struct {
 	// +optional
 	RouterIDCIDR string `json:"routeridcidr,omitempty"`
 
-	// Neighbors is the list of external neighbors to peer with.
+	// Neighbors is the list of external BGP neighbors to peer with.
```
qinqon: I don't understand this commit we already  have a slice, maybe we should reword the commit message ?

internal/conversion/validate_underlay.go
```diff
@@ -32,33 +32,43 @@ func ValidateUnderlays(underlays []v1alpha1.Underlay) error {
...
 	if underlay.Spec.ASN == 0 {
 		return fmt.Errorf("underlay %s must have a valid ASN", underlay.Name)
 	}
 
+	// Validate at least one neighbor is specified
+	if len(underlay.Spec.Neighbors) == 0 {
+		return fmt.Errorf("underlay %s must have at least one neighbor configured", underlay.Name)
+	}
+
+	if dup, found := checkDuplicates(neighborAddressesOf(underlay.Spec.Neighbors)); found {
```
qinqon: Do you know if using golagn Iter we can reduce iterations here ? (maybe is not even worth it)

internal/hostnetwork/underlay.go
```diff
@@ -44,8 +47,37 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
+			}
+			underlayMarkerIP := fmt.Sprintf("172.16.1.%d/32", i+1)
+			if err := assignIPToInterface(underlay, underlayMarkerIP); err != nil {
+				return err
+			}
+			if err := netlink.LinkSetUp(underlay); err != nil {
+				return fmt.Errorf("could not set link up for VRF %s: %v", underlay.Attrs().Name, err)
+			}
+			return nil
+		}); err != nil {
```
qinqon: This is sensitive to UnderlayInterfaces order or updates (don't know if we allow updates) order is not going to change never ever right ? maybe is better if we use the underlay interface name to compute the address ?

qinqon: btw have we check if using "altname" to mark interfaces is maybe better than configuring ip addresses ? 
```
ip link property add dev eth0 altname someothername
```
I think we have already comment it on the past, but I don't remember, good think is that is totally indepotent since all the nics will have the same altname and you don't have to calculate the fake IP, although it will not survive reboots, but we have same issue with IPs upstreams (I think).

Another good thing with altnames is that we can use `netlink.LinkByName` so we get all the underlay interfaces

internal/hostnetwork/underlay.go
```diff
@@ -153,43 +143,53 @@ func HasUnderlayInterface(namespace string) (bool, error) {
...
 		}
 		return nil
 	})
+	return result, err
+}
+
+// IsUnderlayMarkerIP returns true if the given address falls within
+// the underlay marker CIDR range.
+func IsUnderlayMarkerIP(addr string) bool {
+	_, ipNet, err := net.ParseCIDR(underlayMarkerCIDR)
```
qinqon: nit: we may want to improve this by Parsing the underlayMarkerCIDR just ones at "init" block

qinqon (changes requested):
first pass

maiqueb (commented):
pushing my unfinished review.

clab/singlecluster/leafkind/frr.conf
```diff
@@ -1,8 +1,5 @@
 log file /etc/frr/frr.log debug
 
-debug zebra events
-debug zebra vxlan
-debug bgp zebra
```
maiqueb, line 5: why were these deleted ?

maiqueb: oh they're duplicates, right ?

clab/singlecluster/leafkind2/frr.conf
maiqueb, line 1: can't leafkind be a template, and we'd just fill the parts that need to be different ? 

I feel we've done that for every other feature but this one.

e2etests/tests/singlesession.go
maiqueb, line 1: this test only features an L3 VNI. 

I wonder why L2 VNIs don't have a counterpart. It doesn't benefit from redundant links in the fabric ?

e2etests/tests/singlesession.go
```diff
@@ -0,0 +1,201 @@
...
+			LocalCIDR: v1alpha1.LocalCIDRConfig{
+				IPv4: "192.169.10.0/24",
+				IPv6: "2001:db8:169:10::/64",
+			},
+		},
+		VNI: 100,
+	},
+}
+
+var _ = Describe("Single Session Baseline", Ordered, func() {
```
maiqueb, line 87: what's the purpose of these tests ? this is what we had before this PR right ? just a single underlay neighbor. 

I wonder then why don't we try to establish baselines for L2 VNIs.

maiqueb (changes requested):
Rest of my review. 

Tks for this PR.

internal/conversion/host_conversion.go
fedepaol: It's not about readability, copy is safer considering that underlay.Spec is coming from the informer and if we just assign we are doing a shallow copy. I can loop over the items if you prefer, but the assignement is not semantically equivalent

internal/conversion/host_conversion_test.go
fedepaol: I think we do. I'll expand both

internal/conversion/validate_underlay.go
fedepaol: We handle multiple interfaces, yes. But I think it's still easier to allow one single underlay containing the underlay related configuration vs having one interface defined in one underlay instance and another interface (or neighbor) defined in another instance.

This function is invoked both on reconcile and from the webhook.

internal/conversion/validate_underlay.go
fedepaol: we can use empty struct, yes. I'll replace it

internal/conversion/validate_underlay.go
fedepaol: I don't think it's worth the effort tbh

internal/hostnetwork/underlay.go
fedepaol: do we really care? We use the ip as a marker, and what matters is if it belongs to a given cidr.

Altname is a viable alternative, I can check how it works (and if it's too complex I can defer it)

internal/hostnetwork/underlay.go
fedepaol: yeah will do

clab/singlecluster/leafkind/frr.conf
fedepaol: right

clab/singlecluster/leafkind2/frr.conf
fedepaol: yep, will do

e2etests/tests/singlesession.go
fedepaol: I took just one to validate the session works, but I can (and will) easily expand to L2VNI too

internal/conversion/host_conversion.go
maiqueb: OK, I though it was the same, and I usually rely on deep copy ...

Would that make sense ? still, don't bother changing this.

internal/controller/routerconfiguration/underlay_vni_controller.go
fedepaol: right

internal/conversion/host_conversion.go
fedepaol: yep, deepcopy addresses that for k8s types, but I don't think there's an equivalent for slices (and probably uses copy under the hood).

api/v1alpha1/underlay_types.go
fedepaol: yes, it's the result of an initial iteration, squashed with the implementation

internal/hostnetwork/underlay.go
fedepaol: done

api/v1alpha1/underlay_types.go
fedepaol: squashed the commit

internal/conversion/host_conversion_test.go
fedepaol: done, and will file a pr for enforcing formatting in CI

clab/singlecluster/leafkind2/frr.conf
fedepaol: done

e2etests/tests/singlesession.go
fedepaol: done

internal/hostnetwork/underlay.go
fedepaol: switched to altname, much more elegant!

internal/hostnetwork/underlay.go
```diff
@@ -44,8 +46,39 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
+			return err
+		}
+		if err := netnamespace.In(ns, func() error {
+			underlay, err := netlink.LinkByName(underlayInterface)
+			if err != nil {
+				return fmt.Errorf("failed to get underlay nic by name %s: %w", underlayInterface, err)
+			}
+			// Add altname only if it doesn't already exist (idempotent)
+			if !slices.Contains(underlay.Attrs().AltNames, underlayMarkerAltName) {
+				if err := netlink.LinkAddAltName(underlay, underlayMarkerAltName); err != nil {
```
qinqon: Kernel do not allow to have same altname for multiple nics you will need to add something like the link index, name or random to this, you can still search for substring.

Looks like we have even something call link group id, maybe we can use that ? so we mark all the underlay nics with the same group id ? 
```golang
package main
import (
	"fmt"
	"log"
	"github.com/vishvananda/netlink"
)
const underlayGroup uint32 = 4242
func main() {
	// Tag a link as belonging to the underlay group.
	link, err := netlink.LinkByName("eth1")
	if err != nil {
		log.Fatal(err)
	}
	if err := netlink.LinkSetGroup(link, int(underlayGroup)); err != nil {
		log.Fatal(err)
	}
	// List all links in the current netns, filter by group.
	links, err := netlink.LinkList()
	if err != nil {
		log.Fatal(err)
	}
	for _, l := range links {
		if l.Attrs().Group == underlayGroup {
			fmt.Println("underlay nic:", l.Attrs().Name)
		}
	}
}
```

Equivalent CLI: ip link set eth1 group 4242 then ip link show group 4242.

qinqon (changes requested):
alt name has to be unique per nic

internal/hostnetwork/underlay.go
fedepaol: yep I wasn't sure so I waited for CI to tell me, fixing it

qinqon: @fedepaol so are you going to use alt name or group id ? feels to me that group id is exactly what we need.

fedepaol: I went with altname + interface name, now focusing on making it work and then I'll experiment with group id

fedepaol:
@maiqueb @qinqon this is ready for another round

e2etests/tests/sessions.go
```diff
@@ -76,16 +76,22 @@ var _ = Describe("Router Host configuration", Ordered, func() {
...
 	})
 
-	validateTORSession := func() {
-		exec := executor.ForContainer(infra.KindLeaf)
-		for _, node := range nodes {
-			neighborIP, err := infra.NeighborIP(infra.KindLeaf, node.Name)
-			Expect(err).NotTo(HaveOccurred())
-			validateSessionWithNeighbor(infra.KindLeaf, node.Name, exec, neighborIP, Established)
+	validateTORSessions := func() {
+		leaves := []string{infra.KindLeaf, infra.KindLeaf2}
```
maiqueb, line 80: nit: I wonder if it shouldn't be test to push its topology to the helper function. 

Not important, you can ignore.

e2etests/tests/singlesession.go
maiqueb: @fedepaol ping

maiqueb: note to self, now this tests feature L2VNI, and it asserts both for:
- east / west traffic 
- north / south traffic (when exposed via an L3VNI)

internal/conversion/validate_underlay.go
maiqueb: All good. I'm OK with choosing one way of configuring multiple neighbors, and stick to that. 

I don't care which - 1 underlay w/ multiple neighbors is OK for me.

internal/conversion/validate_underlay.go
```diff
@@ -80,3 +90,22 @@ func validateUnderlayEVPN(underlay *v1alpha1.Underlay) error {
...
+
+func neighborAddressesOf(neighbors []v1alpha1.Neighbor) []string {
+	res := make([]string, len(neighbors))
+	for i, n := range neighbors {
+		res[i] = n.Address
+	}
+	return res
+}
+
+func checkDuplicates(items []string) (string, bool) {
```
maiqueb: nit: I know you're trying to have nice semantics, but here, multiple return types are not needed. Or if they are, let's use an error as the second return type. 

you can use empty string (you can even use a const for that ...) as no duplicates found (we won't have an empty string in the items list, right ?

internal/conversion/validate_underlay_test.go
```diff
@@ -75,32 +99,50 @@ func TestValidateUnderlay(t *testing.T) {
...
+							Address: "192.168.1.1",
+						},
+					},
 				},
 			},
 			wantErr: true,
 		},
 		{
-			name: "zero nics",
+			name: "zero nics (valid for Multus)",
```
maiqueb: this no longer makes sense, right ? i.e. we dropped multus support in #317

internal/conversion/validate_underlay_test.go
```diff
@@ -75,32 +99,50 @@ func TestValidateUnderlay(t *testing.T) {
...
+							Address: "192.168.1.1",
+						},
+					},
 				},
 			},
 			wantErr: false,
 		},
 		{
-			name: "valid underlay with no nics",
+			name: "nil nics (valid for Multus)",
```
maiqueb: likewise

website/content/docs/configuration/_index.md
```diff
@@ -38,11 +38,50 @@ spec:
...
+      address: 192.168.11.3
+    - asn: 64513
+      address: 192.168.12.2
+    - asn: 64513
+      address: 192.168.12.3
+```
+
+**Validation requirements**:
+- At least one neighbor must be configured
+- At least one NIC must be configured (unless using Multus)
```
maiqueb: no longer true. multus is gone.

internal/hostnetwork/underlay.go
```diff
@@ -44,8 +46,39 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
+			underlay, err := netlink.LinkByName(underlayInterface)
+			if err != nil {
+				return fmt.Errorf("failed to get underlay nic by name %s: %w", underlayInterface, err)
+			}
+			// Set group ID only if not already set (idempotent)
+			if underlay.Attrs().Group != underlayGroupID {
+				if err := netlink.LinkSetGroup(underlay, int(underlayGroupID)); err != nil {
+					return fmt.Errorf("failed to set group ID on underlay interface %s: %w", underlayInterface, err)
+				}
+			}
```
qinqon, line 77: The LinkSetGroup is idempotent already, no need for the `if underlay.Attrs().Group != underlayGroupID `  guard.

internal/hostnetwork/underlay.go
```diff
@@ -44,8 +46,39 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
 		}
 	}()
 
-	if params.UnderlayInterface != "" {
-		if err := moveUnderlayInterface(ctx, params.UnderlayInterface, ns); err != nil {
+	// Check if there are existing underlay interfaces that aren't in the new list.
+	// This means the underlay configuration changed and requires a pod restart.
```
qinqon: Nit: The comment should be around reconstructing the named namespace on non recovarable errors, restarting the pod no longer rebuild the netns.

internal/hostnetwork/underlay.go
```diff
@@ -202,76 +184,56 @@ func HasUnderlayInterface(namespace string) (bool, error) {
...
+		for _, l := range links {
+			if l.Attrs().Group == underlayGroupID {
+				mtu := l.Attrs().MTU
+				if minMTU == 0 || mtu < minMTU {
+					minMTU = mtu
+				}
+			}
+		}
+		if minMTU == 0 {
+			slog.Info("no underlay link found when finding MTU")
```
qinqon, line 232: Isn't this wrong when we use the vtepInterface ? like we need to choose the specific interface and return the MTU using minMTU may work but we may end up with smaller MTU than really needed.

Maybe we should prevent configuring multiple nics when vtepInterface is configured ?

qinqon (changes requested):
Nothing big, my main concern is the connection between multiple nics and vtepInterface (aka cloud)

internal/hostnetwork/underlay.go
fedepaol: I don't think this can be a concern. Whoever sets the underlayInterface (and please note that logic was totally removed and is waiting to be reworked), won't move nics around. If we end up with multiple nics, we can't do much better than choosing the smallest mtu available unless we know in advance what path each  traffic type will follow.

What will probably happen with underlay mtu is that we'll end up returning a minMTU = 0 because the group is not set on the underlay interface, and because of that we won't set the mtu on the vxlan interface (which is as wrong as erroring imo).

I would say this will require to be addressed when we reintroduce the logic.

qinqon: @fedepaol and what about not allowing multi nic if vtepInterface is selected ? at the end those other nics are not going to be used at all.

fedepaol: we can't here, because vtepinterface was removed

fedepaol: or, I thought it was :P 
I can do that, but at this point I wonder why we kept that after the resiliency rework. @maiqueb does it make sense to remove it? I think it's misleading at the moment

qinqon: Humm, I think I may be missunderstanding something, what's the relation between the named ns feature and the vtepInterface + nics ? it's the same issue.

fedepaol: vtep interface makes sense with multus. Multus won't work, so I might be wrong but I would've expected vtepInterface to be removed too

qinqon: > vtep interface makes sense with multus. Multus won't work, so I might be wrong but I would've expected vtepInterface to be removed too

I don't think is going to be removed, what we are going to do is implement ipvlan and macvlan as part of the undelay "nics" configuration but we will still need to select the interface as vtep.

Taking into account the api improvements and what I have talk with Miguel, this is the ipvlan example:
```yaml
spec:
   inerfaces: <--- that's the old "nics"
   - type: ipvlan
     interfaceName: net1ipvlan
     master: net1
   tunnelEndpoint:
     interface: net1ipvlan <--- that's the old "vtepInterface"
```

fedepaol: right, so you changed the api and vtep interface doesn't make much sense here imo

qinqon: > right, so you changed the api and vtep interface doesn't make much sense here imo

vtepInterface was introduced to bypass vtepCIDR loopback creation, do you mean that we can totally remove it and just use the first underlay interface if vtep cidr is not configured ? 

How do that will work with multiple interfaces ? we may need to validat that "if there is no vtep cidr you should have just one unerdlay.interfaces"

fedepaol: I am saying that given we can't rely on multus anymore, we'll have a way to know which interface is the right one directly from the api, because that interface will be a consequence of the openperouter api and not of multus. So, the field as it is today does not make much sense imo.

fedepaol: that said, I will introduce validation as a reminder for when we reimplement it

qinqon: > I am saying that given we can't rely on multus anymore, we'll have a way to know which interface is the right one directly from the api, because that interface will be a consequence of the openperouter api and not of multus. So, the field as it is today does not make much sense imo.

Ahh right, also usign it to select from the list of interfaces at underlay is not worh it either I supose, we just need a mechanism to bypass the loopback configuration.

I will reflect that on the enhancement.

e2etests/tests/sessions.go
fedepaol: it's local to the file, so I think either way is fine. Can change if required

internal/conversion/validate_underlay.go
fedepaol: done returning an error

website/content/docs/configuration/_index.md
fedepaol: done

internal/conversion/validate_underlay_test.go
fedepaol: done

internal/conversion/validate_underlay_test.go
fedepaol: done

internal/hostnetwork/underlay.go
maiqueb: I'm confused ... leaving vtep interface behind was yet another thing I didn't notice I was dropping support for when I've implemented #317 .

What is the direction forward ? Drop it from the API in a follow-up, design it in either the API enhancement *or* on #439 , then implement its replacement ?

internal/hostnetwork/underlay.go
maiqueb: this one is tricky ... it may be idempotent it the sense that it doesn't fail ... 

but, I really don't *know* if it won't throw any netlink events that might cause type 2 route withdrawal (we've seen this in the past with setting MAC addresses, which is also idempotent).

Just a heads up ... To be on the safe side, I'd keep it. 

@qinqon if you are sure this'll work, I think we can keep it.

qinqon: @maiqueb let's just keep it if you know similar stuff have cause problems in the past.

internal/hostnetwork/underlay.go
qinqon: My 2 cents is to do a follow up where we drop it and use nic address directly if vtepCIDR is not configured.

fedepaol: imo, it needs to be part of https://github.com/openperouter/openperouter/issues/439 . For now we can either drop it or cry if it's set

maiqueb: I would cry if it is set. I'll push a PR with that. 

@qinqon do you agree ?

internal/hostnetwork/underlay.go
fedepaol: done

internal/hostnetwork/underlay.go
maiqueb: @qinqon went ahead and did it himself. Huge thanks.

https://github.com/openperouter/openperouter/pull/461

internal/conversion/host_conversion_test.go
maiqueb: this was done (erm ... finished ...) in #429

maiqueb (approved):
Thank you. From my perspective this is good. 

You just need to fix the conflict.

PR #310, opened 2026-04-13
Static configuration: apply defaults and validation

fedepaol:

A lot of the logic is embedded in the api definition and applied by the apiserver. This includes defaults and CEL validation rules.

Here we apply the same to statically loaded files.

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

**Release note**:

```release-note
Validate static configuration files with CEL rules and apply defaults.
```


internal/controller/routerconfiguration/static_configuration_reader.go
```diff
@@ -110,5 +165,44 @@ func staticConfigToAPIConfig(staticConfig *static.PERouterConfig) conversion.Api
...
+	// in the map, which CEL rejects. We use k8s.io/apimachinery/pkg/util/json
+	// which preserves integer types (unlike encoding/json which uses float64).
+	jsonData, err := json.Marshal(rawMap)
+	if err != nil {
+		return nil, field.ErrorList{field.InternalError(field.NewPath(""), fmt.Errorf("marshalling to JSON: %w", err))}
 	}
+	var normalizedMap map[string]interface{}
+	if err := k8sjson.Unmarshal(jsonData, &normalizedMap); err != nil {
+		return nil, field.ErrorList{field.InternalError(field.NewPath(""), fmt.Errorf("unmarshalling from JSON: %w", err))}
+	}
```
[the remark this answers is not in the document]
maiqueb: @fedepaol why did you mark this as resolved ? which action did you take ?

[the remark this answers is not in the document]
fedepaol: the comment by gemini is wrong, and it's how the comment in the code says. As per https://pkg.go.dev/k8s.io/apimachinery/pkg/util/json#Unmarshal , `Numbers decoded into interface{} fields are converted to int64 or float64. `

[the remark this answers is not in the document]
maiqueb: @gemini-code-assist check the review above. Seems you could be wrong.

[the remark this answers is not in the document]
fedepaol: no its not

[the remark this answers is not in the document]
fedepaol: uses https://github.com/kubernetes-sigs/json/blob/main/json.go#L61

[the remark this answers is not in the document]
maiqueb: @gemini-code-assist seems you're not right. Evaluate this new info.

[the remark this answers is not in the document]
maiqueb: I give up. I don't think it matters, since I don't think we're using memory.

internal/controller/routerconfiguration/static_configuration_reader.go
```diff
@@ -3,13 +3,29 @@
...
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+var (
+	underlayGVK      = schema.GroupVersionKind{Group: "openpe.openperouter.github.io", Version: "v1alpha1", Kind: "Underlay"}
+	l3vniGVK         = schema.GroupVersionKind{Group: "openpe.openperouter.github.io", Version: "v1alpha1", Kind: "L3VNI"}
+	l2vniGVK         = schema.GroupVersionKind{Group: "openpe.openperouter.github.io", Version: "v1alpha1", Kind: "L2VNI"}
+	l3passthroughGVK = schema.GroupVersionKind{Group: "openpe.openperouter.github.io", Version: "v1alpha1", Kind: "L3Passthrough"}
+	rawFRRConfigGVK  = schema.GroupVersionKind{Group: "openpe.openperouter.github.io", Version: "v1alpha1", Kind: "RawFRRConfig"}
 )
```
qinqon, line 28: Is there a way to remove these ? feels like if we move to v1beta1 or the like we are going to forget, I know we have Ci and all but maybe we can make it more dynamic by reading them from the CRDs.

qinqon, line 28: What about using a recursive walker that fixes only the uint64 stuff ? this way you don't have to use two different json parsers

```go
func normalizeUintToInt(m map[string]interface{}) {
    for k, v := range m {
        switch val := v.(type) {
        case uint64:
            m[k] = int64(val)
        case map[string]interface{}:
            normalizeUintToInt(val)
        case []interface{}:
            for i, elem := range val {
                switch e := elem.(type) {
                case uint64:
                    val[i] = int64(e)
                case map[string]interface{}:
                    normalizeUintToInt(e)
                }
            }
        }
    }
}
```

qinqon (changes requested):
Nothing big, PR looks good an interesting.

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+func TestApplyDefaultsPreservation(t *testing.T) {
+	tests := []struct {
+		name     string
+		gvk      schema.GroupVersionKind
+		obj      *unstructured.Unstructured
+		field    string
+		expected any
+	}{
+		{
+			name:     "Underlay routeridcidr not overridden",
```
maiqueb: isn't this test name wrong ? either I'm missing something, or the test is literally overridding the routeridcidr attribute.

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+			gvk:      underlayGVK,
+			obj:      newUnstructured("Underlay", map[string]any{"asn": int64(65000), "routeridcidr": "192.168.0.0/16"}),
+			field:    "spec.routeridcidr",
+			expected: "192.168.0.0/16",
+		},
+		{
+			name:     "L2VNI vxlanport not overridden",
```
maiqueb: likewise

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+			gvk:      l2vniGVK,
+			obj:      newUnstructured("L2VNI", map[string]any{"vxlanport": int64(5000)}),
+			field:    "spec.vxlanport",
+			expected: int64(5000),
+		},
+		{
+			name:     "L3VNI vxlanport not overridden",
```
maiqueb: ditto

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+			gvk:      l3vniGVK,
+			obj:      newUnstructured("L3VNI", map[string]any{"vrf": "testvrf", "vxlanport": int64(9999)}),
+			field:    "spec.vxlanport",
+			expected: int64(9999),
+		},
+		{
+			name:     "RawFRRConfig priority not overridden",
```
maiqueb: likewise

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+		}
+		if val != int64(4789) {
+			t.Errorf("spec.vxlanport = %v, want %v", val, int64(4789))
+		}
+
+		val, ok = getNestedField(obj, "spec.hostmaster.linuxBridge.autoCreate")
+		if !ok {
+			t.Fatal("spec.hostmaster.linuxBridge.autoCreate not found")
+		}
+		if val != false {
```
maiqueb, line 212: nit: use if val instead ?...

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+		if val != "10.0.0.0/24" {
+			t.Errorf("spec.routeridcidr = %v, want 10.0.0.0/24", val)
+		}
+	})
+
+	t.Run("idempotency", func(t *testing.T) {
+		obj := newUnstructured("L2VNI", map[string]any{})
+
+		err := ApplyDefaults(obj, l2vniGVK)
+		if err != nil {
```
maiqueb: nit: maybe consider inlining these. feel free to ignore.

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+			t.Fatalf("first ApplyDefaults() returned error: %v", err)
+		}
+
+		firstVal, _ := getNestedField(obj, "spec.vxlanport")
+
+		err = ApplyDefaults(obj, l2vniGVK)
+		if err != nil {
```
maiqueb: nit: maybe consider inlining these. feel free to ignore.

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+		if err == nil {
+			t.Fatal("expected error for unknown GVK, got nil")
+		}
+		if !strings.Contains(err.Error(), "no CRD schema found") {
+			t.Errorf("expected error to contain 'no CRD schema found', got: %v", err)
+		}
+	})
+}
+
+func TestValidatePositive(t *testing.T) {
```
maiqueb: nit: it took me to find the next set of tests (the negative counterpart) to understand what this block was about. 

Maybe it's just me, but I think we should rename this to something like: `TestValidateSuccessful`, or `TestSuccessfulValidation` .

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+				t.Errorf("expected no validation errors, got %d:", len(errs))
+				for _, e := range errs {
+					t.Errorf("  - %s", e)
+				}
+			}
+		})
+	}
+}
+
+func TestValidateNegative(t *testing.T) {
```
maiqueb: nit: ditto - **I** find it more expressive if we call this `TestFailingValidation` or something similar. 

maybe it's just me, you can ignore it

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+			t.Error("missing name/autoCreate validation error")
+		}
+		if !foundMismatch {
+			t.Error("missing type/config mismatch validation error")
+		}
+	})
+}
+
+func TestValidateErrorCases(t *testing.T) {
+	t.Run("unknown GVK returns no errors (no validator)", func(t *testing.T) {
```
maiqueb: didn't we test this before ? I remember seeing an unknown GVK test a little bit above, in (line 257):
```
t.Run("unknown GVK returns error", func(t *testing.T) {
```

internal/crdschema/crdschema_test.go
```diff
@@ -0,0 +1,806 @@
...
+		unknownGVK := schema.GroupVersionKind{Group: "unknown.io", Version: "v1", Kind: "Unknown"}
+		obj := newUnstructured("Unknown", map[string]any{})
+
+		errs := Validate(context.Background(), obj, unknownGVK)
+		if len(errs) != 0 {
+			t.Errorf("expected no errors for unknown GVK, got %d", len(errs))
```
maiqueb, line 718: maybe we should get access to / print the errors ?

internal/crdschema/parity_test.go
```diff
@@ -0,0 +1,477 @@
...
+			},
+			validate: func(t *testing.T, u *unstructured.Unstructured) {
+				result := unstructuredToTyped[v1alpha1.Underlay](t, u)
+				if result.Spec.RouterIDCIDR != "10.0.0.0/24" {
+					t.Errorf("expected RouterIDCIDR=10.0.0.0/24, got %q", result.Spec.RouterIDCIDR)
+				}
+			},
+		},
+		{
+			name: "L2VNI minimal with linuxBridge gets VXLanPort and AutoCreate defaults",
```
maiqueb: I understand the test name, but I think we should highlight the fact that "auto create defaults" is just that - don't auto create. The "default" - to me - would be to point by name to an existing bridge.

internal/crdschema/parity_test.go
```diff
@@ -0,0 +1,477 @@
...
+					},
+					L2GatewayIPs: []string{"10.10.10.1/24"},
+				},
+			},
+			validate: func(t *testing.T, u *unstructured.Unstructured) {
+				result := unstructuredToTyped[v1alpha1.L2VNI](t, u)
+				if result.Spec.VXLanPort != 5000 {
+					t.Errorf("expected VXLanPort=5000 (unchanged), got %d", result.Spec.VXLanPort)
+				}
+				if result.Spec.HostMaster.LinuxBridge.Name != "mybridge" {
```
maiqueb: shouldn't we also assert the auto create stayed false ?

internal/crdschema/parity_test.go
```diff
@@ -0,0 +1,477 @@
...
+						VTEPCIDR: "10.100.0.0/24",
+					},
+				},
+			},
+			validate: func(t *testing.T, u *unstructured.Unstructured) {
+				result := unstructuredToTyped[v1alpha1.Underlay](t, u)
+				if result.Spec.RouterIDCIDR != "172.16.0.0/16" {
+					t.Errorf("expected RouterIDCIDR=172.16.0.0/16 (unchanged), got %q", result.Spec.RouterIDCIDR)
+				}
+				if result.Spec.ASN != 65100 {
```
maiqueb: do we default the ASN ? Why are we asserting it ? 

Shouldn't we assert just the attributes that are subject to being defaulted ?

maiqueb: do we default the ASN ? Why are we asserting it ? 

Shouldn't we assert just the attributes that are subject to being defaulted ?

maiqueb (changes requested):
Nice PR. 

I'm overall on board, but I'm having some trouble understanding some of the test cases .

internal/controller/routerconfiguration/static_configuration_reader.go
fedepaol: I will try to add a unit test for that

fedepaol: so I tried, I think the amount of effort required is not worth it. It will include a ton of go parsing code to find the types, I am not sure we want to add here.

fedepaol: if the problem is two parsers, I can stick to one. Apart from that I am not sure this is cleaner / more maintainable

internal/crdschema/crdschema_test.go
fedepaol: nope, what we are testing here is that when we provide the attribute, it doesn't get wiped / overridden with the default. Will change the naming

internal/crdschema/crdschema_test.go
fedepaol: we can't as val is a _any_ , so we can't treat it as boolean

internal/crdschema/crdschema_test.go
fedepaol: will do

internal/crdschema/crdschema_test.go
fedepaol: will rename

internal/crdschema/crdschema_test.go
fedepaol: I'll rename it

internal/crdschema/crdschema_test.go
fedepaol: this is for validate, the other one was for applyDefaults. Will rename

internal/crdschema/crdschema_test.go
fedepaol: will do

internal/crdschema/crdschema_test.go
maiqueb: hm! totally lost sight of that when reading the code. apologies ...

internal/crdschema/crdschema_test.go
maiqueb: where is it  ?

maiqueb (approved):
tks for this

PR #313, opened 2026-04-13
feature: integrate kube-api-linter into lint pipeline

qinqon:

/kind cleanup

Integrates [kube-api-linter](https://github.com/kubernetes-sigs/kube-api-linter) (`sigs.k8s.io/kube-api-linter`) as a golangci-lint module plugin to enforce Kubernetes API conventions on Go type definitions.

### Changes

**Linter integration:**
- `.custom-gcl.yml` — golangci-lint module plugin config (auto-detected by the CI action)
- `.golangci.yml` — `kubeapilinter` enabled, scoped to `api/` paths
- `hack/lint.sh` — builds custom binary via `golangci-lint custom`, then runs it

**API type compliance (breaking changes):**
- Godoc comments start with JSON field name
- Fields marked `+optional`/`+required` as appropriate
- `+listType=atomic` for SSA on list fields
- `+default` marker instead of `+kubebuilder:default`
- `+required` marker instead of `+kubebuilder:validation:Required`
- `uint32`/`uint16`/`int` types replaced with `int32`/`int64`
- `metav1.Duration` replaced with `int64` seconds fields (`holdTimeSeconds`, `keepaliveTimeSeconds`, `connectTimeSeconds`)
- Optional `string`/`bool` fields converted to pointers
- `Status` fields converted to pointers

This is an API-breaking change. All consumer code (conversion, validation, controllers, operator helm, e2e tests) has been updated to handle the new pointer types and renamed fields.

**Release note**:

```release-note
API types updated to comply with Kubernetes API conventions via kube-api-linter. Breaking changes: integer fields use int32/int64 instead of uint, duration fields replaced with seconds integers, optional fields use pointers.
```


hack/lint.sh
```diff
@@ -3,12 +3,17 @@ set -o errexit
 set -x
 
 GOLANGCI_LINT_VERSION="${GOLANGCI_LINT_VERSION:-2.9.0}"
+KUBELINTER_VERSION="${KUBELINTER_VERSION:-0.7.3}"
```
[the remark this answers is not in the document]
qinqon: @gemini-code-assist that's wrong latest is v0.8.3, you are trained with old material.

.kube-linter.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - "privileged-container"
+    - "privilege-escalation-container"
+    - "run-as-non-root"
+    - "no-read-only-root-fs"
+    - "host-network"
+    - "host-pid"
+    - "host-ipc"
+    - "drop-net-raw-capability"
+    - "writable-host-mount"
+    - "sensitive-host-mounts"
```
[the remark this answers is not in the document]
qinqon: Addressed — security checks are now enforced globally, with per-manifest `ignore-check.kube-linter.io` annotations only on the networking pods (controller, router, hostbridge) that legitimately need elevated privileges. The nodemarker security context was also hardened in a separate commit.

hack/lint.sh
```diff
@@ -3,12 +3,17 @@ set -o errexit
...
 
 if [ "$ENV" == "container" ]; then
      docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app golangci/golangci-lint:v$GOLANGCI_LINT_VERSION $CMD
+     docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app stackrox/kube-linter:v$KUBELINTER_VERSION lint $KUBELINTER_DIRS
 else
      curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v"$GOLANGCI_LINT_VERSION"
      $CMD
+     go install golang.stackrox.io/kube-linter/cmd/kube-linter@v$KUBELINTER_VERSION
+     kube-linter lint $KUBELINTER_DIRS
 fi
```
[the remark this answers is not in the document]
qinqon: Good point about surfacing all issues in a single pass. However, I'd prefer to keep `set -o errexit` behavior consistent with how the script worked before this change (golangci-lint already used it). We can improve this in a follow-up if needed.

charts/openperouter/templates/router.yaml
```diff
@@ -100,6 +100,12 @@ data:
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
+  annotations:
+    ignore-check.kube-linter.io/privilege-escalation-container: "FRR requires SYS_ADMIN capability"
```
fedepaol: is there an alternative way to achieve this without having the annotations polluting our yamls?

qinqon: Let me check

qinqon: The only alternative is to disable the check globally, so any manifest can do SYS_ADMIN, there is no per file filtering at .kube-linter.yaml

hack/lint.sh
```diff
@@ -3,12 +3,15 @@ set -o errexit
 set -x
 
 GOLANGCI_LINT_VERSION="${GOLANGCI_LINT_VERSION:-2.9.0}"
```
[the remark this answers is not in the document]
qinqon: That's wrong you got trained on old data

hack/lint.sh
```diff
@@ -3,12 +3,15 @@ set -o errexit
...
 
 if [ "$ENV" == "container" ]; then
-     docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app golangci/golangci-lint:v$GOLANGCI_LINT_VERSION $CMD
+     docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app golangci/golangci-lint:v$GOLANGCI_LINT_VERSION \
+          sh -c "golangci-lint custom && ${CUSTOM_BINARY} ${CMD}"
 else
      curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v"$GOLANGCI_LINT_VERSION"
-     $CMD
+     golangci-lint custom
+     $CUSTOM_BINARY $CMD
```
fedepaol: I think we have the bin folder as an env variable somewhere.
Also, let's duplicate CMD instead of having it behind a variable. It's quite compact anyhow and it will make it cleaerer what we are calling

hack/lint.sh
```diff
@@ -3,12 +3,15 @@ set -o errexit
...
 ENV="${ENV:-container}"
 
 if [ "$ENV" == "container" ]; then
-     docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app golangci/golangci-lint:v$GOLANGCI_LINT_VERSION $CMD
+     docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app golangci/golangci-lint:v$GOLANGCI_LINT_VERSION \
+          sh -c "golangci-lint custom && ${CUSTOM_BINARY} ${CMD}"
 else
      curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v"$GOLANGCI_LINT_VERSION"
-     $CMD
+     golangci-lint custom
```
fedepaol: what does "custom" mean here?

qinqon: I have follow the instructions here https://github.com/kubernetes-sigs/kube-api-linter?tab=readme-ov-file#golangci-lint-module,

Looks like you can run custom plugins to golangci-lint not included there

api/v1alpha1/hostsession.go
fedepaol, line 1: let's split the commit in 3:

- api changes
- regeneration of code
- new api structure alignement


Also, did you update examples / docs? If not, please do. If yes, use separate commits for them too

hack/lint.sh
fedepaol: please comment on what this is doing (I guess it's consuming the yaml below)
In general, this can be easier to understand if we split the "regular" lint execution and the kubeapi one in two separate sections of the file

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,121 @@
...
 // Neighbor represents a BGP Neighbor we want FRR to connect to.
 // +kubebuilder:validation:XValidation:rule="!has(self.hostasn) || self.hostasn != self.asn",message="hostASN must be different from asn for eBGP"
 type Neighbor struct {
-	// ASN is the AS number to use for the local end of the session.
+	// asn is the AS number to use for the local end of the session.
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=4294967295
-	ASN uint32 `json:"asn,omitempty"`
+	// +required
+	ASN int64 `json:"asn,omitempty"`
```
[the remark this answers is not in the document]
qinqon: This is legit and a limitation on the linter, fixing an pushing

operator/internal/helm/chart.go
```diff
@@ -96,10 +96,22 @@ func patchChartValues(envConfig envconfig.EnvConfig, crdConfig *operatorapi.Open
...
+	var multusAnnotation string
+	var runOnMaster bool
+	if crdConfig.Spec != nil {
+		if crdConfig.Spec.MultusNetworkAnnotation != nil {
+			multusAnnotation = *crdConfig.Spec.MultusNetworkAnnotation
+		}
+		if crdConfig.Spec.RunOnMaster != nil {
+			runOnMaster = *crdConfig.Spec.RunOnMaster
+		}
+	}
```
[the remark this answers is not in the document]
qinqon: This make sense too, it's a regression.

hack/lint.sh
qinqon: I will create env for jus timeout, that's the important knob

the LOCALBIN is absolute path so we cannot use it with the container, I will create new RELBIN var at Makefile to use it at the lint.sh

hack/lint.sh
qinqon: This is not how it works like `golangci-lint custom` build a golangci-lint binary including your own plugins then you run the golangci-lint binary, I thikn what is wrong the the binary name, it should just be golangci-lint because it runs all the linters is not just the kube-api-lint

Anyhow I will document it.

api/v1alpha1/neighbor.go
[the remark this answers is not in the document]
qinqon: @gemini-code-assist you are wrong looks like required fields need omitempty, when I removed I get 

```
+ docker run --rm -v /home/ellorent/Documents/cnv/upstream/openperouter:/app -w /app golangci/golangci-lint:v2.9.0 sh -c 'golangci-lint custom && bin/golangci-lint-kube-api-linter run --timeout 10m0s ./...'
api/v1alpha1/neighbor.go:12:2: requiredfields: field Neighbor.ASN should have the omitempty tag. (kubeapilinter)
	ASN int64 `json:"asn"`
	^
api/v1alpha1/neighbor.go:23:2: requiredfields: field Neighbor.Address should have the omitempty tag. (kubeapilinter)
	Address *string `json:"address"`
	^

```

api/v1alpha1/hostsession.go
qinqon: Done, added some docs too.

fedepaol:
sorry I wasn't clear about the documentation part. What I meant was, if the api changes need to be reflected in the project documentation and / or examples, please do so

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,121 @@
...
-	PasswordSecret string `json:"passwordSecret,omitempty"`
+	PasswordSecret *string `json:"passwordSecret,omitempty"`
 
-	// HoldTime is the requested BGP hold time, per RFC4271.
-	// Defaults to 180s.
+	// holdTimeSeconds is the requested BGP hold time in seconds, per RFC4271.
+	// Defaults to 180.
 	// +optional
-	HoldTime *metav1.Duration `json:"holdTime,omitempty"`
+	HoldTimeSeconds *int64 `json:"holdTimeSeconds,omitempty"`
```
fedepaol, line 48: is this really better? This will prevent us specifying things in a more human readable way. What's the rationale?

hack/lint.sh
```diff
@@ -3,12 +3,17 @@ set -o errexit
 set -x
 
 GOLANGCI_LINT_VERSION="${GOLANGCI_LINT_VERSION:-2.9.0}"
-CMD="golangci-lint run --timeout 10m0s ./..."
+CUSTOM_BINARY="golangci-lint"
```
fedepaol: this probably belongs to another commit?

hack/lint.sh
fedepaol: > please comment on what this is doing (I guess it's consuming the yaml below) In general, this can be easier to understand if we split the "regular" lint execution and the kubeapi one in two separate sections of the file

this wasn't addressed I think

api/v1alpha1/neighbor.go
qinqon: Does not make much sense, let me check.

api/v1alpha1/l2vni_types.go
```diff
@@ -27,90 +27,92 @@ const (
...
 	// +optional
 	VRF *string `json:"vrf,omitempty"`
 
-	// VNI is the VXLan VNI to be used
+	// vni is the VXLan VNI to be used
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=4294967295
-	VNI uint32 `json:"vni,omitempty"`
+	// +required
+	VNI *int64 `json:"vni,omitempty"`
```
maiqueb, line 47: if this is required, why is it a pointer ?...

api/v1alpha1/l3passthrough_types.go
```diff
@@ -44,11 +45,17 @@ type L3PassthroughStatus struct {
...
+	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   L3PassthroughSpec   `json:"spec,omitempty"`
-	Status L3PassthroughStatus `json:"status,omitempty"`
+	// spec defines the desired state of L3Passthrough.
+	// +required
+	Spec L3PassthroughSpec `json:"spec,omitzero,omitempty"`
+	// status defines the observed state of L3Passthrough.
+	// +optional
```
maiqueb, line 57: shouldn't this be required ? i.e. we can have an empty status, but, we must have it.

api/v1alpha1/neighbor.go
maiqueb: +1 this is worse

qinqon: Looks like is official kubenetes convention, related to metav1.Duration not being friendly with non golang languages, this is the link to the convention 
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units

> "Duration fields must be represented as integer fields with units being part of the field name (e.g. leaseDurationSeconds). We don't use Duration in the API since that would require clients to implement     
> go-compatible parsing."                

we can go an exclude the "noduration" lint though, so whatever you prefer.

maiqueb: Maybe let's add the `noduration`  for now, create a new issue about that, and work on it as a follow-up ? 

Or we could address it now.

qinqon: You mean excluding "noduration" ? I don't see what issue we should open we either decide on it or not, there is nothing to fix.

api/v1alpha1/l2vni_types.go
qinqon: Because the JSON tag uses `omitempty` and the minimum valid value is 0 (`+kubebuilder:validation:Minimum=0`). If VNI were a non-pointer `int64` with `omitempty`, the zero value would be silently omitted during JSON serialization, making it indistinguishable from "field not set". Using a pointer ensures `nil` means "not provided" (caught by `+required` validation) while `*0` means "explicitly set to 0".

This is the standard Kubernetes API convention that kube-api-linter enforces: required fields with `omitempty` must be pointers when the zero value is within the valid range.

qinqon: Because is an int, this is an extraction from the api convention docs:

https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required
> They (requiered) typically do not use pointer types in the Go definition (e.g. AnotherFlag SomeFlag), though required fields where the zero value is a valid value must use pointer types, paired with an omitempty struct tag to avoid spurious null serializations.

So since 0 value is valid it should be a pointer to propertly define "not set" so validation knows that is not set and is able to fail, instead of doing the interpreation of 0 is the not set value.

api/v1alpha1/l3passthrough_types.go
qinqon: Status is created by controllers not by users so it should be optional to not enforce users to create it.

https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md:
> The PUT and POST verbs on objects MUST ignore the status values

https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md#statusoptional                                                                                                                       
> The statusoptional linter checks that all first-level children fields within a status struct are marked as optional. This is important because status fields should be optional to allow for partial updates   
> and backward compatibility.

api/v1alpha1/l2vni_types.go
maiqueb: Tks for the explanation, that helped. 

But, I do wonder: do we want 0 to be a valid number ?... Do we want to limit the range somehow, like - for instance - [cisco](https://blogs.cisco.com/perspectives/a-summary-of-cisco-vxlan-control-planes-multicast-unicast-mp-bgp-evpn-2) does ? 
> Valid VNI values are from 4096 to 16777215

api/v1alpha1/neighbor.go
maiqueb: no, I meant adding `noduration`, and open an issue to track eventually following up kubernetes guidelines around `time.Duration`.

api/v1alpha1/l2vni_types.go
qinqon: If we can choose I will never represent not set as 0, you don't know what values you are going to use in the future and also is better to have all the int fields be the same so all of them use "nil" for not set.

api/v1alpha1/neighbor.go
qinqon: well since this means chaing the API is not something that is easily to be done at a follow up, we have to make a decision now before v1, this is the purpose of this PR more or less.

maiqueb: ok, then let's follow up the guidelines.

api/v1alpha1/l2vni_types.go
maiqueb: > you don't know what values you are going to use in the future

Well, we do. If we decide the allowed range, we can't go back in time, or break the API. 

Still, I'm OK with this proposal.

qinqon: Ok I will follwo the proposal for this one.

api/v1alpha1/neighbor.go
qinqon: Ok, I will folow the proposal but use metav1.Duration at the code.

qinqon:
CI was failing on commit 1129344. Pushed a fix.

qinqon:
Rebased commit c32dc3d on main and pushed (conflicts resolved).

qinqon:
CI is failing on commit c4fa1ea. Investigated but could not push a fix.

qinqon:
CI was failing on commit 6ee37ea. Pushed a fix.

qinqon:
CI was failing on commit 3afd322. Pushed a fix.

qinqon:
Rebased commit 87079cc on main and pushed.

qinqon:
The `e2etests (operator)` failure is unrelated to this PR — it's a pre-existing CI flake caused by a race condition during pod startup. The controller's `ListPodSandbox` call returns 0 items when the CRI sandbox isn't ready yet, which cascades into FRR reload socket errors and the OVS dual-stack test timing out. All other e2e jobs (manifests, helm, systemdmode, hostmode-boot) pass, including the 80 other tests in the operator job itself.

Tracked in #328.

api/v1alpha1/l3vni_types.go
```diff
@@ -21,41 +21,45 @@ import (
...
 	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
 
-	// VRF is the name of the linux VRF to be used inside the PERouter namespace.
+	// vrf is the name of the linux VRF to be used inside the PERouter namespace.
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9_-]*$`
 	// +kubebuilder:validation:MaxLength=15
-	// +kubebuilder:validation:Required
-	VRF string `json:"vrf"`
+	// +required
+	VRF *string `json:"vrf,omitempty"`
```
maiqueb: the VRF is actually required in the L3VNI. 

It's strange to see this as a pointer.

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,120 @@
...
 type BFDSettings struct {
-	// The minimum interval that this system is capable of
+	// receiveInterval is the minimum interval that this system is capable of
 	// receiving control packets in milliseconds.
 	// Defaults to 300ms.
 	// +kubebuilder:validation:Maximum:=60000
 	// +kubebuilder:validation:Minimum:=10
 	// +optional
-	ReceiveInterval *uint32 `json:"receiveInterval,omitempty"`
+	ReceiveInterval *int32 `json:"receiveInterval,omitempty"`
```
maiqueb, line 81: should this be renamed to something that hints the unit ? now that this is not a duration it might make sense.

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,120 @@
...
 
-	// The minimum transmission interval (less jitter)
+	// transmitInterval is the minimum transmission interval (less jitter)
 	// that this system wants to use to send BFD control packets in
 	// milliseconds. Defaults to 300ms
 	// +kubebuilder:validation:Maximum:=60000
 	// +kubebuilder:validation:Minimum:=10
 	// +optional
-	TransmitInterval *uint32 `json:"transmitInterval,omitempty"`
+	TransmitInterval *int32 `json:"transmitInterval,omitempty"`
```
maiqueb, line 89: ditto; should we add the unit to the attribute name ?

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,120 @@
...
 
-	// Configures the minimal echo receive transmission
+	// echoInterval configures the minimal echo receive transmission
 	// interval that this system is capable of handling in milliseconds.
 	// Defaults to 50ms
 	// +kubebuilder:validation:Maximum:=60000
 	// +kubebuilder:validation:Minimum:=10
 	// +optional
-	EchoInterval *uint32 `json:"echoInterval,omitempty"`
+	EchoInterval *int32 `json:"echoInterval,omitempty"`
```
maiqueb, line 105: ditto; should we add the unit to the attribute name ?

.custom-gcl.yml
```diff
@@ -0,0 +1,6 @@
+version: v2.9.0
+name: golangci-lint-kube-api-linter
+destination: ./bin
+plugins:
+  - module: 'sigs.k8s.io/kube-api-linter'
+    version: 'v0.0.0-20260408163332-73b2175ca510'
```
maiqueb, line 6: what's the criteria for bumping the API linter ?

.golangci.yml
```diff
@@ -35,6 +43,12 @@ linters:
           - dupl
           - lll
         path: internal/*
+      - linters:
+          - kubeapilinter
+        path-except: "api/"
+      - linters:
+          - kubeapilinter
+        path: "api/static/"
```
maiqueb, line 82: can't we squash this into a single entry w/ both `path` and `path-except` ?

internal/conversion/validate_vni.go
```diff
@@ -117,11 +118,12 @@ func ValidateVRFs(l2Vnis []v1alpha1.L2VNI, l3Vnis []v1alpha1.L3VNI) error {
 	vrfToVNI := map[string]types.NamespacedName{}
 	for _, l3Vni := range l3Vnis {
 		namespaceName := types.NamespacedName{Namespace: l3Vni.Namespace, Name: l3Vni.Name}
-		l3vni, ok := vrfToVNI[l3Vni.Spec.VRF]
+		vrfName := ptr.Deref(l3Vni.Spec.VRF, "")
```
maiqueb: hm, we *need* to have a VRF in the l3 VNI. 

shouldn't we error somewhere if not ?

maiqueb (commented):
some comments. the 3rd commit is super hard to review - endless minor changes.

api/v1alpha1/l3vni_types.go
qinqon: > the VRF is actually required in the L3VNI.
> 
> It's strange to see this as a pointer.

That's was my question as the slack channel we have two options here:
- We make them non pointers so we check requiered just at openapi json level (we don't want to do == 0 or == "") then I don't know if this will work for systemd mode or the like
- we keep this as a pointer so apart from the openapi we can do proper is not set check at golang code instead of == 0 or == "" <- it's kind of belt and suspender thing.

api/v1alpha1/neighbor.go
qinqon: This was never a duration -> https://github.com/openperouter/openperouter/blob/1af0c2dd43fbd0f199ea9c540d6504f354357cbc/api/v1alpha1/neighbor.go#L78

api/v1alpha1/neighbor.go
qinqon: ditto; It was never a duration.

api/v1alpha1/neighbor.go
qinqon: ditto; It was never a duration.

.custom-gcl.yml
qinqon: Good question, maybe related to kuberntes version ? we should document this.

.golangci.yml
qinqon: Nope, path and path-except are AND'd in a single rule — combining them would match files in api/static/ that are also NOT in api/, which is nothing.

.custom-gcl.yml
maiqueb: yeah. Where ?

api/v1alpha1/neighbor.go
maiqueb: :facepalm: right. My bad.

qinqon:
> some comments. the 3rd commit is super hard to review - endless minor changes.

@maiqueb that's my main concern, the requiered with pointer thing is generating that, so not sure is worth it although we have comment it a lot.

api/v1alpha1/l3vni_types.go
maiqueb: right, now I remember.

it's strange for sure, but, if this is what the k8s API guidelines say ... Might as well try to follow it. 

@fedepaol your thoughts ?

maiqueb (approved):

maiqueb:
@qinqon the e2e tests behave *really* bad. Definitely looks like there's something up w/ this PR. 

I've approved *conceptually*, but we can't get the PR in in this state.

qinqon:
> @qinqon the e2e tests behave _really_ bad. Definitely looks like there's something up w/ this PR.
> 
> I've approved _conceptually_, but we can't get the PR in in this state.

Yep, looking into it.

maiqueb (approved):
Thank you.

@fedepaol are you OK with this ? 
... let's hold off merging anything until we get this in, OK ?

qinqon:
Rebased commit c2e8982 on main and pushed.

api/v1alpha1/l3vni_types.go
fedepaol: cel validation will happen in systemd mode too.
Given it's not optional I'd keep it non pointer unless the guidelines say otherwise (not sure I understood it well)

fedepaol:
> Thank you.
> 
> @fedepaol are you OK with this ? ... let's hold off merging anything until we get this in, OK ?

works for me. Going over the PR right now

api/v1alpha1/duration.go
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package v1alpha1
+
+import "time"
+
+func durationFromPtr[T ~int32 | ~int64](v *T, unit time.Duration) *time.Duration {
```
fedepaol: util functions to the bottom please

api/v1alpha1/duration.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (n *Neighbor) ConnectTime() *time.Duration {
+	return durationFromPtr(n.ConnectTimeSeconds, time.Second)
+}
+
+func (n *Neighbor) SetConnectTime(d *time.Duration) {
+	n.ConnectTimeSeconds = ptrFromDuration[int64](d, time.Second)
+}
+
+// BFD getters use Get prefix to avoid conflict with same-named fields.
```
fedepaol: unnecessary comment imo. Maybe it makes sense to name in a way that is clear that we are converting from / to duration

api/v1alpha1/l3vni_types.go
```diff
@@ -21,41 +21,45 @@ import (
...
 	// +optional
 	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
 
-	// VRF is the name of the linux VRF to be used inside the PERouter namespace.
+	// vrf is the name of the linux VRF to be used inside the PERouter namespace.
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9_-]*$`
 	// +kubebuilder:validation:MaxLength=15
-	// +kubebuilder:validation:Required
-	VRF string `json:"vrf"`
+	// +required
```
fedepaol, line 36: what's the difference between validation:required and required here?

Makefile
```diff
@@ -15,6 +15,8 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+RELBIN=bin
```
fedepaol: what's the point of this

api/v1alpha1/duration.go
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier:Apache-2.0
```
fedepaol: this commit says it's about the fixes but contains also the integration

e2etests/pkg/frrk8s/config.go
```diff
@@ -25,21 +26,24 @@ var (
 // ConfigFromHostSession converts a HostSession object to FRRConfiguration objects.
 // Returns a slice with one configuration for each IP family (IPv4 and/or IPv6).
 func ConfigFromHostSession(hostsession v1alpha1.HostSession, name string, tweak ...func(*frrk8sapi.FRRConfiguration)) ([]frrk8sapi.FRRConfiguration, error) {
-	if hostsession.LocalCIDR.IPv4 == "" && hostsession.LocalCIDR.IPv6 == "" {
+	ipv4 := ptr.Deref(hostsession.LocalCIDR.IPv4, "")
```
fedepaol, line 29: shouldn't we just check for nil here instead of "" here?

fedepaol: same for all the other places where we deref. I think we can assume that if the pointer is set, the value is not nil?

e2etests/systemd_static_suite/validate_routes.go
```diff
@@ -31,10 +32,10 @@ func checkRouteFromLeaf(leaf infra.Leaf, routers openperouter.Routers, vni v1alp
 				return fmt.Errorf("failed to get EVPN info from %s: %w", exec.Name(), err)
 			}
 			for _, prefix := range prefixes {
-				if check == mustContain && !evpn.ContainsType5RouteForVNI(prefix, leaf.VTEPIP, int(vni.Spec.VNI)) {
+				if check == mustContain && !evpn.ContainsType5RouteForVNI(prefix, leaf.VTEPIP, int(ptr.Deref(vni.Spec.VNI, 0))) {
```
fedepaol: if we ever provide a nil pointer this will explode. It's test code, and always driven by the test fixtures.
How bad it is to just dereference the pointer with the * instead of wrapping? Imo it's borderline to better as it's gonna be much easier to spot the error

fedepaol: same for all the other places where we call ptr.Deref

internal/conversion/validate_vni.go
fedepaol: agree

internal/conversion/frr_conversion.go
```diff
@@ -344,46 +344,34 @@ func parseTimers(ht, ka *metav1.Duration) (*uint64, *uint64, error) {
...
-}
-
 func routerIDFromUnderlay(underlay v1alpha1.Underlay, nodeIndex int) (string, error) {
-	routerIDCidr := underlay.Spec.RouterIDCIDR
-	if underlay.Spec.RouterIDCIDR == "" {
-		routerIDCidr = defaultRouterIDCidr
+	routerIDCidr := defaultRouterIDCidr
+	if underlay.Spec.RouterIDCIDR != nil && *underlay.Spec.RouterIDCIDR != "" {
+		routerIDCidr = *underlay.Spec.RouterIDCIDR
+	} else {
```
fedepaol: can we revert to be positive? Or flip the log and embed it in hte if without the else?

internal/conversion/frr_conversion.go
```diff
@@ -344,46 +344,34 @@ func parseTimers(ht, ka *metav1.Duration) (*uint64, *uint64, error) {
...
 	}
 	routerID, err := ipam.RouterID(routerIDCidr, nodeIndex)
 	if err != nil {
-		return "", fmt.Errorf("failed to get router id, cidr %s, nodeIndex %d: %w", underlay.Spec.RouterIDCIDR, nodeIndex, err)
+		return "", fmt.Errorf("failed to get router id, cidr %s, nodeIndex %d: %w", routerIDCidr, nodeIndex, err)
 	}
 	return routerID, nil
 }
+
+func localCIDRStrings(lc v1alpha1.LocalCIDRConfig) (string, string) {
```
fedepaol: the name localCIDRStrings is not descriptive imo. 

Let's have a CIDRForFamily(lc, ipfamily) function.

internal/conversion/host_conversion.go
```diff
@@ -41,7 +42,8 @@ func APItoHostConfig(nodeIndex int, targetNS string, underlayFromMultus bool, ap
 	}
 
 	if len(apiConfig.L3Passthrough) == 1 {
-		vethIPs, err := ipam.VethIPsFromPool(apiConfig.L3Passthrough[0].Spec.HostSession.LocalCIDR.IPv4, apiConfig.L3Passthrough[0].Spec.HostSession.LocalCIDR.IPv6, nodeIndex)
+		ipv4CIDR, ipv6CIDR := localCIDRStrings(apiConfig.L3Passthrough[0].Spec.HostSession.LocalCIDR)
+		vethIPs, err := ipam.VethIPsFromPool(ipv4CIDR, ipv6CIDR, nodeIndex)
```
fedepaol: can't we change vethipsfrompool to take the nil pointer into account instead of dereferencing from outside?

internal/conversion/host_conversion.go
```diff
@@ -68,31 +70,32 @@ func APItoHostConfig(nodeIndex int, targetNS string, underlayFromMultus bool, ap
...
+			return res, fmt.Errorf("failed to get vtep ip, cidr %s, nodeIndex %d: %w", *underlay.Spec.EVPN.VTEPCIDR, nodeIndex, err)
 		}
 		res.Underlay.EVPN.VtepIP = vtepIP.String()
 	}
 
 	for _, vni := range apiConfig.L3VNIs {
 		v := hostnetwork.L3VNIParams{
 			VNIParams: hostnetwork.VNIParams{
-				VRF:           vni.Spec.VRF,
+				VRF:           ptr.Deref(vni.Spec.VRF, ""),
```
fedepaol: can we make the internal fields pointers instead of dereferencing from outside?

internal/conversion/host_conversion.go
```diff
@@ -68,31 +70,32 @@ func APItoHostConfig(nodeIndex int, targetNS string, underlayFromMultus bool, ap
...
 				TargetNS:      targetNS,
 				VTEPIP:        res.Underlay.EVPN.VtepIP,
-				VTEPInterface: underlay.Spec.EVPN.VTEPInterface,
-				VNI:           int(vni.Spec.VNI),
-				VXLanPort:     int(vni.Spec.VXLanPort),
+				VTEPInterface: ptr.Deref(underlay.Spec.EVPN.VTEPInterface, ""),
+				VNI:           int(ptr.Deref(vni.Spec.VNI, 0)),
+				VXLanPort:     int(ptr.Deref(vni.Spec.VXLanPort, 4789)),
```
fedepaol: the default here is super arbitrary

internal/conversion/frr_conversion.go
```diff
@@ -257,37 +259,35 @@ func l3vniToFRR(vni v1alpha1.L3VNI, routerID string, underlayASN uint32, nodeInd
 }
 
 func neighborToFRR(n v1alpha1.Neighbor) (*frr.NeighborConfig, error) {
-	asn, err := frr.NewPeerASN(n.ASN, n.Type)
+	addr := ptr.Deref(n.Address, "")
+	asn, err := frr.NewPeerASN(ptr.Deref(n.ASN, 0), ptr.Deref(n.Type, ""))
```
fedepaol: same here, can we just err if we get nil pointers?

fedepaol (commented):
A general comment. We are trying to be on the safe side by wrapping every access with ptr.Deref.

Imo we can aim to a more readable approach where inner types have the corresponding fields as pointers, and functions handle the pointers and err (where appropriate)

e2etests/pkg/frrk8s/config.go
fedepaol: ignore this comment, doesnt make sense

website/content/docs/contributing/_index.md
```diff
@@ -99,6 +99,17 @@ Check [the dev environment documentation]({{< relref "devenv.md" >}}) for more d
 After deploying the ContainerLab environment together with Kind, the kubeconfig will
 be available at `bin/kubeconfig`.
 
+### Running the linter
```
fedepaol: Imo this is not necessary. What I asked for changes in the docs where needed, because I thought some examples might change (ie for the duration).
Please double check that

api/v1alpha1/l3vni_types.go
qinqon: the `+requiered` is the generic version, the other one is kubebuilder specific.

Makefile
qinqon: To specify the name of the dir where we store binaries, independently of the path.

api/v1alpha1/duration.go
qinqon: I think these helpers are part of the fixes for API more than the integration, the integration code is the stuff about using them.

api/v1alpha1/l3vni_types.go
qinqon: This is the same thing, if we want to check if value is passed or not at controller we need a pointer to propertly check that instead of using == ""

but maybe for VRFs the empty string is not going to be never a valid value (Or is it ? in case users want to remove the VRF or the like?)

hack/lint.sh
qinqon: I will move the linter tool at different commit

e2etests/systemd_static_suite/validate_routes.go
qinqon: If we pass nil pointer it will return `0` it will not explode, using the normal `*` make golang to panic

qinqon:
CI check `static-security-analysis` was failing on commit fba82f0. Pushed a fix.

qinqon:
CI check `unit-tests` was failing on commit d84f605. Pushed a fix.

internal/conversion/validate_vni.go
qinqon: We are not going to have an entry for "" at the vrfToVNI map, so it will fail like in the current code (where we don't check empty VRF either)

internal/conversion/host_conversion.go
qinqon: This is the default at IANA and also marked as default in the field.

qinqon:
> A general comment. We are trying to be on the safe side by wrapping every access with ptr.Deref.
> 
> Imo we can aim to a more readable approach where inner types have the corresponding fields as pointers, and functions handle the pointers and err (where appropriate)

@fedepaol  Agree with that I will refactor the PR with it.

internal/conversion/host_conversion.go
qinqon: Yes, now we don't need the localCIDRStrings function at all.

internal/conversion/host_conversion.go
qinqon: We have chagne the intermediate struct to use pointer we no longer need this here though.

qinqon:
CI failure is unrelated looks like router pods has lost ther 2001:db8:20::/64 routing so it cannot access so they cannot access VRF red.

.custom-gcl.yml
qinqon: I was wrong is not related to kubernetes versions, we just bump whenever we want same as golangci-lint

website/content/docs/contributing/_index.md
qinqon: I will remove this commit, I missunderstood the comment. the website, configs and examples are already updated at the other commits (most of it at the regeneration one).

hack/lint.sh
```diff
@@ -3,12 +3,17 @@ set -o errexit
...
+CUSTOM_BINARY_PATH=${RELBIN}/${CUSTOM_BINARY}
+TIMEOUT="10m0s"
+CMD="run --timeout 10m0s ./..."
```
RamLavi: `CMD` is unused now. can you remove it?

hack/lint.sh
```diff
@@ -3,12 +3,17 @@ set -o errexit
 set -x
 
 GOLANGCI_LINT_VERSION="${GOLANGCI_LINT_VERSION:-2.9.0}"
-CMD="golangci-lint run --timeout 10m0s ./..."
+CUSTOM_BINARY="golangci-lint"
+CUSTOM_BINARY_PATH=${RELBIN}/${CUSTOM_BINARY}
```
RamLavi: nit: consider adding a default value for `RELBIN`, so that if someone runs this directly it won't be an empty string.
 
```
RELBIN="${RELBIN:-bin}
```

internal/conversion/frr_conversion_test.go
```diff
@@ -594,139 +594,6 @@ func TestAPItoFRR(t *testing.T) {
 			},
 			wantErr: false,
 		},
-		{
```
RamLavi, line 597: Can you explain why are these unit tests deleted? From first glance these seem valid functionalities.
If for valid reasons - can you update on the commit?

RamLavi (changes requested):
most concerning issue is the deleted unit tests

internal/conversion/frr_conversion_test.go
qinqon: Good catch! let me investigate

RamLavi (approved):
LGTM

.custom-gcl.yml
ormergi, line 6: I think we can specify the binary here instead of the liner command, it should simplify the scripts
see https://golangci-lint.run/docs/plugins/module-plugins/
Do you think it can work?

api/v1alpha1/l2vni_types.go
```diff
@@ -27,90 +27,92 @@ const (
...
-// +kubebuilder:validation:XValidation:rule="(self.?name.orValue(\"\") == \"\") == self.autoCreate",message="either name must be set or autoCreate must be true, but not both."
+// +kubebuilder:validation:XValidation:rule="(self.?name.orValue(\"\") == \"\") == self.?autoCreate.orValue(false)",message="either name must be set or autoCreate must be true, but not both."
 type LinuxBridgeConfig struct {
-	// Name of the Linux bridge interface.
+	// name of the Linux bridge interface.
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9_-]*$`
 	// +kubebuilder:validation:MaxLength=15
 	// +optional
-	Name string `json:"name,omitempty"`
+	Name *string `json:"name,omitempty"`
```
ormergi, line 79: nit: Shouldn't name be required, when autoCreate=false?

api/v1alpha1/l2vni_types.go
```diff
@@ -27,90 +27,92 @@ const (
...
-// +kubebuilder:validation:XValidation:rule="(self.?name.orValue(\"\") == \"\") == self.autoCreate",message="either name must be set or autoCreate must be true, but not both."
+// +kubebuilder:validation:XValidation:rule="(self.?name.orValue(\"\") == \"\") == self.?autoCreate.orValue(false)",message="either name must be set or autoCreate must be true, but not both."
 type OVSBridgeConfig struct {
-	// Name of the OVS bridge interface.
+	// name of the OVS bridge interface.
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9_-]*$`
 	// +kubebuilder:validation:MaxLength=15
 	// +optional
-	Name string `json:"name,omitempty"`
+	Name *string `json:"name,omitempty"`
```
ormergi, line 95: nit: same here - shouldn't name should be required, when autoCreate=false?

api/v1alpha1/l3vni_types.go
```diff
@@ -21,41 +21,46 @@ import (
...
 	// +kubebuilder:validation:Maximum=4294967295
 	// +optional
-	VNI uint32 `json:"vni,omitempty"`
+	VNI *int64 `json:"vni,omitempty"`
 
-	// VXLanPort is the port to be used for VXLan encapsulation.
-	// +kubebuilder:default:=4789
-	VXLanPort uint32 `json:"vxlanport,omitempty"`
+	// vxlanport is the port to be used for VXLan encapsulation.
+	// +default=4789
```
ormergi, line 46: nit: We should avoid aetting defaults via CRD markers it makes it harder to change. Having defaults set programmatically allow easier transition to different defaults.

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,121 @@
...
 	PassiveMode *bool `json:"passiveMode,omitempty"`
 
-	// For multi hop sessions only: configure the minimum
+	// minimumTtl configures, for multi hop sessions only, the minimum
 	// expected TTL for an incoming BFD control packet.
 	// +kubebuilder:validation:Maximum:=254
 	// +kubebuilder:validation:Minimum:=1
 	// +optional
-	MinimumTTL *uint32 `json:"minimumTtl,omitempty"`
+	MinimumTTL *int32 `json:"minimumTtl,omitempty"`
```
ormergi: nit: Shouldnt JSON tags use capital letters for acronyms? (TTL vs Ttl)

Makefile
fedepaol: yes but are we actually using it?

api/v1alpha1/underlay_types.go
```diff
@@ -22,46 +22,52 @@ import (
...
 // EVPNConfig contains EVPN-VXLAN configuration for the underlay.
 // +kubebuilder:validation:XValidation:rule="(self.?vtepcidr.orValue(\"\") != \"\") != (self.?vtepInterface.orValue(\"\") != \"\")",message="exactly one of vtepcidr or vtepInterface must be specified"
 type EVPNConfig struct {
-	// VTEPCIDR is CIDR to be used to assign IPs to the local VTEP on each node.
+	// vtepcidr is the CIDR to be used to assign IPs to the local VTEP on each node.
 	// A loopback interface will be created with an IP derived from this CIDR.
 	// Mutually exclusive with vtepInterface.
 	// +optional
-	VTEPCIDR string `json:"vtepcidr,omitempty"`
+	VTEPCIDR *string `json:"vtepcidr,omitempty"`
```
ormergi: nit: same for the JSON tag - vtepcider -> vtepCIDR

hack/lint.sh
```diff
@@ -3,12 +3,17 @@ set -o errexit
 set -x
 
 GOLANGCI_LINT_VERSION="${GOLANGCI_LINT_VERSION:-2.9.0}"
-CMD="golangci-lint run --timeout 10m0s ./..."
+CUSTOM_BINARY="golangci-lint"
+RELBIN=${RELBIN:-bin}
+CUSTOM_BINARY_PATH=${RELBIN}/${CUSTOM_BINARY}
```
fedepaol: I find all this nested definition confusing
CUSTOM_BINARY is golangci-lint but we also pass it below as parameter to itself?

Can we untangle by giving meaningful names and maybe hardcode when possible?
Imo what makes sense is a global BIN folder where binaries are, and all the rest hardcoded.

hack/lint.sh
```diff
@@ -3,12 +3,17 @@ set -o errexit
...
 ENV="${ENV:-container}"
 
 if [ "$ENV" == "container" ]; then
-     docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app golangci/golangci-lint:v$GOLANGCI_LINT_VERSION $CMD
+     docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app golangci/golangci-lint:v$GOLANGCI_LINT_VERSION \
+          sh -c "golangci-lint custom --name ${CUSTOM_BINARY} && ${CUSTOM_BINARY_PATH} run --timeout $TIMEOUT ./..."
 else
      curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v"$GOLANGCI_LINT_VERSION"
-     $CMD
+     golangci-lint custom --name ${CUSTOM_BINARY}
```
fedepaol: having a build_custom_golangci.sh would probably help in making this simpler. We can consume it from the container too I think

api/v1alpha1/duration.go
fedepaol: I don't agree but I won't block on this.
But if you say "fix the linter errors", the commit should be laser focused in fixing them.
The helpers are about how the rest of the codebase interact with the changes.

api/v1alpha1/l2vni_types.go
```diff
@@ -128,17 +130,23 @@ type L2VNIStatus struct {
...
+	metav1.TypeMeta `json:",inline"`
+	// metadata is the standard object metadata.
+	// +optional
 	metav1.ObjectMeta `json:"metadata"`
 
-	Spec   L2VNISpec   `json:"spec,omitempty"`
-	Status L2VNIStatus `json:"status,omitempty"`
+	// spec defines the desired state of L2VNI.
+	// +optional
+	Spec *L2VNISpec `json:"spec,omitempty"`
```
fedepaol: spec is optional?

api/v1alpha1/l3vni_types.go
fedepaol: thanks

e2etests/systemd_static_suite/validate_routes.go
fedepaol: yes, what I am saying is that given Spec.VNI is coming from the test fixture, it should never be nil. This, because we know we are producing the value).

Dereferencing it and protecting against nil will make finding the error harder imo

e2etests/tests/hostconfiguration.go
```diff
@@ -642,7 +642,7 @@ var _ = ginkgo.Describe("Router Host configuration", func() {
 
 				ginkgo.By(fmt.Sprintf("validating Underlay for pod %s", p.Name))
 
-				vtepIP := vtepIPForPod(cs, underlay.Spec.EVPN.VTEPCIDR, p)
+				vtepIP := vtepIPForPod(cs, ptr.Deref(underlay.Spec.EVPN.VTEPCIDR, ""), p)
```
fedepaol: if we cahnge vtepIPForPod to accept a pointer we can avoid all these derefs

e2etests/tests/hostconfiguration.go
```diff
@@ -1311,8 +1311,8 @@ var _ = ginkgo.Describe("Router Host configuration", func() {
 
 			l3PassthroughParams := l3passthroughParams{
 				HostVeth: &veth{
-					NSIPv4: routerIPWithNetmask(passthroughWithNodeSelector.Spec.HostSession.LocalCIDR.IPv4),
-					NSIPv6: routerIPWithNetmask(passthroughWithNodeSelector.Spec.HostSession.LocalCIDR.IPv6),
+					NSIPv4: routerIPWithNetmask(ptr.Deref(passthroughWithNodeSelector.Spec.HostSession.LocalCIDR.IPv4, "")),
```
fedepaol: same, we can change routerIPWithNetmask to accept a pointer

e2etests/tests/validate_session.go
```diff
@@ -14,19 +14,20 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
 )
 
 const Established = true
 
 func validateFRRK8sSessionForHostSession(name string, hostsession v1alpha1.HostSession, established bool, frrk8sPods ...*corev1.Pod) {
 	var cidrs []string
-	Expect(hostsession.LocalCIDR.IPv4 != "" || hostsession.LocalCIDR.IPv6 != "").To(BeTrue(), "either IPv4 or IPv6 CIDR must be provided")
+	Expect(ptr.Deref(hostsession.LocalCIDR.IPv4, "") != "" || ptr.Deref(hostsession.LocalCIDR.IPv6, "") != "").To(BeTrue(), "either IPv4 or IPv6 CIDR must be provided")
```
fedepaol: let's assign these to variables (I think we did it somewhere above) as we are doing logic on those values

internal/conversion/frr_conversion.go
```diff
@@ -257,37 +256,35 @@ func l3vniToFRR(vni v1alpha1.L3VNI, routerID string, underlayASN uint32, nodeInd
 }
 
 func neighborToFRR(n v1alpha1.Neighbor) (*frr.NeighborConfig, error) {
+	addr := n.Address
```
fedepaol: this is not needed I thikn?

internal/conversion/frr_conversion.go
```diff
@@ -257,37 +256,35 @@ func l3vniToFRR(vni v1alpha1.L3VNI, routerID string, underlayASN uint32, nodeInd
...
-		Addr:         n.Address,
+		Addr:         addr,
 		Port:         n.Port,
 		IPFamily:     neighborFamily,
-		EBGPMultiHop: n.EBGPMultiHop,
+		EBGPMultiHop: ptr.Deref(n.EBGPMultiHop, false),
+		Password:     ptr.Deref(n.Password, ""),
 	}
-	res.HoldTime, res.KeepaliveTime, err = parseTimers(n.HoldTime, n.KeepaliveTime)
+	res.HoldTime, res.KeepaliveTime, err = parseTimers(n.HoldTime(), n.KeepaliveTime())
```
fedepaol: here we are calling HoldTime() which converts HoldTimeSeconds to duration... just to call parseTimers which converts it back to seconds.

internal/conversion/frr_conversion.go
```diff
@@ -257,37 +256,35 @@ func l3vniToFRR(vni v1alpha1.L3VNI, routerID string, underlayASN uint32, nodeInd
...
 	}
 
-	if n.ConnectTime != nil {
-		connectSecond, err := durationToUint64(n.ConnectTime.Duration / time.Second)
-		if err != nil {
-			return nil, fmt.Errorf("neighbor %s: invalid connecttime %v: %w", neighName, n.ConnectTime.Duration, err)
-		}
-		res.ConnectTime = ptr.To(connectSecond)
+	if ct := n.ConnectTime(); ct != nil {
+		res.ConnectTime = ptr.To(uint64(ct.Seconds()))
```
fedepaol: same here

internal/conversion/validate_hostsession.go
```diff
@@ -48,19 +49,21 @@ func ValidateHostSessions(l3VNIs []v1alpha1.L3VNI, l3Passthrough []v1alpha1.L3Pa
 	existingCIDRsV4 := map[string]string{}
 	existingCIDRsV6 := map[string]string{}
 	for _, s := range hostSessions {
-		if s.LocalCIDR.IPv4 != "" {
-			if err := validateCIDR(s, s.LocalCIDR.IPv4, existingCIDRsV4); err != nil {
+		ipv4CIDR := ptr.Deref(s.LocalCIDR.IPv4, "")
```
fedepaol: let's call this localIPV4CIDR (same for ipv6 below)

internal/conversion/validate_vni.go
```diff
@@ -120,11 +121,12 @@ func ValidateVRFs(l2Vnis []v1alpha1.L2VNI, l3Vnis []v1alpha1.L3VNI) error {
 	vrfToVNI := map[string]types.NamespacedName{}
 	for _, l3Vni := range l3Vnis {
 		namespaceName := types.NamespacedName{Namespace: l3Vni.Namespace, Name: l3Vni.Name}
-		l3vni, ok := vrfToVNI[l3Vni.Spec.VRF]
+		vrfName := l3Vni.Spec.VRF
```
fedepaol: this is not necessary I think

api/v1alpha1/l2vni_types.go
fedepaol: Sorry getting back on this, I just think we shouldn't allow 0 and make vni required and non pointer.

fedepaol: same for l3vni

internal/hostnetwork/vxlan.go
```diff
@@ -46,8 +47,9 @@ func checkVXLanConfigured(vxLan *netlink.Vxlan, bridgeIndex, loopbackIndex int,
 		return fmt.Errorf("vxlanid is not vni: %d, %d", vxLan.VxlanId, params.VNI)
 	}
 
-	if vxLan.Port != params.VXLanPort {
-		return fmt.Errorf("port is not one coming from params: %d, %d", vxLan.Port, params.VXLanPort)
+	vxlanPort := int(ptr.Deref(params.VXLanPort, 4789))
```
fedepaol: paramsVXLanPort

internal/hostnetwork/vxlan.go
```diff
@@ -176,17 +178,18 @@ func validateVxlan(vxLan *netlink.Vxlan, params VNIParams) error {
 		return nil
 	}
 
-	if params.VTEPInterface == "" {
+	vtepIfaceName := ptr.Deref(params.VTEPInterface, "")
```
fedepaol: paramsVTEPInterfaceName

hack/lint.sh
qinqon: > having a build_custom_golangci.sh would probably help in making this simpler. We can consume it from the container too I think

Moved the script to ./hack/golangci-lint.sh so now we have "./hack/golangci-lint.sh build" and "./hack/golagnci-lint.sh run" that we can call from Makefile

api/v1alpha1/l2vni_types.go
ormergi: I couldn't find ref for this but from what I know the common convention is having both Spec and Status non pointers.

hack/lint.sh
qinqon: > I find all this nested definition confusing
> CUSTOM_BINARY is golangci-lint but we also pass it below as parameter to itself?
> 
> Can we untangle by giving meaningful names and maybe hardcode when possible?
> Imo what makes sense is a global BIN folder where binaries are, and all the rest hardcoded.

I have simplify the script and now it has "build" and "run" functions and also an internal run to wrap it with container if needed

api/v1alpha1/hostsession.go
```diff
@@ -7,45 +7,46 @@ package v1alpha1
...
 // +kubebuilder:validation:XValidation:rule="!has(self.hostasn) || !has(self.hosttype)",message="HostASN and HostType cannot be set together"
 type HostSession struct {
-	// ASN is the local AS number to use to establish a BGP session with
+	// asn is the local AS number to use to establish a BGP session with
 	// the default namespace.
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=4294967295
 	// +required
-	ASN uint32 `json:"asn,omitempty"`
+	ASN int64 `json:"asn,omitempty"`
```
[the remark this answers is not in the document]
ormergi: Shouldn't required fields have omitempty?

api/v1alpha1/l2vni_types.go
qinqon: > spec is optional?

Right, fixing

e2etests/tests/validate_session.go
qinqon: > let's assign these to variables (I think we did it somewhere above) as we are doing logic on those values

Adding directly them to "cidrs" slice and checking the slice

internal/conversion/frr_conversion.go
qinqon: > here we are calling HoldTime() which converts HoldTimeSeconds to duration... just to call parseTimers which converts it back to seconds.

Feels redundant but let's keep logic using Duration, if in the future we change the api from seconds to something else we will not need to change the logic just the API and their conversors

internal/conversion/frr_conversion.go
qinqon: > same here

ditto

internal/conversion/frr_conversion.go
fedepaol: I wouldn't bother with future. Also, the api is getting stable so I don't foresee it (unless we want to get back to duration)

fedepaol: @maiqueb wdyt?

api/v1alpha1/duration.go
qinqon: ok moving the helpers change to the integration commit.

internal/conversion/frr_conversion.go
maiqueb: wait, so you're saying that we could directly do something like below ? @fedepaol 

```golang
res.HoldTime = n.HoldTime()
res.KeepaliveTime = n.KeepaliveTime()
```

e2etests/systemd_static_suite/validate_routes.go
qinqon: Ok using * instead

internal/conversion/frr_conversion.go
qinqon: @maiqueb this is not the case the `parseTimers` is doing extra stuff.

fedepaol: > wait, so you're saying that we could directly do something like below ? @fedepaol
> 
> ```go
> res.HoldTime = n.HoldTime()
> res.KeepaliveTime = n.KeepaliveTime()
> ```

validation and extra stuff aside, we woudln't even need n.HoldTime() but we could use n.HoldTime as both the receiver and the value are of the same type

maiqueb: my bad, I meant the attribute (which is in seconds), not the method that returns a duration from the attribute. 

From what I understand, `parseTimers` is simply doing some validation ... 

Unless it's impossible, I'd move this validation to the API (CEL), and directly assign the values to the result, as @fedepaol seems to suggest. 

@qinqon in case I'm missing something, can you elaborate what "extra stuff" ? 
> the parseTimers is doing extra stuff

qinqon:
Note that I have move VNI's from uint32 to int32 since we have enough room at an int32 to the max VNI value 16.777.215

api/v1alpha1/l2vni_types.go
qinqon: Make sense adding CEL to it, although I think we are already validating it beyond the types.

api/v1alpha1/l2vni_types.go
qinqon: fixed

api/v1alpha1/l3vni_types.go
qinqon: This default value was already configured ont he base, I am not going to change that since it's not related to kube api linter conventions

api/v1alpha1/neighbor.go
qinqon: There is no kube api linter or kubernetes api conventions for that, it's kind of golang naming convention, so it do not apply (although I like it better).

qinqon: Changing it, now is the time to do it.

api/v1alpha1/underlay_types.go
qinqon: This do not follow the conventions since it should be either `vtepCidr` or `vtepCIDR` 

https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

Changint it to `vtepCIDR`

hack/golangci-lint.sh
```diff
@@ -0,0 +1,27 @@
...
+	     docker run --rm -v $(git rev-parse --show-toplevel):/app -w /app golangci/golangci-lint:v$GOLANGCI_LINT_VERSION $@
+	else
+	   curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v"$GOLANGCI_LINT_VERSION"
+	   $@
+	fi
+
+}
+
+function build() {
+	_run golangci-lint custom --name golangci-lint
```
fedepaol: nit: it probably makes sense to use a different name for the binary?

operator/api/v1alpha1/openperouter_types.go
```diff
@@ -68,11 +68,17 @@ type OpenPERouterStatus struct{}
...
+	metav1.TypeMeta `json:",inline"`
+	// metadata is the standard object metadata.
+	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   OpenPERouterSpec   `json:"spec,omitempty"`
-	Status OpenPERouterStatus `json:"status,omitempty"`
+	// spec defines the desired state of OpenPERouter.
+	// +optional
+	Spec *OpenPERouterSpec `json:"spec,omitempty"`
```
fedepaol: spec became optional here

api/v1alpha1/duration.go
```diff
@@ -0,0 +1,69 @@
...
+
+package v1alpha1
+
+import "time"
+
+func (n *Neighbor) HoldTime() *time.Duration {
```
fedepaol: let's remove these if we go on the directoin of removing the double conversion. Also, are the setters used anywhere?

internal/conversion/frr_conversion.go
qinqon: The functions is also doing some validations, I will move them to CEL 
```golang
func parseTimers(ht, ka *time.Duration) (*uint64, *uint64, error) {
	if ht == nil && ka != nil || ht != nil && ka == nil {
		return nil, nil, fmt.Errorf("one of KeepaliveTime/HoldTime specified, both must be set or none")
	}

	if ht == nil && ka == nil {
		return nil, nil, nil
	}

	if *ht != 0 && *ht < 3*time.Second {
		return nil, nil, fmt.Errorf("invalid hold time %q: must be 0 or >=3s", ht)
	}

	if *ka > *ht {
		return nil, nil, fmt.Errorf("invalid keepaliveTime %q, must be lower than holdTime %q", ka, ht)
	}

	htSeconds := uint64(ht.Seconds())
	kaSeconds := uint64(ka.Seconds())

	return &htSeconds, &kaSeconds, nil
}

```

api/v1alpha1/duration.go
qinqon: I will remove the hole duration.go thing, we don't really need it also if production code ned Duration it should have Duration convertion into internal, having helpers is againt kubernetes api standars, people importing just the API will import the helpers and helpers then to introduce extra imports.

operator/api/v1alpha1/openperouter_types.go
qinqon: @fedepaol so looks like all the fields are OPenPERouterSpec are optional, when this happend the linter enforce the upper field be optional too, if we make it requiered th elinter fails.

For the operator it kind of make sense since usually users will just create an empty OpenPERouter CR to trigger deployment.

fedepaol: but it would work even if you leave it as it was, as it will be fed with zero values. I think pointers makes sense if you want to differentiate between the user omitted it vs it put it empty.
Here doesn't make sense imo

.custom-gcl.yml
qinqon: Good point — moved the binary name into `.custom-gcl.yml` (`name: golangci-lint-kube-api-linter`) and dropped the `--name` flag from `hack/golangci-lint.sh`. The script now relies on the YAML config as the single source of truth.

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,120 @@
...
+	Address *string `json:"address,omitempty"`
 
-	// Port is the port to dial when establishing the session.
+	// port is the port to dial when establishing the session.
 	// Defaults to 179.
 	// +optional
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=16384
-	Port *uint16 `json:"port,omitempty"`
+	Port *int32 `json:"port,omitempty"`
```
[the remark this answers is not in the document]
qinqon: These defaults were not enforced before (no `+default=` marker nor programmatic default), only documented. Adding them now would be a behavioral change unrelated to this PR. Neither kube-api-linter nor Kubernetes API conventions require `+default=` markers when a value is documented as a default, so leaving as-is.

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,120 @@
...
-	PasswordSecret string `json:"passwordSecret,omitempty"`
+	PasswordSecret *string `json:"passwordSecret,omitempty"`
 
-	// HoldTime is the requested BGP hold time, per RFC4271.
-	// Defaults to 180s.
+	// holdTimeSeconds is the requested BGP hold time in seconds, per RFC4271.
+	// Defaults to 180.
 	// +optional
-	HoldTime *metav1.Duration `json:"holdTime,omitempty"`
+	HoldTimeSeconds *int64 `json:"holdTimeSeconds,omitempty"`
```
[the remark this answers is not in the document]
qinqon: These defaults were not enforced before (no `+default=` marker nor programmatic default), only documented. Adding them now would be a behavioral change unrelated to this PR. Neither kube-api-linter nor Kubernetes API conventions require `+default=` markers when a value is documented as a default, so leaving as-is.

api/v1alpha1/neighbor.go
```diff
@@ -2,121 +2,120 @@
...
 
-	// KeepaliveTime is the requested BGP keepalive time, per RFC4271.
-	// Defaults to 60s.
+	// keepaliveTimeSeconds is the requested BGP keepalive time in seconds, per RFC4271.
+	// Defaults to 60.
 	// +optional
-	KeepaliveTime *metav1.Duration `json:"keepaliveTime,omitempty"`
+	KeepaliveTimeSeconds *int64 `json:"keepaliveTimeSeconds,omitempty"`
```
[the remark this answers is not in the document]
qinqon: These defaults were not enforced before (no `+default=` marker nor programmatic default), only documented. Adding them now would be a behavioral change unrelated to this PR. Neither kube-api-linter nor Kubernetes API conventions require `+default=` markers when a value is documented as a default, so leaving as-is.

operator/api/v1alpha1/openperouter_types.go
qinqon: Ahh wait you mean optional without pointers.

**UPDATE**
```
operator/api/v1alpha1/openperouter_types.go:78:2: optionalfields: field OpenPERouter.Spec should be a pointer. (kubeapilinter)
	Spec OpenPERouterSpec `json:"spec,omitzero"`
	^
```
So looks like we cannot.

fedepaol: yep

qinqon: @fedepaol @maiqueb options:
- Make it a pointer so it keeps optional
- Keep it as struct and add `//nolint:kubeapilinter`

My two cents move it to pointer so we don't need to customize the linter

maiqueb: I agree with @fedepaol that there's no need to differentiate between a user provided empty configuration and user omitted. 

I'd lean towards getting the linter to stay away from it. 

TBH, everytime I look at the updated API, I like it less. 

I'm hoping I just need to update my standards.

qinqon: > I'd lean towards getting the linter to stay away from it.

@fedepaol you ok with this ? 

> TBH, everytime I look at the updated API, I like it less.

I think we are use to api as a golang struct representation, but it goes beyond that.

qinqon: anyhow pushed a `//nolint:kubeapilinter` fix in this case make sense to not use a pointer for the optional field.

fedepaol:
LGTM!
Thanks!

PR #323, opened 2026-04-20
CI gating improvements

fedepaol:

Trying to cover and automate a few recurring themes in CI

>
> /kind bug
kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

There are a lot of recurring patterns that can be (at least partially) automated by CI.
Here we try to cover them either with CI or by instructing gemini.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -19,6 +19,93 @@ Also see [Claude.md](../Claude.md) for detailed Go style guidelines including:
 - Dependency management
 - Control flow recommendations
 
+### Unnecessary Comments
+Remove obvious comments that restate what the code already says. Comments should explain *why*, never *what*. AI-generated boilerplate comments are especially problematic — they add noise without value.
+
+**Bad:** `// Create the client` above `client := NewClient()`
+**Good:** `// Retry with a new client because FRR drops idle connections after 90s`
+
+If removing the comment wouldn't confuse a future reader, remove it.
+
+### Code Organization — Newspaper Structure
+Follow [newspaper code structure](https://kentcdodds.com/blog/newspaper-code-structure): public/exported functions at the top, private/utility functions at the bottom. Callers before callees. This makes files scannable — readers find the "what" first, the "how" below.
+
+- Exported methods and types go at the top of the file
+- Exception: `init()` functions go at the top, near package-level vars they initialize
+- Unexported helpers go at the bottom
+- When a function calls another function in the same file, the caller should appear above the callee
+
+### Flat Control Flow — Early Returns Over Nested Ifs
+Avoid complex nested `if` conditions. Especially if they include ors and mix ands and ors or negations.
+
+**Bad:** complex condition guarding a nested block inline
+```go
+if (gvk.Group == "" && gvk.Kind == "Service") || gvk.Kind == "Foo" {
+    // logic
+}
+```
+
+**Good:** extract the block into a function with early returns for each condition
+```go
+
+func isKindToSkip(gvk *gvk) bool {
+    if gvk.Kind == "Foo" {
+        return true
+    }
+    if gvk.Group == "" && gvk.Kind == "Service" {
+        return true
+    }
+    return false
+}
+```
+
+```go
+if isKindToSkip(gvk) {
+    // logic
+}
+```
+
+### Return Early — Avoid Intermediate Variables
+When each branch of a switch or if/else produces a complete result, return it directly instead of assigning to intermediate variables and returning at the end. This makes each branch self-contained and avoids tracking state across the function.
+
+**Bad:** accumulate into variables, return once at the end
+```go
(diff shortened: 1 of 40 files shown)


fedepaol:
@maiqueb ptal. If the logic makes sense to you I'll proceed with adjusting the codebase.

.gemini/styleguide.md
```diff
@@ -19,6 +19,32 @@ Also see [Claude.md](../Claude.md) for detailed Go style guidelines including:
...
+
+**Bad:** `// Create the client` above `client := NewClient()`
+**Good:** `// Retry with a new client because FRR drops idle connections after 90s`
+
+If removing the comment wouldn't confuse a future reader, remove it.
+
+### Code Organization — Newspaper Structure
+Follow [newspaper code structure](https://kentcdodds.com/blog/newspaper-code-structure): public/exported functions at the top, private/utility functions at the bottom. Callers before callees. This makes files scannable — readers find the "what" first, the "how" below.
+
+- Exported methods and types go at the top of the file
```
maiqueb, line 33: correct. the only notable exception to this would be - imho - the `init()` function. Can we add a special reference to that ?

.golangci.yml
```diff
@@ -35,6 +66,9 @@ linters:
           - dupl
           - lll
         path: internal/*
+      - linters:
+          - forbidigo
+        path-except: '(_test\.go|^e2etests/)'
```
maiqueb, line 71: IIUC this is allowing time.Sleep in tests & e2e tests. 

I think we should avoid this, but, in all honesty, this is a solid step forward. At worst, we can merge it as is, and improve later on.

maiqueb (approved):
Tks for this PR.

.gemini/styleguide.md
fedepaol: yep, will add it

.golangci.yml
fedepaol: it's actually the other way around. This allows it everywhere but in tests.
It's an exception to the exclusion

maiqueb: tks for educating me. why would we allow sleeps in the production code ?

fedepaol: we should not, I just extracted the most common comments and they were related about sleeps in tests, hence the CI config. I'll expand to all the codebase

fedepaol: fixed

internal/conversion/host_conversion.go
```diff
@@ -159,6 +137,36 @@ func APItoHostConfig(nodeIndex int, targetNS string, underlayFromMultus bool, ap
...
+			client.ObjectKeyFromObject(l2vni),
+		)
+	}
+
+	return &hostnetwork.HostMaster{
+		Name:       name,
+		Type:       l2vni.Spec.HostMaster.Type,
+		AutoCreate: autoCreate,
+	}, nil
+}
```
maiqueb, line 171: I know you're only moving code, but I think we can also improve this a little bit, and maybe extract a bit of wisdom for claude & friends.

This is SUPER opinionated, but I think not having to deal with variables (and return early) makes the code more explicit. 

```suggestion
func convertHostMaster(l2vni *v1alpha1.L2VNI) (*hostnetwork.HostMaster, error) {
	switch l2vni.Spec.HostMaster.Type {
	case v1alpha1.LinuxBridge:
		if l2vni.Spec.HostMaster.LinuxBridge != nil {
		    return &hostnetwork.HostMaster{
        		Name:           l2vni.Spec.HostMaster.LinuxBridge.Name,
		        Type:             l2vni.Spec.HostMaster.Type,
		        AutoCreate: l2vni.Spec.HostMaster.LinuxBridge.AutoCreate,
       	}, nil
		}
	case v1alpha1.OVSBridge:
		if l2vni.Spec.HostMaster.OVSBridge != nil {
		    return &hostnetwork.HostMaster{
        		Name:           l2vni.Spec.HostMaster.OVSBridge.Name,
		        Type:             l2vni.Spec.HostMaster.Type,
		        AutoCreate: l2vni.Spec.HostMaster.OVSBridge.AutoCreate,
       	}, nil
		}
	default:
		return nil, fmt.Errorf(
			"unknown host master type %q for L2VNI %s",
			l2vni.Spec.HostMaster.Type,
			client.ObjectKeyFromObject(l2vni),
		)
	}
}
```

maiqueb (approved):
just a nit.

internal/conversion/host_conversion.go
fedepaol: yeah I like this more. Will try to find the proper wording to instruct gemini

fedepaol: done

maiqueb (approved):

PR #325, opened 2026-04-20
Fix health probe port race between static and k8s reconcilers

fedepaol:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

In host mode, when the K8s API is available immediately (e.g. on a control-plane node), stopStaticReconciler() cancels the context but the static manager's health probe listener on :9081 hasn't released the port yet. The K8s reconciler then fails to bind the same port with "address already in use", crashing the controller.

Wait for the static reconciler goroutine to fully exit before starting the K8s reconciler, ensuring the health probe port is released.

Spotted in https://github.com/openperouter/openperouter/actions/runs/24667715133/job/72130625439?pr=310

**Release note**:

```release-note
Fix start race condition where k8s api is available already, the static controller dies but the health port is not free yet, causing the k8sapicontroller to die because the port is not ready. 
```

### DIFF
--- a/cmd/hostcontroller/main.go
+++ b/cmd/hostcontroller/main.go
@@ -206,7 +206,9 @@ func runHostMode(
 	staticControllerCtx, stopStaticReconciler := context.WithCancel(ctx)
 	defer stopStaticReconciler()
 
+	staticDone := make(chan struct{})
 	go func() {
+		defer close(staticDone)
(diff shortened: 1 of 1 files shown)


maiqueb (approved):
tks. 

This makes sense to me. Good catch. 

I need to start playing w/ host mode ...

PR #329, opened 2026-04-21
enhancement: controller-provisioned underlay interfaces

maiqueb:

>
> /kind bug
> /kind cleanup
> /kind feature

/kind design

> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

Propose replacing `nics []string` and `--underlay-from-multus` with a discriminated-union `underlayInterface` field (physical | macvlan | ipvlan) to support named netns deployments where Multus cannot attach interfaces.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/enhancements/controller-provisioned-underlay-interfaces.md
+++ b/enhancements/controller-provisioned-underlay-interfaces.md
@@ -1,0 +1,615 @@
+# Controller-Provisioned Underlay Interfaces
+
+## Summary
+
+Replace the current underlay connectivity model — where the controller either
+moves a physical NIC into the router network namespace or delegates to Multus —
+with a single, explicit API field that lets the operator choose **how** the
+router connects to the physical underlay. The new `underlayInterface` field in
+the Underlay CRD uses a discriminated union (type enum + per-type sub-struct)
+to support three modes: `physical` (move the entire NIC), `macvlan` (derive a
+macvlan child), and `ipvlan` (derive an ipvlan child). The controller provisions
+the interface directly — Multus is no longer involved in underlay connectivity
+for the router.
+
+## Motivation
+
+### Goals
+
+- **Deprecate Multus for router underlay connectivity.** With the persistent
+  named netns model (see [router-resiliency.md](router-resiliency.md)), the
+  router runs as a `hostNetwork` pod or a Podman quadlet. Neither deployment
+  model integrates with Multus CNI, which operates on pod network namespaces
+  managed by the container runtime. The controller must provision underlay
+  interfaces itself.
+
+- **Explicit interface type selection.** Today, the operator must coordinate
+  two separate knobs — the `nics` field in the Underlay CRD and the
+  `--underlay-from-multus` controller flag — to express a single intent.
+  The new API collapses this into one declarative field with explicit type
+  semantics.
+
+- **Enable NIC sharing.** Moving a physical NIC into the router netns
+  (`physical` mode) makes the NIC exclusively available to the router. The
+  `macvlan` and `ipvlan` modes let the parent device remain on the host,
+  enabling scenarios such as:
+  - The host using the same NIC for management traffic.
+  - Multiple router instances (future redundant-instances enhancement) deriving
+    separate child interfaces from the same parent.
+
+- **Consistent API pattern.** Follow the existing discriminated-union pattern
+  established by `HostMaster` in the L2VNI CRD (`type` enum + per-type
+  sub-struct), so operators encounter a familiar structure across the API
+  surface.
+
+### Non-Goals
+
+- **SR-IOV support.** While the discriminated-union pattern makes it trivial to
+  add an `sriov` variant in the future, designing or implementing SR-IOV
+  support is out of scope for this enhancement.
+- **Changes to workload pod networking.** Multus-attached secondary interfaces
+  on application pods and KubeVirt VMs are completely unaffected.
+- **Redundant router instances.** macvlan/ipvlan enables NIC sharing, which is
+  a prerequisite for running multiple router instances per node, but the
+  multi-instance design itself is a separate enhancement.
+- **Per-node interface overrides.** All nodes matching the Underlay's
+  `nodeSelector` use the same interface configuration. Per-node customization
+  (e.g. different parent device names on different hardware) is handled by
+  creating multiple Underlay objects with non-overlapping node selectors.
+
+## Proposal
+
+### Overview
+
+Replace `UnderlaySpec.Nics []string` and the `--underlay-from-multus`
+controller flag with a single `underlayInterface` field. This field is a
+discriminated union: a `type` enum selects one of three modes, and a
+corresponding per-type sub-struct carries the parameters relevant to that mode.
+
+The three modes are:
+
+| Mode | Behavior | Host NIC availability |
+|------|----------|----------------------|
+| `physical` | Moves the entire NIC into the router netns | NIC is exclusively owned by the router |
+| `macvlan` | Creates a macvlan child from the parent device and moves the child into the router netns | Parent stays on the host |
+| `ipvlan` | Creates an ipvlan child from the parent device and moves the child into the router netns | Parent stays on the host |
+
+### API
+
+```yaml
+apiVersion: network.openperouter.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay
+spec:
+  asn: 64514
+  routeridcidr: "10.0.0.0/24"
+  underlayInterface:
+    type: macvlan
+    macvlan:
+      parentDevice: eth1
+      mode: bridge
+  evpn:
+    vtepcidr: "100.65.0.0/24"
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+#### Go Types
+
+```go
+// UnderlayInterface defines how the router connects to the physical underlay.
+// The type field selects the provisioning mode; exactly one of the corresponding
+// sub-structs (physical, macvlan, ipvlan) must be set to match the type.
+//
+// This follows the same discriminated-union pattern as HostMaster in L2VNISpec.
+//
+// +kubebuilder:validation:Required
+// +kubebuilder:validation:XValidation:rule="(self.type == 'physical' && has(self.physical) && !has(self.macvlan) && !has(self.ipvlan)) || (self.type == 'macvlan' && has(self.macvlan) && !has(self.physical) && !has(self.ipvlan)) || (self.type == 'ipvlan' && has(self.ipvlan) && !has(self.physical) && !has(self.macvlan))",message="type/config mismatch: the sub-struct must match the type field"
+type UnderlayInterface struct {
+	// Type selects how the router obtains underlay connectivity.
+	// +kubebuilder:validation:Enum=physical;macvlan;ipvlan
+	// +kubebuilder:validation:Required
+	Type string `json:"type"`
+
+	// Physical moves a host NIC exclusively into the router netns.
+	// Must be set when type is "physical".
+	// +optional
+	Physical *PhysicalInterface `json:"physical,omitempty"`
+
+	// Macvlan creates a macvlan child interface from a host NIC.
+	// The parent device stays on the host; the child is moved into the
+	// router netns. Must be set when type is "macvlan".
+	// +optional
+	Macvlan *MacvlanInterface `json:"macvlan,omitempty"`
+
+	// Ipvlan creates an ipvlan child interface from a host NIC.
+	// The parent device stays on the host; the child is moved into the
+	// router netns. Must be set when type is "ipvlan".
+	// +optional
+	Ipvlan *IpvlanInterface `json:"ipvlan,omitempty"`
+}
+
+// PhysicalInterface specifies a host NIC to move into the router netns.
+// The host loses access to this device while the router owns it.
+type PhysicalInterface struct {
+	// InterfaceName is the name of the host NIC to move.
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9._-]*$`
+	// +kubebuilder:validation:MaxLength=15
+	// +kubebuilder:validation:Required
+	InterfaceName string `json:"interfaceName"`
+}
+
+// MacvlanInterface specifies a macvlan child interface to create from a
+// host NIC. The parent device remains on the host.
+type MacvlanInterface struct {
+	// ParentDevice is the host NIC from which the macvlan is derived.
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9._-]*$`
+	// +kubebuilder:validation:MaxLength=15
+	// +kubebuilder:validation:Required
+	ParentDevice string `json:"parentDevice"`
+
+	// Mode is the macvlan operating mode.
+	// - bridge: all child interfaces can communicate directly (most common).
+	// - vepa: all traffic goes through the parent's external bridge/switch,
+	//   even between children on the same host.
+	// - private: children are completely isolated from each other.
+	// - passthru: a single child takes over the parent; only one child allowed.
+	// +kubebuilder:validation:Enum=bridge;vepa;private;passthru
+	// +kubebuilder:default=bridge
+	// +optional
+	Mode string `json:"mode,omitempty"`
+}
+
+// IpvlanInterface specifies an ipvlan child interface to create from a
+// host NIC. The parent device remains on the host. Unlike macvlan, ipvlan
+// shares the parent's MAC address — only one MAC appears on the wire.
+type IpvlanInterface struct {
+	// ParentDevice is the host NIC from which the ipvlan is derived.
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9._-]*$`
+	// +kubebuilder:validation:MaxLength=15
+	// +kubebuilder:validation:Required
+	ParentDevice string `json:"parentDevice"`
+
+	// Mode is the ipvlan operating mode.
+	// - l2: operates at L2, child participates in ARP (most common).
+	// - l3: operates at L3, no ARP/NDP — the parent acts as a router.
+	// - l3s: like l3, but with netfilter/conntrack integration.
+	// +kubebuilder:validation:Enum=l2;l3;l3s
+	// +kubebuilder:default=l2
+	// +optional
+	Mode string `json:"mode,omitempty"`
+}
+```
+
+#### Examples
+
+**Physical (current behavior, replaces `nics: [eth1]`):**
+```yaml
+underlayInterface:
+  type: physical
+  physical:
+    interfaceName: eth1
+```
+
+**Macvlan (replaces Multus-based underlay):**
+```yaml
+underlayInterface:
+  type: macvlan
+  macvlan:
+    parentDevice: eth1
+    mode: bridge
+```
+
+**Ipvlan (shared MAC, useful when MAC learning is constrained):**
+```yaml
+underlayInterface:
+  type: ipvlan
+  ipvlan:
+    parentDevice: eth1
+    mode: l2
+```
+
+### User Stories
+
+#### Story 1: Operator Migrating from Multus-Based Underlay
+
+As a cluster operator currently using Multus to plumb the underlay NIC into the
+router pod, I want to switch to the named netns deployment model without losing
+underlay connectivity, so that I benefit from persistent data plane resiliency.
+
+**Before (Multus):**
+```yaml
+# Underlay CRD — nics omitted, Multus handles it
+spec:
+  asn: 64514
+  evpn:
+    vtepcidr: "100.65.0.0/24"
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+---
+# Separate NetworkAttachmentDefinition + controller flag
+# --underlay-from-multus=true
+```
+
+**After (controller-provisioned macvlan):**
+```yaml
+spec:
+  asn: 64514
+  underlayInterface:
+    type: macvlan
+    macvlan:
+      parentDevice: eth1
+      mode: bridge
+  evpn:
+    vtepcidr: "100.65.0.0/24"
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+The intent is fully captured in one CRD. No separate NAD, no controller flag.
+
+#### Story 2: Operator Using Dedicated Physical NIC
+
+As a cluster operator with a dedicated physical NIC for underlay traffic, I want
+to move it exclusively into the router netns for maximum performance and
+isolation.
+
+```yaml
+underlayInterface:
+  type: physical
+  physical:
+    interfaceName: eth1
+```
+
+This is equivalent to the current `nics: [eth1]` behavior.
+
+#### Story 3: Operator Needing Shared NIC
+
+As a cluster operator on hardware with limited NICs, I want the host and the
+router to share the same physical NIC, so that I don't need a dedicated NIC for
+underlay traffic.
+
+```yaml
+underlayInterface:
+  type: macvlan
+  macvlan:
+    parentDevice: eth1
+    mode: bridge
+```
+
+The host retains `eth1` for management; the router gets a macvlan child with
+its own MAC and IP addressing.
+
+#### Story 4: Operator on a Network with MAC Restrictions
+
+As a cluster operator on a network that limits the number of MAC addresses per
+port (e.g. certain cloud or campus environments), I want the router's underlay
+interface to share the physical NIC's MAC address.
+
+```yaml
+underlayInterface:
+  type: ipvlan
+  ipvlan:
+    parentDevice: eth1
+    mode: l2
+```
+
+Only one MAC appears on the wire regardless of how many ipvlan children exist.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| Breaking change to `nics` field | This is a `v1alpha1` API with no stability guarantees. Provide clear migration documentation. The `physical` mode is a direct equivalent. |
+| macvlan/ipvlan performance overhead vs physical NIC | Minimal for macvlan in bridge mode (~1-2% throughput impact in benchmarks). Document that `physical` mode is preferred when a dedicated NIC is available. |
+| macvlan child cannot communicate with parent on the same host | This is a well-known Linux kernel constraint. Document it: if the host needs to reach the router via the underlay NIC, use `ipvlan` or a separate veth pair. |
+| ipvlan requires parent and children to share the same L3 subnet | Document the L3 implications of ipvlan mode selection. |
+| Operator forgets to set the sub-struct matching the type | CEL validation rejects the resource at admission time with a clear error message. |
+
+## Design Details
+
+### API Pattern: Discriminated Union
+
+The `UnderlayInterface` struct uses the same discriminated-union pattern as
+`HostMaster` in the L2VNI CRD:
+
+```go
+// L2VNI's HostMaster (existing pattern)
+type HostMaster struct {
+    Type        string             `json:"type"`         // "linux-bridge" | "ovs-bridge"
+    LinuxBridge *LinuxBridgeConfig `json:"linuxBridge,omitempty"`
+    OVSBridge   *OVSBridgeConfig   `json:"ovsBridge,omitempty"`
+}
+
+// UnderlayInterface (proposed — same pattern)
+type UnderlayInterface struct {
+    Type     string             `json:"type"`     // "physical" | "macvlan" | "ipvlan"
+    Physical *PhysicalInterface `json:"physical,omitempty"`
+    Macvlan  *MacvlanInterface  `json:"macvlan,omitempty"`
+    Ipvlan   *IpvlanInterface   `json:"ipvlan,omitempty"`
+}
+```
+
+A CEL validation on the struct enforces that exactly the sub-struct matching the
+`type` enum is set. This gives operators:
+- An explicit discriminator (`type`) that makes the intent unambiguous in YAML
+  and in code (switch on `Type`, not on which pointer is non-nil).
+- Per-type fields that are only present when relevant — no unused `macvlanMode`
+  field when the type is `physical`.
+- Clear `kubectl explain` output — each sub-struct is its own section with
+  contextual documentation.
+
+### Controller Provisioning Flow
+
+The controller's `SetupUnderlay` function is extended to handle the three
+interface types. The provisioning logic runs inside the controller's
+reconciliation loop, in the same phase where it currently moves the physical
+NIC.
+
+#### Physical Mode
+
+No change from current behavior:
+
+1. `moveUnderlayInterface(interfaceName, targetNS)` — moves the NIC from the
+   host netns into the router netns.
+2. Assigns the marker address (`172.16.1.1/32`) for detection.
+3. Sets the interface UP.
+
+#### Macvlan Mode
+
+1. Look up the parent device on the host by name.
+2. Create a macvlan link: `netlink.LinkAdd(&netlink.Macvlan{...})` with the
+   specified mode and parent.
+3. Move the macvlan child into the router netns.
+4. Assign the marker address for detection.
+5. Set the interface UP.
+6. The parent device stays on the host, untouched.
+
+#### Ipvlan Mode
+
+1. Look up the parent device on the host by name.
+2. Create an ipvlan link: `netlink.LinkAdd(&netlink.IPVlan{...})` with the
+   specified mode and parent.
+3. Move the ipvlan child into the router netns.
+4. Assign the marker address for detection.
+5. Set the interface UP.
+6. The parent device stays on the host, untouched.
+
+### Naming Convention
+
+The macvlan/ipvlan child interface is created with a deterministic name derived
+from the parent device. The interface is renamed inside the router netns to
+match the `interfaceName` / `parentDevice` so that FRR configuration and
+existing references work consistently:
+
+- **Physical**: interface keeps its original name (e.g. `eth1`).
+- **Macvlan/Ipvlan**: child is renamed to `ul-<parentDevice>` in the router
+  netns (e.g. `ul-eth1`). The `ul-` prefix signals "underlay" and avoids
+  collisions with the parent name (which lives in a different netns but could
+  cause confusion in logs and debugging).
+
+### Idempotency and Reconciliation
+
+The marker address (`172.16.1.1/32`) detection mechanism already used for
+physical NICs applies identically to macvlan/ipvlan children:
+
+- If the marker is found on the expected interface in the router netns, the
+  underlay is already configured — no-op.
+- If the marker is found on a **different** interface (e.g. the operator changed
+  `parentDevice`), return `UnderlayExistsError` to signal that the netns must be
+  rebuilt.
+- If no marker is found, provision the interface.
+
+### Underlay Change Detection
+
+Changing the underlay interface type or parent device is a destructive operation
+that requires a netns rebuild. The controller detects this via the
+`UnderlayExistsError` returned by `SetupUnderlay` when the configured underlay
+doesn't match the desired one. This triggers `HandleNonRecoverableError`, which
+deletes the netns and lets the next reconciliation rebuild it from scratch.
+
+### Removal of `--underlay-from-multus` Flag
+
+The `--underlay-from-multus` controller flag becomes unnecessary. Its current
+purpose is to tell the controller "don't validate `nics` — Multus will provide
+the underlay interface." With the new API, the `type` field is the single source
+of truth for how the underlay is provisioned. The flag is removed.
+
+### Changes Required
+
+| Component | Current | Proposed | Effort |
+|-----------|---------|----------|--------|
+| `UnderlaySpec` | `Nics []string` | `UnderlayInterface *UnderlayInterface` | API type change |
+| New types | — | `UnderlayInterface`, `PhysicalInterface`, `MacvlanInterface`, `IpvlanInterface` | New Go types |
+| CEL validation | NIC name pattern on items | Discriminated-union match (type ↔ sub-struct) | New validation rule |
+| `--underlay-from-multus` flag | Boolean flag in `cmd/hostcontroller/main.go` | Removed | Delete flag + wiring |
+| `SetupUnderlay()` | Moves physical NIC | Switch on type: move NIC / create macvlan / create ipvlan | Extend function |
+| `UnderlayParams` | `UnderlayInterface string` | Carries full `UnderlayInterface` struct | Struct change |
+| `APItoHostConfig()` | Reads `Nics[0]` + `underlayFromMultus` | Reads `UnderlayInterface.Type` + sub-struct fields | Conversion update |
+| Webhook validation | NIC name format | Type/sub-struct consistency | Update webhook |
+| Helm values schema | `nics` list | `underlayInterface` block | Values schema change |
+| Helm templates | Renders `nics` | Renders `underlayInterface` with type | Template update |
+| E2E test setup | `CreateMacvlanNad` for Multus underlay | Direct `underlayInterface` in CRD | Test simplification |
+| Documentation | Describes `nics` + Multus flag | Describes three interface modes | Doc rewrite |
+
+### Migration Path
+
+Since this is a `v1alpha1` API, no automatic migration is provided. The
+migration is a one-time CRD update:
+
+| Current configuration | New configuration |
+|----------------------|-------------------|
+| `nics: [eth1]` | `underlayInterface: {type: physical, physical: {interfaceName: eth1}}` |
+| `nics: []` + `--underlay-from-multus=true` + Multus NAD for underlay | `underlayInterface: {type: macvlan, macvlan: {parentDevice: eth1, mode: bridge}}` |
+
+The release notes will include a migration guide with examples for both paths.
+
+### Test Plan
+
+- **Unit tests**: `UnderlayInterface` CEL validation — verify that:
+  - `type: physical` with `physical` set is accepted.
+  - `type: macvlan` with `macvlan` set is accepted.
+  - `type: ipvlan` with `ipvlan` set is accepted.
+  - Type/sub-struct mismatch is rejected with a clear message.
+  - Missing sub-struct is rejected.
+  - Multiple sub-structs set simultaneously is rejected.
+- **Unit tests**: `SetupUnderlay` with macvlan and ipvlan modes — verify
+  interface creation, parent device lookup, namespace move, marker assignment,
+  and idempotent re-runs (using network namespace mocks or netns test fixtures).
+- **Integration tests**: In a real (non-mock) netns environment, verify:
+  - macvlan child is created from the parent and moved into the target netns.
+  - ipvlan child is created from the parent and moved into the target netns.
+  - Parent device remains in the host netns and is functional.
+  - FRR discovers the interface inside the netns.
+  - `HasUnderlayInterface` correctly detects the provisioned interface.
+- **Reconciliation tests**: Verify that changing the underlay type (e.g.
+  `physical` to `macvlan`) triggers `UnderlayExistsError` and the controller
+  handles it via netns rebuild.
+- **E2E tests**: Full traffic test with each interface type:
+  - `physical`: existing E2E coverage applies.
+  - `macvlan`: underlay traffic flows through the macvlan child; VXLAN tunnels
+    establish and carry EVPN routes.
+  - `ipvlan`: underlay traffic flows through the ipvlan child; verify that
+    shared MAC does not interfere with EVPN Type-2 route advertisement.
+- **Migration test**: Apply an Underlay with the old `nics` field, upgrade the
+  controller, apply the new `underlayInterface` field, verify convergence.
+
+### Graduation Criteria
+
+#### Alpha
+
+- `UnderlayInterface` type and sub-structs added to the API with CEL
+  validation.
+- `Nics` field removed from `UnderlaySpec`.
+- `--underlay-from-multus` flag removed from the controller.
+- `SetupUnderlay` extended to handle `physical`, `macvlan`, and `ipvlan` modes.
+- `physical` mode is functionally equivalent to the current `nics` behavior.
+- Basic unit and integration tests for all three modes.
+
+#### Beta
+
+- E2E tests covering all three modes with real traffic verification.
+- Helm chart and operator bindata updated with new values schema.
+- Migration guide published in release notes.
+- macvlan and ipvlan modes validated in at least one real-hardware environment
+  (not just Kind/containerlab).
+
+#### GA
+
+- Production validation across multiple deployment environments and NIC
+  vendors.
+- Performance benchmarks comparing physical, macvlan, and ipvlan modes
+  published.
+- Documentation covers mode selection guidance (when to use each mode).
+
+## Drawbacks
+
+- **Breaking API change.** Operators must update their Underlay CRD when
+  upgrading. Since this is `v1alpha1` and the migration is a straightforward
+  field rename, this is acceptable.
+- **Slightly more verbose YAML.** The nested sub-struct adds one level of
+  indentation compared to `nics: [eth1]`. This is the intentional trade-off for
+  type safety — every field visible in the YAML is relevant to the selected
+  mode.
+- **macvlan/ipvlan have known kernel constraints.** macvlan children cannot
+  communicate with the parent on the same host. ipvlan children share the
+  parent's MAC address, which affects ARP behavior. These are Linux kernel
+  properties, not API design issues, but they must be documented so operators
+  make informed choices.
+
+## Alternatives
+
+### Alternative 1: Flat Discriminated Union (Single Struct, No Sub-Structs)
+
+Use a `type` enum with mode-specific fields as optional flat fields on the same
+struct:
+
+```yaml
+underlayInterface:
+  parentDevice: eth1
+  type: macvlan
+  macvlanMode: bridge
+```
+
+```go
+type UnderlayInterface struct {
+    ParentDevice string       `json:"parentDevice"`
+    Type         string       `json:"type"`
+    MacvlanMode  *MacvlanMode `json:"macvlanMode,omitempty"`
+    IpvlanMode   *IpvlanMode  `json:"ipvlanMode,omitempty"`
+}
+```
+
+**Why not chosen:**
+
+- **Unused fields leak into the YAML.** `macvlanMode` appears in the struct
+  even when `type: physical`. Operators see it in `kubectl explain` and wonder
+  if they should set it. Conditional field relevance is a source of confusion.
+- **Conditional validation.** "macvlanMode required when type=macvlan AND
+  forbidden when type=physical" is a chain of rules. The sub-struct pattern
+  makes invalid combinations structurally impossible.
+- **Naming ambiguity.** `parentDevice` means "device to move" for `physical`
+  and "device to derive from" for `macvlan`/`ipvlan`. The sub-struct pattern
+  uses `interfaceName` for physical and `parentDevice` for macvlan/ipvlan,
+  matching the semantic difference.
+- **Extensibility.** Adding SR-IOV would add `numVFs`, `resourceName`,
+  `pfName`, etc. as flat optional fields — the struct grows unwieldy. With
+  sub-structs, `Sriov *SriovInterface` is self-contained.
+
+### Alternative 2: One-of Sub-Structs Without Type Enum
+
+Use mutually exclusive sub-struct pointers without an explicit `type`
+discriminator, following the Kubernetes `VolumeSource` pattern:
+
+```yaml
+underlayInterface:
+  macvlan:
+    parentDevice: eth1
+    mode: bridge
+```
+
+```go
+type UnderlayInterface struct {
+    Physical *PhysicalInterface `json:"physical,omitempty"`
+    Macvlan  *MacvlanInterface  `json:"macvlan,omitempty"`
+    Ipvlan   *IpvlanInterface   `json:"ipvlan,omitempty"`
+}
+```
+
+**Why not chosen:**
+
+- **No explicit discriminator.** Code must check which pointer is non-nil to
+  determine the type. A `Type` enum makes switch statements clean and
+  serialization unambiguous.
+- **Inconsistent with existing API.** The `HostMaster` struct in L2VNI already
+  uses the type-enum + sub-struct pattern. Mixing two different union patterns
+  in the same API surface creates inconsistency.
+- **JSON round-trip ambiguity.** Without a discriminator, a malformed resource
+  with two sub-structs set is harder to diagnose — the error is "exactly one
+  must be set" rather than "the sub-struct must match the type."
+
+### Alternative 3: Keep Multus for Underlay
+
+Continue using Multus `NetworkAttachmentDefinition` to plumb the underlay
+interface into the router pod.
+
+**Why not chosen:**
+
+- **Incompatible with named netns.** Multus operates on pod network namespaces
+  managed by the container runtime. With `hostNetwork: true` pods or Podman
+  quadlets, the router does not have a runtime-managed netns for Multus to
+  target.
+- **Split configuration.** The underlay intent is split across the Underlay CRD,
+  a `NetworkAttachmentDefinition`, and a controller flag. This makes the system
+  harder to reason about and debug.
+- **External dependency.** Multus is an optional cluster component. Requiring it
+  for router underlay connectivity adds a dependency that not all deployments
+  want.
+
+## Implementation History
+
+- 2026-04-21: Initial proposal drafted.
--- a/enhancements/controller-provisioned-underlay-interfaces.md
+++ b/enhancements/controller-provisioned-underlay-interfaces.md
@@ -1,0 +1,497 @@
+# Controller-Provisioned Underlay Interfaces
+
+## Summary
+
+Add a `CNI` provisioning mode to the `interfaces` union introduced by
+[PR #341](https://github.com/openperouter/openperouter/pull/341), so the
+controller can invoke any CNI plugin via
+[libcni](https://pkg.go.dev/github.com/containernetworking/cni/libcni)
+to provision underlay interfaces in the router netns. Replaces the removed
+Multus-based underlay integration.
+
+## Motivation
+
+### Goals
+
+- **Replace the removed Multus underlay integration.** The router resiliency
+  work removed the `--underlay-from-multus` controller flag and all associated
+  code paths. With the persistent named netns model (see
+  [router-resiliency.md](router-resiliency.md)), the router runs as a
+  `hostNetwork` pod or a Podman quadlet — neither integrates with Multus CNI,
+  which operates on pod network namespaces managed by the container runtime.
+  Users who relied on Multus-based macvlan underlay for NIC sharing currently
+  have **no alternative**.
+
+- **Flexible NIC sharing and IPAM.** Operators need to choose how the
+  underlay NIC is shared (macvlan, ipvlan, SR-IOV, bridge, OVS) and how IPs
+  are assigned (static, DHCP, pool-based) based on their network environment.
+
+- **Work in both Kubernetes and host/systemd modes.** Operators can
+  provision CNI-based underlay interfaces regardless of the deployment
+  mode. The same interface types and IPAM options are available in both
+  environments.
+
+- **Support day-0 operations.** New deployments should be able to install
+  OpenPERouter and have underlay interfaces provisioned automatically on
+  first startup, without manual pre-configuration of network devices on
+  each node.
+
+- **Consistent API pattern.** Extend the discriminated-union pattern already
+  established by `UnderlayInterface` and `HostMaster` in the L2VNI CRD, so
+  operators encounter a familiar structure across the API surface.
+
+### Non-Goals
+
+- **Implementing per-interface-type provisioning logic in the controller.**
+  Macvlan, ipvlan, SR-IOV, OVS, and bridge interface creation is delegated
+  entirely to CNI plugins. The controller invokes the plugin; the plugin
+  handles netlink operations.
+
+- **Redundant router instances.** NIC sharing via CNI plugins (e.g. macvlan)
+  is a prerequisite for running multiple router instances per node, but the
+  multi-instance design itself is a separate enhancement.
+
+- **Plugin-specific field validation.** The controller validates structural
+  plugin JSON config correctness and plugin binary existence, but does not
+  validate plugin-specific fields (e.g. checking `master` device exists
+  for macvlan).
+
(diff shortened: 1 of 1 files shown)


enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,615 @@
...
+  `nodeSelector` use the same interface configuration. Per-node customization
+  (e.g. different parent device names on different hardware) is handled by
+  creating multiple Underlay objects with non-overlapping node selectors.
+
+## Proposal
+
+### Overview
+
+Replace `UnderlaySpec.Nics []string` and the `--underlay-from-multus`
+controller flag with a single `underlayInterface` field. This field is a
```
maiqueb, line 65: maybe having an `underlayInterface` attribute in an `underlay` struct is not good. It stutters. 

suggestions ? @qinqon / @fedepaol

maiqueb:
@gemini-code-assist review

maiqueb:
@gemini-code-assist review

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,629 @@
...
+4. Assign the link group ID (`4242`) for detection.
+5. Set the interface UP.
+6. The parent device stays on the host, untouched.
+
+### Naming Convention
+
+All underlay interfaces — regardless of provisioning mode — are renamed to
+`ul-<deviceName>` inside the router netns. This ensures FRR configuration,
+`HasUnderlayInterface` detection, and controller logic use a single,
+predictable naming scheme with no type-dependent branching:
```
qinqon: Since we have the net group id, why do we need this ?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,629 @@
...
+`UnderlayExistsError` returned by `SetupUnderlay` when the configured underlay
+doesn't match the desired one. This triggers `HandleNonRecoverableError`, which
+deletes the netns and lets the next reconciliation rebuild it from scratch.
+
+### Removal of `--underlay-from-multus` Flag
+
+The `--underlay-from-multus` controller flag becomes unnecessary. Its current
+purpose is to tell the controller "don't validate `nics` — Multus will provide
+the underlay interface." With the new API, the `type` field is the single source
+of truth for how the underlay is provisioned. The flag is removed.
```
qinqon: This was already removed, clean this up.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,629 @@
...
+
+Since this is a `v1alpha1` API, no automatic migration is provided. The
+migration is a one-time CRD update:
+
+| Current configuration | New configuration |
+|----------------------|-------------------|
+| `nics: [eth1]` | `underlayInterface: {type: physical, physical: {interfaceName: eth1}}` |
+| `nics: []` + `--underlay-from-multus=true` + Multus NAD for underlay | `underlayInterface: {type: macvlan, macvlan: {parentDevice: eth1, mode: bridge}}` |
+
+The release notes will include a migration guide with examples for both paths.
```
qinqon: We don't need migration path, we can break things since we are a v1alpoha1

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,629 @@
...
+- macvlan and ipvlan modes validated in at least one real-hardware environment
+  (not just Kind/containerlab).
+
+#### GA
+
+- Production validation across multiple deployment environments and NIC
+  vendors.
+- Performance benchmarks comparing physical, macvlan, and ipvlan modes
+  published.
+- Documentation covers mode selection guidance (when to use each mode).
```
qinqon: Not sure we need that since we are going to deliver v1alpha1

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,629 @@
...
+
+## Drawbacks
+
+- **Breaking API change.** Operators must update their Underlay CRD when
+  upgrading. Since this is `v1alpha1` and the migration is a straightforward
+  field rename, this is acceptable.
```
qinqon: We don't care

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,629 @@
...
+- **Slightly more verbose YAML.** The nested sub-struct adds one level of
+  indentation compared to `nics: [eth1]`. This is the intentional trade-off for
+  type safety — every field visible in the YAML is relevant to the selected
+  mode.
+- **macvlan/ipvlan have known kernel constraints.** macvlan children cannot
+  communicate with the parent on the same host. ipvlan children share the
+  parent's MAC address, which affects ARP behavior. These are Linux kernel
+  properties, not API design issues, but they must be documented so operators
+  make informed choices.
```
qinqon: This is not related to openperouter, they are solutions for different problem with different drawbacks, users already know them.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,556 @@
...
+  modes let the parent device remain on the host, enabling scenarios such as:
+  - The host using the same NIC for management traffic.
+  - Multiple router instances (future redundant-instances enhancement) deriving
+    separate child interfaces from the same parent.
+
+- **Consistent API pattern.** Extend the discriminated-union pattern already
+  established by `UnderlayInterface` (PR #341) and `HostMaster` in the L2VNI
+  CRD, so operators encounter a familiar structure across the API surface.
+
+### Non-Goals
```
qinqon, line 43: Let's also add normal Vlans

qinqon (changes requested):
Are we going to add the IPAM here or that will be a follow up ?

maiqueb:
@gemini-code-assist review

maiqueb:
@gemini-code-assist review

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: where do you want me to add it ? 

like, it should be possible, why not ? just use a NAD w/ the [VLAN](https://www.cni.dev/plugins/current/main/vlan/) plugin referenced.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,1258 @@
...
+Extend the `interfaces` discriminated-union slice introduced by the API
+improvements enhancement ([PR #341](https://github.com/openperouter/openperouter/pull/341))
+with a new provisioning mode — `CNI` — so the controller can invoke any CNI
+plugin to provision underlay interfaces in the router netns. PR #341 defines the
+union and its first mode (`NetworkDevice`, which moves an existing host device
+into the router netns). This enhancement adds the `CNI` mode, which delegates
+interface creation and IPAM to the CNI plugin ecosystem via
+[libcni](https://pkg.go.dev/github.com/containernetworking/cni/libcni),
+replacing the Multus-based underlay integration removed in PR #317.
+
```
qinqon: Simplify this, don't reference the PR just say "interfaces" exist and don't explain the NetworkDevice stuff, just mention that we are going to incrase "interfaces" types with CNI

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,1258 @@
...
+  part of the router resiliency work. With the persistent named netns model
+  (see [router-resiliency.md](router-resiliency.md)), the router runs as a
+  `hostNetwork` pod or a Podman quadlet — neither integrates with Multus CNI,
+  which operates on pod network namespaces managed by the container runtime.
+  Users who relied on Multus-based macvlan underlay for NIC sharing currently
+  have **no alternative**. The `CNI` mode in this enhancement restores that
+  capability by invoking CNI plugins via `libcni` — the same library Multus
+  and other CNI meta-plugins use — so operators reuse existing
+  `NetworkAttachmentDefinition` (NAD) configurations without Multus itself.
+
```
qinqon: The goal is to extend the underlay interfaces configuration to support mechanism like macvlan or ipvlan, then just mention that the enhancement propose to use CNI for this, don't mention that it got removed here, is not part of the goal

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,1258 @@
...
+- **Enable NIC sharing without per-type controller code.** The `NetworkDevice`
+  mode (PR #341) moves a host device exclusively into the router netns. The
+  `CNI` mode delegates interface creation to the plugin ecosystem — macvlan,
+  ipvlan, bridge, SR-IOV, OVS — without the controller implementing per-type
+  netlink provisioning logic. Each new interface type works via a NAD config
+  change, not a controller code change.
+
```
qinqon: I don't understand any of this, feels like a mix of stuff without much meaning or goal

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,1258 @@
...
+- **Consistent API pattern.** Extend the discriminated-union pattern already
+  established by `UnderlayInterface` (PR #341) and `HostMaster` in the L2VNI
+  CRD, so operators encounter a familiar structure across the API surface.
+
```
qinqon, line 42: I don't think we are touching L2VNI right ?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,1258 @@
...
+##### NetworkDevice
+
+Moves an existing host device into the router netns (PR #341):
+
+```yaml
+interfaces:
+  - type: NetworkDevice
+    networkDevice:
+      interfaceName: eth1
+```
```
qinqon: This is not part of the enhancement, remove it

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,1258 @@
...
+  "name": "macvlan-underlay",
+  "type": "macvlan",
+  "master": "eth1",
+  "mode": "bridge",
+  "ipam": {
+    "type": "host-local",
+    "subnet": "192.168.1.0/24",
+    "rangeStart": "192.168.1.10",
+    "rangeEnd": "192.168.1.50"
+  }
```
qinqon: Remove the IPAM part from here since that's different chapter

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,1258 @@
...
+
+```yaml
+interfaces:
+  - type: CNI
+    cni:
+      networkAttachmentDefinition:
+        name: ipvlan-underlay
+        namespace: default
+```
+
```
qinqon, line 194: Let's replace RawConfig with Content, we already know this is Config
Also we don't need the extra config level 
```yaml
interfaces:
- type: CNI
  cni:
    config:
      type: Content
      content:
        cniVersion: "1.0.0"
        name: macvlan-underlay
        plugins:
          - type: macvlan
            master: eth1
            mode: bridge
            ipam:
              type: host-local
              subnet: "192.168.1.0/24"
              rangeStart: "192.168.1.10"
              rangeEnd: "192.168.1.50"
```

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,1258 @@
...
+  - type: NetworkDevice
+    networkDevice:
+      interfaceName: eth1
+      ipam:
+        type: Native
+        native:
+          cidrs:
+            - "192.168.1.0/24"
+```
+
```
qinqon: I don't think we need a type why not just cidr field ? at same level as interfaceName ?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,855 @@
...
+controller can invoke any CNI plugin via
+[libcni](https://pkg.go.dev/github.com/containernetworking/cni/libcni)
+to provision underlay interfaces in the router netns. Replaces the removed
+Multus-based underlay integration.
+
+## Motivation
+
+### Goals
+
+- **Replace the removed Multus underlay integration.** The router resiliency
```
qinqon, line 16: This reads weird, you cannot replace what is delete, the goal is to extend interfaces NetworkDevice to add other mechanism than "move nic", the fact that is done with CNI is an implementation details

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,855 @@
...
+- **Support day-0 operations.** New deployments should be able to install
+  OpenPERouter and have underlay interfaces provisioned automatically on
+  first startup, without manual pre-configuration of network devices on
+  each node.
+
+- **Consistent API pattern.** Extend the discriminated-union pattern already
+  established by `UnderlayInterface` and `HostMaster` in the L2VNI CRD, so
+  operators encounter a familiar structure across the API surface.
+
+### Non-Goals
```
qinqon, line 43: nit: Add new line between bullets

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,855 @@
...
+
+- **Implementing per-interface-type provisioning logic in the controller.**
+  Macvlan, ipvlan, SR-IOV, OVS, and bridge interface creation is delegated
+  entirely to CNI plugins. The controller invokes the plugin; the plugin
+  handles netlink operations.
```
qinqon, line 48: This is implementation details, is just saying "don't do golang netlink directly" as non goals, this is low level for a non-goal

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,855 @@
...
+- **Changes to workload pod networking.** Multus-attached secondary interfaces
```
qinqon: I would remove this one, does not make much sense, underlay is underlay

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+  on application pods and KubeVirt VMs are completely unaffected.
+- **Redundant router instances.** NIC sharing via CNI plugins (e.g. macvlan)
+  is a prerequisite for running multiple router instances per node, but the
+  multi-instance design itself is a separate enhancement.
+- **Per-node interface overrides.** All nodes matching the Underlay's
+  `nodeSelector` use the same interface configuration. Per-node customization
+  (e.g. different parent device names on different hardware) is handled by
+  creating multiple Underlay objects with non-overlapping node selectors.
```
qinqon: So this is a non goal because we can already do it ? I would remove this then, sice it's already implemeted

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+- **File watcher for CNI conflist directory changes.** The controller
+  reconciles on static config or CRD changes, not on conflist file changes
+  independently.
```
qinqon: This is just a drawback of one of the options, but not a "non goals" is very
implementation details

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+assignments per router, I want to assign a deterministic static IP to each
+node's underlay interface from a single configuration.
+
+#### Story 4: Migrating from Multus
+
+As an operator who previously used Multus to plumb the underlay NIC into
+the router pod, I want a replacement that restores NIC sharing without
+Multus, so that I can upgrade to the named netns deployment model without
+requiring a dedicated physical NIC.
+
```
qinqon: We don't need that since we don't need migration from old versions, we are allow to break things up.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+  asn: 64514
+  routeridcidr: "10.0.0.0/24"
+  interfaces:
+    - type: CNI
+      cni:
+        config:
+          type: API
+          api:
+            name: macvlan-underlay
+            namespace: default
```
qinqon: this can just be cni.[type|api]  we don't need the extra "config"

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+(`k8s.cni.cncf.io/v1`). The NAD CRD can be installed standalone; Multus
+is not required.
+
+#### Examples
+
+##### CNI with macvlan
+
+Replaces Multus-based underlay — the parent NIC stays on the host.
+
+**Kubernetes mode (API config source):**
```
qinqon: I would rename this on kubernetes API server mode

qinqon: I would rename this on "kubernetes API server mode" and explain at what scenario works and what it do not on a little prhase behind this

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+            namespace: default
+  tunnelEndpoint:
+    cidrs:
+      - "100.65.0.0/24"
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+**Host/systemd mode (Path config source):**
```
qinqon: Name this just "File path mode" saying this works just for host/systemd mode is not true
you can go and use MCO to install a file and use it at openperouter pod mode.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+            filePath: /etc/openperouter/cni/net.d/macvlan-underlay.conflist
+  tunnelEndpoint:
+    cidrs:
+      - "100.65.0.0/24"
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+**Either mode (RawConfig — self-contained):**
```
qinqon: Same here just say this is Raw config mode, and mention that it allows to be used a three openperouter modes

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+```yaml
+interfaces:
+  - type: CNI
+    cni:
+      config:
+        type: RawConfig
+        rawConfig:
+          config:
```
qinqon: Remove this "config" level is not needed

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+      asn: 65000
+```
+
+**Either mode (RawConfig — self-contained):**
+```yaml
+interfaces:
+  - type: CNI
+    cni:
+      config:
+        type: RawConfig
```
qinqon: I would call this just "Content"

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+                - type: macvlan
+                  master: eth1
+                  mode: bridge
+                  capabilities:
+                    ips: true
+                  ipam:
+                    type: static
+        runtimeConfig:
+          ips:
+            - "192.168.1.10/24"
```
qinqon: This is not a good use case for exercise the runtimeConfig thing since
we can directly pass it as part of the content

```yaml
  interfaces:
    - type: CNI
      cni:
        config:
          type: RawConfig
          rawConfig:
            config:
              cniVersion: "1.0.0"
              name: macvlan-static
              plugins:
                - type: macvlan
                  master: eth1
                  mode: bridge
                  capabilities:
                    ips: true
                  ipam:
                    type: static
                    addresses:
                    - 192.168.1.100/24
```

The idea behind runtimConfig stuff was to pass the openperouter IPAM IP there, so I think this is just implementation details, we don't want a hyper flexible extra runtimeConfig parameter stuff on the APII would just aim for openperouter IPAM integration with specific stuff

```yaml
interfaces:
    - type: CNI
      cni:
        cidrs:
        - 192.168.1.0/24
        config:
          type: RawConfig
          rawConfig:
            config:
              cniVersion: "1.0.0"
              name: macvlan-static
              plugins:
                - type: macvlan
                  master: eth1
                  mode: bridge
                  capabilities:
                    ips: true
                  ipam:
                    type: static
```

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,849 @@
...
+          config:
+            cniVersion: "1.0.0"
+            name: macvlan-underlay
+            plugins:
+              - type: macvlan
+                master: eth1
+                mode: bridge
+                ipam:
+                  type: dhcp
+      interfaceName: underlay0
```
qinqon, line 255: Do we really need this feature ?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,677 @@
...
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9._-]*$`
+	// +kubebuilder:validation:MaxLength=15
+	// +required
+	InterfaceName string `json:"interfaceName,omitempty"`
+
+	// +optional
+	IPAM *InterfaceIPAM `json:"ipam,omitempty"`
+}
+
+// CNIDevice specifies how to invoke a CNI plugin to provision an underlay
```
qinqon: I would just use "mac" at rawConfig level, the runtimeConfig part is implementation details
```yaml
  interfaces:
    - type: CNI
      cni:
        mac: "02:42:c0:a8:01:0a"
        config:
          type: RawConfig
          rawConfig:
            config:
              cniVersion: "1.0.0"
              name: macvlan-dhcp
              plugins:
                - type: macvlan
                  master: eth1
                  mode: bridge
                  capabilities:
                    mac: true
                  ipam:
                    type: dhcp

```

qinqon: I would just use "mac" at rawConfig level, the runtimeConfig part is implementation details
```yaml
  interfaces:
    - type: CNI
      cni:
        mac: "02:42:c0:a8:01:0a"
        config:
          type: RawConfig
          rawConfig:
            config:
              cniVersion: "1.0.0"
              name: macvlan-dhcp
              plugins:
                - type: macvlan
                  master: eth1
                  mode: bridge
                  capabilities:
                    mac: true
                  ipam:
                    type: dhcp
```
btw another option is to make openperouter generate the mac address for us so we just specify the macPrefix and openperouter use the node id to generate it, this way we don't need an underlay per node just because we want stable macs

```yaml
  interfaces:
    - type: CNI
      cni:
        macPrefix: "02:42:c0:a8:01:"
        config:
          type: RawConfig
          rawConfig:
            config:
              cniVersion: "1.0.0"
              name: macvlan-dhcp
              plugins:
                - type: macvlan
                  master: eth1
                  mode: bridge
                  capabilities:
                    mac: true
                  ipam:
                    type: dhcp

**Review by qinqon — CHANGES_REQUESTED** [review]
First pass

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 32 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: > I would just use "mac" at rawConfig level, the runtimeConfig part is implementation details
> 
> ```yaml
>   interfaces:
>     - type: CNI
>       cni:
>         mac: "02:42:c0:a8:01:0a"
>         config:
>           type: RawConfig
>           rawConfig:
>             config:
>               cniVersion: "1.0.0"
>               name: macvlan-dhcp
>               plugins:
>                 - type: macvlan
>                   master: eth1
>                   mode: bridge
>                   capabilities:
>                     mac: true
>                   ipam:
>                     type: dhcp
> ```

I would agree, if the user did NOT have to specify capabilities in the configuration. Since they have, this "implementation detail" is relevant. 

> 
> btw another option is to make openperouter generate the mac address for us so we just specify the macPrefix and openperouter use the node id to generate it, this way we don't need an underlay per node just because we want stable macs
> 
> ```yaml
>   interfaces:
>     - type: CNI
>       cni:
>         macPrefix: "02:42:c0:a8:01:"
>         config:
>           type: RawConfig
>           rawConfig:
>             config:
>               cniVersion: "1.0.0"
>               name: macvlan-dhcp
>               plugins:
>                 - type: macvlan
>                   master: eth1
>                   mode: bridge
>                   capabilities:
>                     mac: true
>                   ipam:
>                     type: dhcp
> ```

Could make sense - please kick off an enhancement on top of this if you want to pursue this route.

Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: as I said: I could agree but the user must allow this via the configuration via capabilities. 

Hence, it is not *just* implementation details.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 15 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md:194]: nope, let's not. content can be anything. might as well call it "bag" because you can put anything in it.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 12 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md:42]: no. what does that have to do with re-using established API patterns ?

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 19 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md:48]: a non goal imho deserves an explanation.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 17 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md:16]: sure you can - let's take this exampel: you had a mobile phone, but you lost it. 

You go to the store to buy a new one, essentially replacing the one you lost. 

I'm just saying the goal is to replace the "lost" functionality.  

Where am I referring CNI as the implementation medium for this ?

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 11 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: this has been removed long time ago.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 10 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: you're commenting on old code.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 9 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: this has been reworked already. please review the new version.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 16 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: this is no longer part of the enhancement

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 22 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: this has been removed already

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 23 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: This is old text. The new one is about sharing the interface, which was only allowed if you used multus.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 24 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: well ...

I want to have this close to what CNI does:
- CNI configuration
- CNI runtime params
- CNI arguments

Hence the config.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 25 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: this was removed

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 28 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: I've explained why it is there.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 30 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md]: this is a good point for this particular point, however, you need capabilities + runtime config for pinning the MAC on macvlan. 

I can simplify the examples though.

File: enhancements/controller-provisioned-underlay-interfaces.md (hunk 31 above)
Comment (maiqueb) [reply enhancements/controller-provisioned-underlay-interfaces.md:255]: maybe.

File: enhancements/controller-provisioned-underlay-interfaces.md [hunk 33]
```diff
@@ -0,0 +1,677 @@
...
+`Interfaces []UnderlayInterface` — a discriminated-union slice whose `type`
+field selects how each underlay link is obtained. It defines the first mode
+(`NetworkDevice`); this enhancement adds `CNI`.
+
+After this enhancement, the two modes are:
+
+| Mode | Behavior | Host NIC availability |
+|------|----------|----------------------|
+| `NetworkDevice` | Moves an existing host device into the router netns | Device is exclusively owned by the router |
+| `CNI` (this enhancement) | Invokes a CNI plugin to provision an interface in the router netns | Depends on the plugin (e.g. macvlan keeps parent on host) |
```
fedepaol: CNIProvided?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,677 @@
...
+
+##### CNI with macvlan
+
+Replaces Multus-based underlay — the parent NIC stays on the host.
+
+```yaml
+interfaces:
+  - type: CNI
+    cni:
+      config:
```
fedepaol: what's the point of the config field? Feels unnecessary nested to me

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: I would go even one step above and attach cni to the config content

cni:
   type: Raw
   config: xxxx

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: I am not following. Can I have two examples on how it would look like with static / dhcp - CNI provided and with the ipam bits provided by openperouter?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,677 @@
...
+              name: macvlan-dhcp
+              plugins:
+                - type: macvlan
+                  master: eth1
+                  mode: bridge
+                  capabilities:
+                    mac: true
+                  ipam:
+                    type: dhcp
+        runtimeConfig:
```
fedepaol, line 302: so runtimeConfig is when openperouter takes care of things?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,677 @@
...
+          mac: "02:42:c0:a8:01:0a"
```
fedepaol, line 303: does this mean each interface provided by this underlay will get the same mac?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,677 @@
...
+Repeat for each node with a different `nodeSelector` and MAC.
+
+> **DHCP daemon requirement:** The CNI `dhcp` IPAM plugin requires a DHCP
+> daemon process running on each node (`dhcp daemon` or via systemd socket
+> activation at `/run/cni/dhcp.sock`). See
+> [CNI DHCP plugin docs](https://www.cni.dev/plugins/current/ipam/dhcp/) for
+> setup.
+>
+> **Lease lifecycle across restarts:**
+> - **Router pod restart (netns preserved):** The CNI cache survives on the
```
fedepaol: this is (still) not true for systemd mode

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,677 @@
...
+The controller's reconciliation pipeline is extended to handle the `CNI` type.
+The provisioning logic runs in the same phase where it currently moves host
+devices.
+
+#### CNI Mode
+
+The controller invokes CNI ADD / CHECK / DEL via `libcni` as part of the
+underlay reconciliation mechanism:
+
+- **CNI ADD** — provisions the interface in the router netns. The controller
```
fedepaol: I think the cni configuration of underlay must be immutable

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,677 @@
...
+  resolves (and validates) the config source, parses the config, merges
+  `runtimeConfig` capabilities, and calls `AddNetworkList`.
+- **CNI CHECK** — validates that a previously-provisioned interface is still
+  correctly configured. The controller uses the CNI cache together with CNI
```
fedepaol: do we envision periodic check?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: no. it's pure CNI spec. it's a way for you to pass runtime parameters to plugins. 

for instance, say to macvlan interfaces (whose config is all the same) which MAC they'll use.

Or which IP you'll use in static IPs.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: we could. How would it work though ? every couple of secs / minutes check the underlay is properly configured ?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I can ...

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,677 @@
...
+            plugins:
+              - type: macvlan
+                master: eth1
+                mode: bridge
+                ipam:
+                  type: host-local
+                  subnet: "192.168.1.0/24"
+                  rangeStart: "192.168.1.10"
+                  rangeEnd: "192.168.1.50"
+```
```
maiqueb: host local is a notoriously bad example ... use something else.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: yes, but, as you can see, this underlay is pinned to a particular node. 

We just want to pin the MAC on the interface so the DHCP lease renewal is done as it should. Otherwise, if the router pod dies, and a new macvlan is created, it'll get a new MAC, which'll get a new IP, which'll invalidate the BGP session.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: could you elaborate ? ... I'm ignorant on the systemd front ...

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: It would surely simplify somethings.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I want our API to resemble the CNI spec, which'll be familiar to users.

Thus, we have:
- config => where you put the plugin configuration
- runtimeConfig => where you put workload specificc parameters (static IP, MAC for that interface, etc)
- CNI params => interface name, etc.

maiqueb (commented):
Replying to both reviews.

enhancements/controller-provisioned-underlay-interfaces.md
qinqon: ok missunderstood

enhancements/controller-provisioned-underlay-interfaces.md
qinqon: disregard this.

enhancements/controller-provisioned-underlay-interfaces.md
qinqon: If the answer is mayber, we do not right now.

enhancements/controller-provisioned-underlay-interfaces.md
qinqon: Well this is the enhancement to decide where do we put this new API, the runtimeConfig stuff haven't being decidd yet on this enhancement.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: it's reference configurations for each use case. 

I think it's valid to add this documentation.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: DOne.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I'd rather have `Raw`. At best, I can compromise and remove the `config` suffix.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I just noticed I am wrong though. We don't want to have nested "config -> rawConfig -> config". 

I'll remove the inner "config", which was unintended.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I removed that altogether from this enhancement (openperouter provided IPAM). 

All we have in the part in the "Future Work" section. 

Do you want me to re-introduce that ?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: it's a matter of not implementing it / implementing it when the need arises. I didn't explain much (... close to nothing) about it, and it's just a couple of examples.

Do you think this is harmful or confusing ?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Making openperouter generate the mac addresses from a prefix and a node index is far from "where do we put this new API". It is clearly a new feature. Hence my request to stay focused on the problem we're solving.

@qinqon

qinqon: Thing is, forcing users to configure one Underlay per node only because we want to use DHCP in a reliable manner may be a little bad UX, but I understand your point.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Done.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Removed the inner config. Now it stutters less.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Replaced w/ a static IPs example.

enhancements/controller-provisioned-underlay-interfaces.md
qinqon: I though we were going to have just RawConfig maybe we should remove the enum discrimination and just put the content here like 
```yaml
cni:
  config: 
    cniVersion: ...
     ....
```

enhancements/controller-provisioned-underlay-interfaces.md
qinqon: is just that it has to be clear that this is a future feature to illustrate the use of extra fields at the "config" level

qinqon (changes requested):
Since we are going to implement just "Raw" option, let's remove the "type" and put all the content under "config",  the alternative to using CNI was implementing the netlink calls ourselves and in that case we were going to put the config directly no "path" or "api" option for it, let's do the same for this:

```yaml
interfaces:
  - type: CNI
    cni:
      config:
        cniVersion: "1.0.0"
        name: ipvlan-underlay
        plugins:
          - type: ipvlan
            master: eth1
            mode: l2
```

maiqueb:
> Since we are going to implement just "Raw" option, let's remove the "type" and put all the content under "config", the alternative to using CNI was implementing the netlink calls ourselves and in that case we were going to put the config directly no "path" or "api" option for it, let's do the same for this:
> 
> ```yaml
> interfaces:
>   - type: CNI
>     cni:
>       config:
>         cniVersion: "1.0.0"
>         name: ipvlan-underlay
>         plugins:
>           - type: ipvlan
>             master: eth1
>             mode: l2
> ```

I'd rather not, and I think we should still have the discriminated union in case we want to introduce alternative means of configuration. 

Path / API are just things I thought of which integrate neatly with what we have, but we might find other alternatives (or customers might request integrations for) that would require us to expand the API. 

Having the discriminated union in place introduces a lot of flexibility in the API - even if we're just implementing a single configuration option.

qinqon:
> > Since we are going to implement just "Raw" option, let's remove the "type" and put all the content under "config", the alternative to using CNI was implementing the netlink calls ourselves and in that case we were going to put the config directly no "path" or "api" option for it, let's do the same for this:
> > ```yaml
> > interfaces:
> >   - type: CNI
> >     cni:
> >       config:
> >         cniVersion: "1.0.0"
> >         name: ipvlan-underlay
> >         plugins:
> >           - type: ipvlan
> >             master: eth1
> >             mode: l2
> > ```
> 
> I'd rather not, and I think we should still have the discriminated union in case we want to introduce alternative means of configuration.
> 
> Path / API are just things I thought of which integrate neatly with what we have, but we might find other alternatives (or customers might request integrations for) that would require us to expand the API.
> 
> Having the discriminated union in place introduces a lot of flexibility in the API - even if we're just implementing a single configuration option.

Ok, fair enough.

LGTM

qinqon (approved):
LGTM

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: but..?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+| Mode | Behavior | Host NIC availability |
+|------|----------|----------------------|
+| `NetworkDevice` | Moves an existing host device into the router netns | Device is exclusively owned by the router |
+| `CNI` (this enhancement) | Invokes a CNI plugin to provision an interface in the router netns | Depends on the plugin (e.g. macvlan keeps parent on host) |
+
+### Supported CNI Plugins
+
+Any CNI plugin that operates solely on the network namespace — without
+requiring extra host paths mounted into the router pod — works out of the
+box. Plugins that depend on external sockets or host directories (e.g.
```
fedepaol, line 125: do we foresee blacklisting those non supported cnis?

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: sorry but I still don't follow. We have 3 levels of nesting: cni - config- rawconfig and we are dumping a big json in rawconfig.

We don't have other options rather than rawConfig right? If so, can't we just have cni - rawConfig? What am I missing?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+      - "100.65.0.0/24"
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+Repeat for each node with a different `nodeSelector` and IP — only the
+`addresses` entry differs.
+
+The `runtimeConfig` field is opaque JSON — the controller passes it as
```
fedepaol: what does this refer to? We don't have runtimeConfig here in the example

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: thanks

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: thanks, now it's clear

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+    cidrs:
+      - "100.65.0.0/24"
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+Repeat for each node with a different `nodeSelector` and MAC.
+
+> **Capability requirement:** `libcni` performs capabilities filtering — it
```
fedepaol: iiuc, this is something that we get for free from libcni, so not sure what this section is about as it refers to a generic cni behavior. Mind adding a proper section title to this content?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+|------|------------|
+| CNI plugin binaries not installed on host | Clear error at reconcile time with plugin name and search path. Host/systemd package bundles common plugins. |
+| CNI cache lost but interface exists | Group ID 4242 detection still identifies the interface. IPAM lease may leak — document as a known edge case. |
+| CNI ADD is not idempotent (calling twice fails) | Controller checks `GetNetworkListCachedResult()` before `AddNetworkList()`. If cache is lost, existing interface detected via group ID. |
+| CNI DEL fails during teardown | DEL errors are logged but do not block teardown. CNI spec mandates plugins handle repeated DEL calls gracefully. |
+| `runtimeConfig` keys silently stripped by `libcni` | `libcni` only forwards keys the plugin declares in its `"capabilities"` block. Document prominently; consider logging a warning when `runtimeConfig` is set but the config has no capabilities. |
+| DHCP IPAM: macvlan random MAC causes IP instability | Document that DHCP with macvlan requires MAC pinning via `runtimeConfig.mac`. |
+| DHCP IPAM: DHCP daemon not running | CNI ADD fails immediately with a socket connection error. Document the daemon requirement. |
```
fedepaol, line 373: I general I agree we shouldn't do more validation than what CNIs do.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+	// netns. When IPAM is configured, the controller assigns deterministic
+	// per-node IPs from CIDR pools.
+	// +optional
+	NetworkDevice *NetworkDeviceConfig `json:"networkDevice,omitempty"`
+
+	// cni invokes a CNI plugin to provision an interface in the router
+	// netns. Configuration is sourced from the config discriminated union.
+	// IPAM is handled by the CNI plugin.
+	// +optional
+	CNI *CNIDevice `json:"cni,omitempty"`
```
fedepaol: can we have CNIDevice - CNIDeviceConfig , simmetric to NetworkDevice above?

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: we should always (and I tend to forget) the amount of automation our users have at their disposal to configure this stuff. So, what seems a bad ux for us is something easily solvable with ansible.

In general, I like not to overcrowd this with the openperouter's poor man ipam bits.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+	// Works identically regardless of config source variant.
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Type=object
+	// +optional
+	RuntimeConfig *apiextensionsv1.JSON `json:"runtimeConfig,omitempty"`
+}
+
+// CNIConfigSource selects where the CNI plugin configuration (config
+// JSON) comes from. Currently only RawConfig is supported; the union
+// can be extended with additional source variants (e.g. NAD reference,
```
fedepaol: do we really foresee this happening?

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: I guess? I am asking because it's documented here, or let me rephrase: when do we expect check to be called?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+- Startup validation: embedded JSON parse, plugin binary existence.
+- Integration tests: CNI ADD/DEL, runtimeConfig/capability filtering,
+  idempotency (cache hit), dual-stack IP extraction, error paths.
+
+#### Phase 2: Day 0 configuration (GA)
+
+Delivers the out-of-box experience for host/systemd deployments: install the
+package, write a config, start the service.
+
+- Host-mode package (RPM/deb/tarball) bundles statically-linked CNI plugin
```
fedepaol: I'd have this part of the first phase. Delivering CNIs is a day 1 problem too (unless we require to have them deployed - somehow) and I'd like to have a common solution, just because it will make day 1 and day 0 more similar.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+  binaries: `macvlan`, `ipvlan`, `static`, `dhcp`.
+- Installation path: `/opt/openperouter/cni/bin/` (plugins).
+- Controller's `CNI_PATH` includes `/opt/openperouter/cni/bin/` by default.
+- `--cni-bin-dir` flag for custom plugin paths.
+- Migration guide published in release notes.
+- E2E tests: static IPs, DHCP+MAC pinning, persistence across restart.
```
fedepaol: do we envision adding e2e test after merging?

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,667 @@
...
+- Production validation across multiple environments and NIC vendors.
+- Mode selection guidance documentation.
+
+## Drawbacks
+
+- **Slightly more verbose YAML.** The nested sub-struct adds indentation
+  compared to `nics: [eth1]`. This is the trade-off for type safety.
+- **Operational dependency on CNI plugin binaries.** Most Kubernetes clusters
```
fedepaol: given it's us invoking the CNIs, I think we can sort this out and openperouter delivering them

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: could be. mostly, I just want to have the discriminated union because I don't know how this'll grow. 

And, FWIW, I do see some value on the API (point at an existing NAD) for this - it allows you to re-use the NAD currently used for configuring the workload attachment. Without it, the user will have to duplicate this configuration.

FWIW, yes, I'm doing PATH just to integrate directly with the CNI ecosystem via file based. 

But probably pointing it to a config map would be preferable - I honestly don't know. Still I'd like to keep that option on the table if the need ever arises.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: It's more verbose, and I don't get why it's better, but I can blindly comply is what I didn't say. 

Why is it better ?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: We could. Is that a design decision ?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Let me try to wrap my head around having those 3 configs under CNI. 

Both you and quique are asking for the same, probably I'm missing something.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I'll move this to the appropriate place. The examples changes, and I failed to keep the text consistent.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: OK, nothing to do then, right ?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Sure we can.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I'll keep it as is then.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: at least whenever we reconcile the underlay, and we see the CNI cache is still there. We should issue check to see if the PLUGIN actually says everything's proper.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: no, this was AI slop gone wrong here. 

I'll fix it.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: still it's a drawback, right ? more stuff to do.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: let's keep a single thread to discuss the nesting level please. 

I'll resolve this one.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: resolving.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Let me try to understand then... All that's in phase 2 would go into phase 1 ?... 

Just saying that both "Installation path: `/opt/openperouter/cni/bin/` (plugins).", "Controller's `CNI_PATH` includes `/opt/openperouter/cni/bin/` by default." and "`--cni-bin-dir` flag for custom plugin paths." are part of that. 

What's the point of having a phased implementation then ?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Done.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Done.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I can go for:
```
  - type: CNI
    cni:
      type: RawConfig
      rawConfig:
        cniVersion: "1.0.0"
        name: macvlan-underlay
        plugins:
          - type: macvlan
            master: eth1
            mode: bridge
            ipam:
              type: static
              addresses:
                - address: "192.168.1.10/24"
```
or even take away the CNI.Type attribute, which will prevent us from introducing other types if the need presents itself. 

Your call @fedepaol . I know @qinqon wants to go for simple and just have raw config.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Added a line about this - let's validate, and only allow a subset.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Done.

maiqueb (commented):
@fedepaol please take another look

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,671 @@
...
+
+As an operator on a network that limits the number of MAC addresses per
+port, I want the router's underlay interface to share the physical NIC's
+MAC address, so that only one MAC appears on the wire.
+
+#### Story 7: DHCP-Managed Networks
+
+As an operator whose network uses DHCP for IP management, I want the
+router's underlay interface to obtain its IP via DHCP from the existing
+infrastructure, with a stable identity across restarts.
```
fedepaol, line 103: Can we also add a story about bonds? IIUC, today it's not possible to move a bond so the bond cni might be useful here

maiqueb: sure I guess ... should it also be added to the list of allowed CNIs ?

fedepaol: yep

maiqueb: also added an example with static IPAM.

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: well it's more about the preservation of the netns (and the lack of it in systemd mode)

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: yep, it was just a spare comment :-)

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: ok, makes sense

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: right, it was a more general consideration on maybe it's better to always ship them - and I think the decision is not clear from this proposal.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: so what do you want me to do about systemd mode ? throw a disclaimer this is only true for k8s api mode ? 
@fedepaol

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: seems a good compromise to me. If the need for type arises, we can still add it later I guess.

maiqueb: OK. let's just contemplate the fact this means that we'll break the API when if/when we add the "type" attribute.

Will update.

fedepaol: how does it break the api? It won't as long as we default to raw if we add it

maiqueb: that's not how a discriminated union works. You *have* to specify the type.

maiqueb: I'll still use the type unless someone objects hard enough.

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: nah, let's assume we'll have it thanks to yet another never ending pr I am crafting for you :P

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: can you suggest a proper title ? CNI specifics ? Enabling runtime parameters ? ...

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I'm OK with *always* shipping them. 

The question now becomes: do we drop the phases thing ? Or do we do this in phase 2 ?

fedepaol: if we always ship them, and we don't defer systemd mode, then we must ship them from the get go

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: right, doesnt' make much sense imo and shipping cnis is less of a hassle imo

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: I don't know what you meant by dumping this here :-) , but it really feels dumped and it's odd because it doesn't have much to do with what comes before / after, unless it does and I did not understand.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: we can even throw a separate PR only for shipping those in a sidecar, which would be a pre-req of the implementation work.

maiqueb: Done.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Done. Added another section.

enhancements/controller-provisioned-underlay-interfaces.md
qinqon: You can resolve this.

qinqon:
@maiqueb about the cni.Type discrimination union I think we can start without "type" and just have "config" with the content, later on if a need for path or api pops up we can add the "typ", "api" and "path" fields, meaning that we can start just with "config" without "type"

maiqueb:
> @maiqueb about the cni.Type discrimination union I think we can start without "type" and just have "config" with the content, later on if a need for path or api pops up we can add the "typ", "api" and "path" fields, meaning that we can start just with "config" without "type"

I'd rather not, because that would either break backwards compatibility (`type` is required) or force us to default to `RawConfig` to preserve it, which breaks how a discriminated union works. 

Let's proceed with type please.

qinqon:
> > @maiqueb about the cni.Type discrimination union I think we can start without "type" and just have "config" with the content, later on if a need for path or api pops up we can add the "typ", "api" and "path" fields, meaning that we can start just with "config" without "type"
> 
> I'd rather not, because that would either break backwards compatibility (`type` is required) or force us to default to `RawConfig` to preserve it, which breaks how a discriminated union works.
> 
> Let's proceed with type please.

Ok, I will no bring this up again, we can proceed.

enhancements/controller-provisioned-underlay-interfaces.md
```diff
@@ -0,0 +1,704 @@
...
+	// networkDevice moves an existing host network device into the router
+	// netns. When IPAM is configured, the controller assigns deterministic
+	// per-node IPs from CIDR pools.
+	// +optional
+	NetworkDevice *NetworkDeviceConfig `json:"networkDevice,omitempty"`
+
+	// cni invokes a CNI plugin to provision an interface in the router
+	// netns. IPAM is handled by the CNI plugin.
+	// +optional
+	CNIDevice *CNIDeviceConfig `json:"cni,omitempty"`
```
fedepaol: can we call the json field cniDevice too?

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol: resolving as we went for CNIDevice I think?

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: Done.

maiqueb:
Also moved the bond section to future work, since it isn't an urgent use case - we should tackle it as a follow-up to this enhancement.

fedepaol:
thanks!

Issue #328, opened 2026-04-21
Flaky e2e: OVS dual-stack test fails due to CRI pod sandbox race condition


qinqon:
## Summary

The `e2etests (operator)` job intermittently fails on the test `OVS bridge autocreate for dual stack` (`evpn_l2.go:205`) due to a race condition during controller startup.

## Root cause

When the controller pod starts, it calls `ListPodSandbox` via CRI to resolve the router pod's network namespace. If the CRI sandbox isn't ready yet, the call returns 0 items and the controller fails immediately with:

```
ListPodSandbox returned 0 item for pod <uid>
```

This cascades into:
1. `failed to retrieve target namespace` errors on all reconcile attempts
2. `dial unix /etc/frr/reload.sock: connect: no such file or directory` — FRR socket not available
3. `no underlays configured` — reconciler can't process the underlay
4. OVS dual-stack test times out waiting for data plane convergence

The controller-runtime requeue does eventually recover, but the OVS test has a tight timing window and times out before convergence.

## Evidence

- PR #313 run: https://github.com/openperouter/openperouter/actions/runs/24673653242/job/72152867424
- 80/81 tests pass in the operator job; only the OVS dual-stack test fails
- All other e2e jobs (manifests, helm, systemdmode, hostmode-boot) pass

## Relevant code

- `internal/pods/pods.go:112-113` — `podSandboxID()` returns error immediately when sandbox not found, no retry

fedepaol:
Was this preprocessed? Do you agree with the suggested fix?

maiqueb:
you say:
> The controller-runtime requeue does eventually recover, but the OVS test has a tight timing window and times out before convergence.

What are the tight timing window requirements ? maybe we can adjust the test, rather than add production code. 
FWIW, the suggested fix sounds plausible to me.

fedepaol:
> you say:
> 
> > The controller-runtime requeue does eventually recover, but the OVS test has a tight timing window and times out before convergence.
> 
> What are the tight timing window requirements ? maybe we can adjust the test, rather than add production code. FWIW, the suggested fix sounds plausible to me.

We already have the reconcile loop that retries, I think we should leverage that instead of adding backoffs around.

I would try to understand if we are running the tests too early (and if this was the first test where the controller was ready and the router wasn't), or if our logic in checking if the router pod is available is not effective. Also, all the tests expect to run when the router pods are ready, so there's still something odd here.

qinqon:
[...]

PR #332, opened 2026-04-23
webhook(l2vni): reject l2gatewayip on disconnected L2VNI

RamLavi:
/kind bug

Add validation in `ValidateL2VNIs` to reject L2VNIs that have l2gatewayips set without a VRF. Also adds an `IsDisconnected()` helper to the L2VNI API type.

A disconnected L2VNI (no spec.vrf) provides east-west L2 forwarding only.
Setting l2gatewayips without a VRF creates unreachable gateway IPs since there is no L3VNI to route traffic.

Fixes: #280

**Release note**:

```release-note
reject l2gatewayip on disconnected L2VNI
```

### DIFF
--- a/internal/conversion/validate_vni.go
+++ b/internal/conversion/validate_vni.go
@@ -80,6 +80,9 @@ func ValidateL2VNIs(l2Vnis []v1alpha1.L2VNI) error {
 				return err
 			}
 		}
+		if len(vni.Spec.L2GatewayIPs) > 0 && !hasVRF(vni) {
+			return fmt.Errorf("l2gatewayips cannot be set without spec.vrf for vni %q", vni.Name)
+		}
 		if len(vni.Spec.L2GatewayIPs) > 0 {
 			_, err := ipfamily.ForCIDRStrings(vni.Spec.L2GatewayIPs...)
 			if err != nil {
@@ -238,6 +241,10 @@ func cidrsOverlap(cidr1, cidr2 string) (bool, error) {
 	}
 
 	return false, nil
+}
+
+func hasVRF(l2vni v1alpha1.L2VNI) bool {
+	return l2vni.Spec.VRF != nil && *l2vni.Spec.VRF != ""
 }
 
 func isValidInterfaceName(name string) error {
--- a/e2etests/tests/evpn_l2.go
+++ b/e2etests/tests/evpn_l2.go
@@ -382,7 +382,7 @@ var _ = Describe("Routes between bgp and the fabric - vtepInterface", func() {
 		redistributeConnectedForLeafKind(nodes)
 
 		l2VniRedWithGateway := l2VniRed.DeepCopy()
-		l2VniRedWithGateway.Spec.VRF = nil
+		l2VniRedWithGateway.Spec.VRF = ptr.To("red")
 		l2VniRedWithGateway.Spec.L2GatewayIPs = []string{"192.171.24.1/24"}
 		l2VniRedWithGateway.Spec.HostMaster = &v1alpha1.HostMaster{
 			Type: linuxBridgeHostAttachment,
--- a/e2etests/tests/hostconfiguration.go
+++ b/e2etests/tests/hostconfiguration.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/utils/ptr"
 )
 
 var ValidatorPath string
@@ -727,6 +728,7 @@ var _ = ginkgo.Describe("Router Host configuration", func() {
 			},
 			Spec: v1alpha1.L2VNISpec{
 				VNI:          400,
+				VRF:          ptr.To("second"),
 				L2GatewayIPs: []string{"192.168.1.4/24"},
 			},
 		}
--- a/e2etests/tests/webhooks.go
+++ b/e2etests/tests/webhooks.go
@@ -229,6 +229,17 @@ var _ = Describe("Webhooks", func() {
 					VXLanPort: 4789,
 				},
 			}}, "duplicate vni"),
+			Entry("when trying to create an L2VNI with l2gatewayips but no VRF", []v1alpha1.L2VNI{{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "l2vni-no-vrf-gw",
+					Namespace: openperouter.Namespace,
+				},
+				Spec: v1alpha1.L2VNISpec{
+					VNI:          213,
+					VXLanPort:    4789,
+					L2GatewayIPs: []string{"10.100.0.1/24"},
+				},
+			}}, "l2gatewayips cannot be set without spec.vrf"),
 			Entry("when trying to create an L2VNI with an invalid IPv4 address", []v1alpha1.L2VNI{{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "l2-invalid-ip4",
@@ -236,6 +247,7 @@ var _ = Describe("Webhooks", func() {
 				},
 				Spec: v1alpha1.L2VNISpec{
 					VNI:          201,
+					VRF:          ptr.To("test-vrf-2"),
 					VXLanPort:    4789,
 					L2GatewayIPs: []string{"not-an-ip-address"},
 				},
@@ -247,6 +259,7 @@ var _ = Describe("Webhooks", func() {
 				},
 				Spec: v1alpha1.L2VNISpec{
 					VNI:          202,
+					VRF:          ptr.To("test-vrf-2"),
 					VXLanPort:    4789,
 					L2GatewayIPs: []string{"256.256.256.256/24"},
 				},
@@ -258,6 +271,7 @@ var _ = Describe("Webhooks", func() {
 				},
 				Spec: v1alpha1.L2VNISpec{
 					VNI:          203,
+					VRF:          ptr.To("test-vrf-2"),
 					VXLanPort:    4789,
 					L2GatewayIPs: []string{"192.168.1.1/24", "invalid-ip"},
 				},
@@ -269,6 +283,7 @@ var _ = Describe("Webhooks", func() {
 				},
 				Spec: v1alpha1.L2VNISpec{
 					VNI:          204,
+					VRF:          ptr.To("test-vrf-2"),
 					VXLanPort:    4789,
 					L2GatewayIPs: []string{"192.168.1.1/24", "2001:db8::1/64", "10.0.0.1/24"},
 				},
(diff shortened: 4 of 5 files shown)


internal/webhooks/l2vni_webhook.go
```diff
@@ -101,6 +101,11 @@ func validateL2VNIDelete(_ *v1alpha1.L2VNI) error {
 }
 
 func validateL2VNI(l2vni *v1alpha1.L2VNI) error {
+	vrfNotSet := l2vni.Spec.VRF == nil || *l2vni.Spec.VRF == ""
+	if vrfNotSet && l2vni.Spec.L2GatewayIP != "" {
+		return fmt.Errorf("l2gatewayip cannot be set without spec.vrf: a disconnected L2VNI (no VRF) cannot act as a gateway")
+	}
```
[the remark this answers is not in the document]
RamLavi: fixed

RamLavi:
Ran it locally:
```
[root@zeus22 openperouter]# cat <<'EOF' | kubectl apply -f -
kind: L2VNI
metadata:
  name: bad-disconnected
spec:
  vni: 999
  vxlanport: 4789
  l2gatewayip: "10.100.0.1/24"
EOF
Error from server (Forbidden): error when creating "STDIN": admission webhook "l2vnivalidationwebhook.openperouter.io" denied the request: l2gatewayip cannot be set without spec.vrf
[root@zeus22 openperouter]#
[root@zeus22 openperouter]# cat <<'EOF' | kubectl apply -f -
kind: L2VNI
metadata:
  name: good-with-vrf
spec:
  vni: 998
  vxlanport: 4789
  vrf: "my-vrf"
  l2gatewayip: "10.100.0.1/24"
EOF
l2vni.openpe.openperouter.github.io/good-with-vrf created
[root@zeus22 openperouter]#
[root@zeus22 openperouter]# cat <<'EOF' | kubectl apply -f -
kind: L2VNI
metadata:
  name: good-disconn 
spec:
  vni: 997
  vxlanport: 4789
EOF
l2vni.openpe.openperouter.github.io/good-disconn created
```

maiqueb (approved):
tks, this makes clearer what a L2VNI without VRF is.

maiqueb:
hm, seems you'll need to adjust more tests ... :)

internal/webhooks/l2vni_webhook.go
```diff
@@ -102,6 +102,11 @@ func validateL2VNIDelete(_ *v1alpha1.L2VNI) error {
 }
 
 func validateL2VNI(l2vni *v1alpha1.L2VNI) error {
+	vrfNotSet := l2vni.Spec.VRF == nil || *l2vni.Spec.VRF == ""
```
fedepaol: the check must go where we validate the vnis. By doing this we'll be able to leverage teh logic both in reconcile and in webhooks. Webhooks must be dumb and leverage what we have under https://github.com/openperouter/openperouter/blob/main/internal/conversion/validate_vni.go

fedepaol (changes requested):

internal/webhooks/l2vni_webhook.go
RamLavi: DONE

RamLavi:
[change](https://github.com/openperouter/openperouter/compare/d633b6ec868d02ee28f66facbc06a42917014866..24a777fe6619d382328a4369da63d80878924318): Address @maiqueb and @fedepaols' reviews

maiqueb:
@RamLavi I think you'll also need to relax the API definition: https://github.com/openperouter/openperouter/blob/1af0c2dd43fbd0f199ea9c540d6504f354357cbc/api/v1alpha1/l2vni_types.go#L144

AFAIU, this forces a VRF onto ALL the L2VNIs. We don't want that.

RamLavi:
> @RamLavi I think you'll also need to relax the API definition:
> 
> https://github.com/openperouter/openperouter/blob/1af0c2dd43fbd0f199ea9c540d6504f354357cbc/api/v1alpha1/l2vni_types.go#L144
> 
> AFAIU, this forces a VRF onto ALL the L2VNIs. We don't want that.

@maiqueb you mean in a separate followup PR right?

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/a4a898f9407c5ee5eb96ed601131de436fef81bf..6395b298be37a493f0105a9da4e24a4a94e079e0): removed the second commit, I mistakenly thought the issue came from the webhook.

maiqueb:
> > @RamLavi I think you'll also need to relax the API definition:
> > https://github.com/openperouter/openperouter/blob/1af0c2dd43fbd0f199ea9c540d6504f354357cbc/api/v1alpha1/l2vni_types.go#L144
> > 
> > AFAIU, this forces a VRF onto ALL the L2VNIs. We don't want that.
> 
> @maiqueb you mean in a separate followup PR right?

IDK if this'll work unless you relax that.

RamLavi:
> > > @RamLavi I think you'll also need to relax the API definition:
> > > https://github.com/openperouter/openperouter/blob/1af0c2dd43fbd0f199ea9c540d6504f354357cbc/api/v1alpha1/l2vni_types.go#L144
> > > 
> > > AFAIU, this forces a VRF onto ALL the L2VNIs. We don't want that.
> > 
> > 
> > @maiqueb you mean in a separate followup PR right?
> 
> IDK if this'll work unless you relax that.

I'm not sure it's related. Can you explain?

RamLavi:
FYI The only CI run [failing](https://github.com/openperouter/openperouter/actions/runs/25013500207/job/73337676716?pr=332) is due to a known flake https://github.com/openperouter/openperouter/issues/328

@qinqon can you take a look?

maiqueb:
> > > > @RamLavi I think you'll also need to relax the API definition:
> > > > https://github.com/openperouter/openperouter/blob/1af0c2dd43fbd0f199ea9c540d6504f354357cbc/api/v1alpha1/l2vni_types.go#L144
> > > > 
> > > > AFAIU, this forces a VRF onto ALL the L2VNIs. We don't want that.
> > > 
> > > 
> > > @maiqueb you mean in a separate followup PR right?
> > 
> > 
> > IDK if this'll work unless you relax that.
> 
> I'm not sure it's related. Can you explain?

you're trying to make it clearer what an L2VNI without a VRF is right ? a disconnected overlay, without any routing involved. As part of that, you're making it clear that an L2VNI without a VRF cannot have gateway IPs. 

I'm saying the VRF name method of an L2VNI will return the metadata name of the CR whenever the L2VNI doesn't feature a vrf name, which makes it very confusing for the reader. It might work now, but, the model is not right. But as you say, this might be off scope.

internal/conversion/validate_vni.go
```diff
@@ -80,6 +80,9 @@ func ValidateL2VNIs(l2Vnis []v1alpha1.L2VNI) error {
 				return err
 			}
 		}
+		if (vni.Spec.VRF == nil || *vni.Spec.VRF == "") && len(vni.Spec.L2GatewayIPs) > 0 {
```
qinqon: nit: Feels more natural to check how we print the error

```golang
len(vni.Spec.L2GatewayIPs) > 0 && (vni.Spec.VRF == nil || *vni.Spec.VRF == "")
```

qinqon (changes requested):
Just a nit

internal/conversion/validate_vni.go
maiqueb: tbh I'd rather have a well named helper to sort this out over the parenthesis and `&&` and `||` . 

It also aligns better to what we want to have in the repo: https://github.com/openperouter/openperouter/blob/main/.gemini/styleguide.md#flat-control-flow--early-returns-over-nested-ifs

qinqon: make sense

RamLavi: DONE

RamLavi:
> > > > > @RamLavi I think you'll also need to relax the API definition:
> > > > > https://github.com/openperouter/openperouter/blob/1af0c2dd43fbd0f199ea9c540d6504f354357cbc/api/v1alpha1/l2vni_types.go#L144
> > > > > 
> > > > > AFAIU, this forces a VRF onto ALL the L2VNIs. We don't want that.
> > > > 
> > > > 
> > > > @maiqueb you mean in a separate followup PR right?
> > > 
> > > 
> > > IDK if this'll work unless you relax that.
> > 
> > 
> > I'm not sure it's related. Can you explain?
> 
> you're trying to make it clearer what an L2VNI without a VRF is right ? a disconnected overlay, without any routing involved. As part of that, you're making it clear that an L2VNI without a VRF cannot have gateway IPs.
> 
> I'm saying the VRF name method of an L2VNI will return the metadata name of the CR whenever the L2VNI doesn't feature a vrf name, which makes it very confusing for the reader. It might work now, but, the model is not right. But as you say, this might be off scope.

ACK, I'll do it in a follow up then

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/6395b298be37a493f0105a9da4e24a4a94e079e0..99afd36097b2ad4412a834960c897cef6a5bc2c7): Address @qinqon's and @maiqueb's nit.

api/v1alpha1/l2vni_types.go
```diff
@@ -135,6 +135,10 @@ type L2VNI struct {
 	Status L2VNIStatus `json:"status,omitempty"`
 }
 
+func (v L2VNI) IsDisconnected() bool {
```
qinqon: From api kubevirt conventions having helpers inside the /api package is not good thing, that's why they have apimachinery, if we are following that guideline this should be a normal free function., also we are just using it at one place it does not need to be public exposed

Also "disconnected" concept feels to generic, maybe just a free function like `l2vniHasVRF(vni)`

qinqon (changes requested):

api/v1alpha1/l2vni_types.go
RamLavi: Good point, will move it out of the API package. I'll make it a private function in the conversion package where it's used — something like hasVRF(l2vni). 

Note that in the follow-up PR (relaxing VRFName) it's used in 3 places, so having a named helper still makes sense.

RamLavi: DONE

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/99afd36097b2ad4412a834960c897cef6a5bc2c7..24b751a12cd1c023a10b22fee677c396633fabc4): address @qinqon's review

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/24b751a12cd1c023a10b22fee677c396633fabc4..21831536a66847ffce4a39c842ba680d29f5b91f): Rebase

qinqon (approved):
LGTM

maiqueb:
LGTM. Plan is to merge when CI shows no regressions.

PR #333, opened 2026-04-23
tests: improve error matching in webhook tests

maiqueb:

>
> /kind bug

/kind cleanup

> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

This PR uses the gomega matchers in a better way, by using a single idiomatic expectation instead of 2.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/tests/webhooks.go
+++ b/e2etests/tests/webhooks.go
@@ -61,8 +61,7 @@ var _ = Describe("Webhooks", func() {
 				err := Updater.Update(config.Resources{
 					L3VNIs: []v1alpha1.L3VNI{vni},
 				})
-				Expect(err).To(HaveOccurred())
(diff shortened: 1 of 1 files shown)


RamLavi (commented):
LGTM but I think the bot is right about the error message mismatch. Do you wanna add a commit (and then I'll lgtm) or do you want it in a separate PR (in which case I will lgtm now)?

maiqueb:
> LGTM but I think the bot is right about the error message mismatch. Do you wanna add a commit (and then I'll lgtm) or do you want it in a separate PR (in which case I will lgtm now)?

I have no clue how the test was passing before then. I haven't updated the error msg in the test ... 

Let's see what CI says first.

fedepaol:
thanks!

PR #335, opened 2026-04-23
design: Support for DPDK/grout

zeeke:
This enhancement adds github.com/DPDK/grout as an optional, DPDK-accelerated data plane that runs alongside FRR as a sidecar container.

>
> /kind bug
> /kind cleanup
> /kind feature
/kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

**Release note**:

```release-note
NONE
```

### DIFF
--- a/enhancements/grout-dataplane.md
+++ b/enhancements/grout-dataplane.md
@@ -1,0 +1,264 @@
+# Grout: DPDK-Accelerated Dataplane for OpenPERouter
+
+## Summary
+
+OpenPERouter currently relies on the Linux kernel's networking stack for data
+plane forwarding (VXLAN encap/decap, VRF routing, bridge learning). While
+correct and stable, kernel-based forwarding has inherent performance limits
+due to interrupt-driven packet processing and kernel/user-space transitions.
+
+This enhancement adds **grout** as an optional, DPDK-accelerated data plane
+that runs alongside FRR as a sidecar container ([GitHub](https://github.com/DPDK/grout)). 
+When enabled, grout replaces the kernel networking stack for packet forwarding while FRR continues to
+handle the control plane via its `dplane_grout` zebra [module](https://docs.frrouting.org/en/latest/basic.html#loadable-module-support).
+The integration is opt-in and does not affect the default kernel-based data path.
+
+## Motivation
+
+### Goals
+
+- **Higher forwarding throughput**: leverage DPDK poll-mode drivers to bypass
+  kernel overhead for VXLAN encapsulation, decapsulation, and routing.
+- **Opt-in activation**: grout is disabled by default. Operators enable it via
+  a Helm value (`openperouter.grout.enabled: true`) with no impact on existing
+  deployments.
+- **Reuse existing control plane**: FRR remains the routing daemon. Zebra's
+  `dplane_grout` module pushes forwarding entries to grout instead of the
+  kernel, requiring no changes to BGP, EVPN, or route exchange logic.
+- **Support underlay and passthrough flows**: the initial integration covers
+  underlay interface setup and L3 passthrough via grout ports. Following milestone
+  will cover all the kernel based scenarios, adding HW acceleration via SR-IOV
+  Virtual Functions plugged into workload pods. 
+
+### Non-Goals
+
+- Replacing the kernel-based data plane. Both modes coexist; the kernel path
+  remains the default.
+
+## Proposal
+
+### Overview
+
+Grout runs as a privileged sidecar container in the router DaemonSet pod. It
+exposes a UNIX socket (`/var/run/grout/grout.sock`) that serves two consumers:
+
+1. **FRR (zebra)**: uses the `dplane_grout` module to program forwarding
+   entries into grout instead of the kernel's netlink interface. The socket path
+   is injected via the `GROUT_SOCK_PATH` environment variable.
+2. **The controller**: uses the `grcli` CLI  to configure grout ports, addresses, 
+   vrfs and routes on the grout instance.
+
+When grout is enabled, the controller will take an alternative configuration path
+instead of `configureInterfaces()`, delegating interface setup to a
+dedicated grout package.
+
+
+### User Stories
+
+#### Story 1: High-Throughput PE Router
+
+As a network operator running OpenPERouter on nodes with DPDK-capable NICs, I
+want to enable grout so that VXLAN encap/decap and routing happen in user-space
+with poll-mode drivers, achieving higher packet-per-second throughput than the
+kernel data plane.
+
+
+## Design Details
+
+### Helm Configuration
+
+Grout is configured under `openperouter.grout` in the Helm values:
+
+```yaml
+openperouter:
+  grout:
+    enabled: true
+    image:
+      repository: quay.io/openperouter/router
+      tag: "main-grout"
+      pullPolicy: ""
+    resources:
+      requests:
+        memory: "512Mi"
+        cpu: "250m"
+      limits:
+        memory: "1Gi"
+        cpu: "500m"
+```
+
+When `grout.enabled` is `true`, the Helm templates add:
+
+- The `-M dplane_grout` module flag to FRR's zebra process
+- The `GROUT_SOCK_PATH` environment variable to the FRR container
+- A grout sidecar container in the router pod
+- A `grout-socket` hostPath volume at `/var/run/openperouter/grout`
+- `--grout-enabled=true` and `--grout-socket` flags to the controller
+
+### Controller Integration
+
+The controller receives two new CLI flags:
+
+- `--grout-enabled` (bool): switches the reconciler to the grout data path
+- `--grout-socket` (string, default `/var/run/grout/grout.sock`): path to the
+  grout UNIX socket
+
+When grout is enabled, the reconciler configures FRR as before. After that, the host network
+configuration is skipped in favor of a Grout configuration.
+
+#### Configuring Grout
+
+The following paragraphs show how the controller implements the PERouter logic by
+configuring the deployed grout instance.
+
+##### L3Passthrough
+
+The following actions are executed by the controller when the example Underlay + L3Passthrough
+configuration is in place.
+
+```
+apiVersion: ...
+kind: Underlay
+metadata:
+  name: underlay
+  namespace: openperouter-system
+spec:
+  asn: 64514
+  nics:
+    - eth3
+  neighbors:
+    - asn: 64512
+      address: 192.168.11.2
+---
+apiVersion: ...
+kind: L3Passthrough
+metadata:
+  name: passthrough
+  namespace: openperouter-system
+spec:
+  hostsession:
+    asn: 64514
+    hostasn: 64515
+    localcidr:
+      ipv4: 192.169.10.0/24
+```
+
+Underlay
+1. The controller moves `eth3` (from `Underlay.spec.nics[0]`) interface to the 
+   router’s namespace. This step is common to the kernel datapath.
+2. controller creates a grout port for the underlay, linking it to the eth3 nic:
+
+   `grcli interface add port underlay devargs net_tap0,remote=eth3,iface=tap_underlay`
+
+3. move all the ip addresses from eth3 to the grout underlay port
+  
+   `grcli address add <eth3_ip> iface underlay`
+
+4. set kernel route for all the moved ip addresses to use grout. `main` is the default vrf 
+   control plane interface. This is needed to make sure all the connections created by FRR
+   go through Grout
+  
+   `ip route add <eth3_net> dev main src <eth3 IP addr>`
+
+
+L3Passthrough
+1. The controller creates a pair grout_port/tap_interface (compared to the veth pair of kernel datapath):
+
+   `grcli interface add port pt-ns devargs net_tap1,iface=pt-host`
+
+   `grcli address add 192.169.10.1/31 iface pt-ns`
+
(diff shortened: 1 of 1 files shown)


enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,165 @@
...
+- grout runs in `--test-mode`, which means no huge pages are used at the moment.
+
+#### M2
+
+- L3VNI scenario implemented with grout.
+
+#### M3
+- L2VNI scenario implemented with grout.
+- Full e2e test coverage matching the kernel data path test suite.
+- Performance benchmarks comparing grout vs kernel forwarding throughput.
```
rjarry: Performance benchmarks should be performed with SR-IOV support.

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,165 @@
...
+  using grcli
+
+## Drawbacks
+
+- **Privileged container**: grout requires raw device access, increasing the
+  security surface of the router pod.
+- **Incomplete VNI support**: L2VNI and L3VNI are stubbed. Operators who need
+  VNI forwarding acceleration must wait for follow-up work.
+- **Hardware dependency**: DPDK poll-mode drivers require specific NIC hardware
+  and driver binding (vfio-pci, uio_pci_generic), limiting portability.
```
rjarry: Most NICs have a driver in DPDK in 2026. This should work everywhere.

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,165 @@
...
+- **Increased resource consumption**: grout's poll-mode operation consumes
+  dedicated CPU cores, unlike the interrupt-driven kernel path.
```
rjarry: I think this last point should be rephrased. When idle, grout does not actively poll Rx queues. It has a power saving mechanism involving micro-sleeps to allow CPUs to enter lower C-states.

Also, if we are comparing apples to apples, grout will have a much higher million-packets-per-second/Watts ratio than linux. The only downside is that you need to dedicate CPUs to it if you want to achieve zero packet loss. Linux does not allow dedicating CPUs for networking but also does not guarantee zero packet loss.

zeeke: rephrased, thanks

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,161 @@
...
+
+OpenPERouter currently relies on the Linux kernel's networking stack for data
+plane forwarding (VXLAN encap/decap, VRF routing, bridge learning). While
+correct and stable, kernel-based forwarding has inherent performance limits
+due to interrupt-driven packet processing and kernel/user-space transitions.
+
+This enhancement adds **grout** as an optional, DPDK-accelerated data plane
+that runs alongside FRR as a sidecar container ([GitHub](https://github.com/DPDK/grout)). 
+When enabled, grout replaces the kernel networking stack for packet forwarding while FRR continues to
+handle the control plane via its `dplane_grout` zebra (module)[https://docs.frrouting.org/en/latest/basic.html#loadable-module-support].
```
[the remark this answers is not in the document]
maiqueb: this one seems real.

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,172 @@
...
+## Design Details
+
+### Helm Configuration
+
+Grout is configured under `openperouter.grout` in the Helm values:
+
+```yaml
+openperouter:
+  grout:
+    enabled: false
```
maiqueb: nit: maybe set this to true. Doesn't make sense to specify the img of a component you don't want to use imho

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Controller Integration
+
+The controller receives two new CLI flags:
+
+- `--grout-enabled` (bool): switches the reconciler to the grout data path
+- `--grout-socket` (string, default `/var/run/grout/grout.sock`): path to the
+  grout UNIX socket
+
+When grout is enabled, the reconciler FRR as before. After that, the host network
```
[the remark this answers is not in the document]
maiqueb: this one is right

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,161 @@
...
+- E2E tests run with `--label-filter=passthrough` for the grout configuration
+- A set of `e2etests` lanes will be added to test against a grout kind/clab deployment.
+- As the features will be implemented incrementally, the `GINKGO_ARGS="--label-filter=..."`
+  variable will be used to have a stable CI signal.
+
+### Test Plan
+
+- **E2E tests**: All the tests under `e2etests/` are supposed to run as is, without
+  any specific change, on a grout based deployment.
+- **Grout diagnostics**: - `e2etests/tests/dump.go` needs to be extended to gather
```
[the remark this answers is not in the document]
maiqueb: please take this one into account

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,172 @@
...
+### Milestone
+
+#### M1
+
+- Grout sidecar deployed as opt-in via Helm values.
+- Underlay and passthrough setup implemented.
+- FRR zebra configured with `dplane_grout` module when grout is enabled.
+- E2E passthrough tests passing with grout-enabled configuration.
+- Controller issues `grcli` command to configure grout.
+- Grout interfaces are based on TAP devices. No hardware acceleration at this stage.
```
maiqueb, line 211: in which milestone will the grout interfaces be accelerated ? I assume means using vfio, rather than virtio, right ?

maiqueb (commented):
very shallow review.

enhancements/grout-dataplane.md
zeeke: correct. I like to show the proposed default values for the configuration, but the `enabled` field can be set to True to give sense to the document.

enhancements/grout-dataplane.md
zeeke: Milestone 4 in my view:
> #### M4
> - Support for hardware acceleration with SR-IOV NICs, CPU isolation and huge pages

maiqueb: OK. 

So we tackle underlay + L3 passthrough first. Then we extend that to L3VNI. Then, to L2VNI. Finally, we enable vfio, which will grant the "user" the benefits of accelerated hardware. 

I wonder if we shouldn't have different order though: first give the user the benefits of HW accelaration in milestone 2 for a limited subset of CRDs, and then, incorporate more CRDs into the "accelerated" path. 

Does this make sense in your opinion @zeeke ? Is it technically feasible ?

zeeke: It definitely makes sense to me. It let us explore problems with the HW acceleration early.
Also, L3VNI and L2VNI could be developed in parallel. The TAP based datapath should be implemented first because they are supposed to be a little "easier" to test, as the should work with the same test suite we have for kernel datapath.
Updating the doc

maiqueb: thank you. You can resolve this comment.

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,183 @@
...
+  kernel overhead for VXLAN encapsulation, decapsulation, and routing.
+- **Opt-in activation**: grout is disabled by default. Operators enable it via
+  a Helm value (`openperouter.grout.enabled: true`) with no impact on existing
+  deployments.
+- **Reuse existing control plane**: FRR remains the routing daemon. Zebra's
+  `dplane_grout` module pushes forwarding entries to grout instead of the
+  kernel, requiring no changes to BGP, EVPN, or route exchange logic.
+- **Support underlay and passthrough flows**: the initial integration covers
+  underlay interface setup and L3 passthrough via grout ports. Following milestone
+  will cover all the kernel based scenarios.
```
fedepaol: despite the fact that it will come later in the implementation, we should mention also integration with workloads using sr-iov secondary nics (which will be enabled by grout iiuc)

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,183 @@
...
+- Underlay and passthrough setup implemented.
+- FRR zebra configured with `dplane_grout` module when grout is enabled.
+- E2E passthrough tests passing with grout-enabled configuration.
+- Controller issues `grcli` command to configure grout.
+- Grout interfaces are based on TAP devices. No hardware acceleration at this stage.
+- grout runs in `--test-mode`, which means no huge pages are used at the moment.
+
+#### M1a
+_depends on M1_
+- Support for hardware acceleration with SR-IOV NICs, CPU isolation and huge pages for L3Passthrough scenario.
```
fedepaol: I feel like the scope of this enhancement stops at M1, leaving a lot uncovered. Does it make sense to expand on that section too (how do we think we'll integrate sriov vfs, for primary and secondary interface, and how do we envision the proper setup required to support hw acceleration).

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,183 @@
...
+configuration is skipped in favor of a Grout configuration.
+
+### CI and Build
+
+- A new image of the OpenPERouter will be shipped for this scenario. 
+- A new CI job (`build-grout-images`) builds a Docker image tagged
+  `quay.io/openperouter/router:grout` and based on `Dockerfile.grout`
+  Current `router:main` image is based on Alpine Linux, which is a non-libc distro and
+  DPDK/grout does not build on such system. Unless refactor the main image, which
+  is not a goal of this proposal, grout integration will be provided as a side image.
```
fedepaol, line 185: this means we'll deliver a openperouter image built for this purpose instead of leveraging the same one right? Is this the end goal or temporary? What are the drawbacks of shipping one single image?

What will be the naming scheme if we cut a release?

zeeke: The main problem in having a single image is that we have to move the current Alpine-based project image to a glibc one (like CentOS, Fedora, or Ubuntu). FRR only ships Alpine-based images AFAIK.

The two images will follow the versioning schema:
`quay.io/openperouter/router:v1.2.3`
`quay.io/openperouter/router:v1.2.3-grout`

I slightly prefer having a single image for the entire project, but I understand the Alpine-based image is  lighter, smaller, and it ships with a ready-to-use FRR build.

enhancements/grout-dataplane.md
zeeke: added some description. The implementation will require an additional design doc, I think.

enhancements/grout-dataplane.md
fedepaol: Fair enough, but can we install upstream frr instead of the one coming from grout (at least as an end goal)?

enhancements/grout-dataplane.md
fedepaol: It's fine if we want to split in multiple steps / design (and it makes total sense too, as we'll discover as we go), but then I think it'd make sense to rescope the title / description to this narrower surface

zeeke: I rephrased the Milestone chapter. Hope it's clearer now.

enhancements/grout-dataplane.md
zeeke: Yes, I put it in a work to tackle later. I agree on the value of shipping an upstream, official FRR version

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,264 @@
...
+  
+   `ip route add <eth3_net> dev main src <eth3 IP addr>`
+
+
+L3Passthrough
+1. The controller creates a pair grout_port/tap_interface (compared to the veth pair of kernel datapath):
+
+   `grcli interface add port pt-ns devargs net_tap1,iface=pt-host`
+
+   `grcli address add 192.169.10.1/24 iface pt-ns`
```
maiqueb: could this be a /31 ? do you only need one of these "veths" ?

enhancements/grout-dataplane.md
maiqueb: FWIW, I agree this can be tackled as a follow up goal. Not so sure about the order - but, to me, that's not relevant since it can be implemented in parallel with any other milestone.

enhancements/grout-dataplane.md
```diff
@@ -0,0 +1,264 @@
...
+_depends on M3_ 
+- HW acceleration support for L2VNI. The API must be enhanced with the information about a set of SR-IOV Virtual Function.
+  The VFs will be set on a configurable VLAN and will be used for intra-node communications. 
+- The first VF are attached to Grout. The other VFs stay in the host network namespace.
+- The user creates `sriov` NetworkAttachmentDefinitions to move the VF to the workload pod and leverage the L2VNI 
+  connectivity with kernel or DPDK applications.
+
+#### M4 - Optional
+- Implement a Golang library in github.com/DPDK/grout to issue commands to grout, instead of
+  using grcli
```
maiqueb: maybe re-order the priorities ? i.e. make M4 go after M5 ? like, this one is optional, the other isn't marked optional.

maiqueb (commented):
I'm very close to LGTMing.

enhancements/grout-dataplane.md
zeeke: sure, fixing also in L174

maiqueb (approved):
Tks

zeeke:
> I'm very close to LGTMing.

Thanks! When you have time, please take another look. @fedepaol as well

fedepaol:
Thanks!

PR #336, opened 2026-04-23
Remove hostNetwork, hostPID, and privileged from controller pod

qinqon:
## Summary

- Remove `hostNetwork: true`, `hostPID: true`, and `privileged: true` from the controller DaemonSet (K8s mode only)
- Mount host network namespace via `/proc/1/ns/net` and enter it explicitly using a new `InHost()` wrapper for netlink operations that need it
- Replace `privileged: true` with explicit capabilities (`NET_ADMIN`, `NET_RAW`, `SYS_ADMIN`) and `seccompProfile: Unconfined` (required for `setns()`)
- Simplify health probes: remove configurable `healthProbePort` Helm value and `--health-probe-bind-address` flag since port conflicts are no longer possible without host networking
- Drop `NET_BIND_SERVICE` capability (port 9081 is unprivileged)

## Test plan

- [ ] `make lint` passes
- [ ] `go build ./...` compiles
- [ ] Unit tests pass (`go test ./internal/hostnetwork/... -tags runasroot`)
- [ ] Controller pod starts without `hostNetwork`, `hostPID`, or `privileged`
- [ ] Health/readiness probes pass on pod IP
- [ ] VNI setup creates veths with host leg in host NS, PE leg in router NS
- [ ] Underlay interface moves to router NS correctly
- [ ] E2E test suite passes: `make e2etests`
- [ ] Verify on clusters with SELinux/AppArmor enforcing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/openperouter/README.md
+++ b/charts/openperouter/README.md
@@ -27,7 +27,6 @@ Kubernetes: `>= 1.19.0-0`
 | fullnameOverride | string | `""` |  |
 | nameOverride | string | `""` |  |
 | openperouter.affinity | object | `{}` |  |
-| openperouter.controller.healthProbePort | int | `9081` | Health probe port for liveness and readiness checks |
 | openperouter.controller.resources | object | `{}` |  |
 | openperouter.cri | string | `"containerd"` |  |
 | openperouter.frr.image.pullPolicy | string | `""` |  |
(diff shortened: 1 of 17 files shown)


fedepaol:
not sure I like the change about `hostNetwork: true`

qinqon:
> not sure I like the change about `hostNetwork: true`

This is an experiment connected with resilience, the idea is to have openperouter pods have norma namespace and just tap into host namespace when needed apart from security improvements is more kuberntes friendly with probes and monitoring.

qinqon:
Closing, it make things more complicated and the security gain is not clear

PR #338, opened 2026-04-24
feature: Support for DPDK/grout - L3Passthrough

zeeke:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

This PR introduces the integration with the DPDK/grout dataplane.

depends on:
- https://github.com/openperouter/openperouter/pull/335

**Release note**:

```release-note
DPDK/grout support for Underlay and L3Passthrough
```


zeeke:
Test failures are addressed in
- https://github.com/openperouter/openperouter/pull/390

charts/openperouter/templates/router.yaml
```diff
@@ -245,6 +260,78 @@ spec:
...
+            until [ -e /var/run/netns/perouter ]; do
+              echo "Waiting for named netns /var/run/netns/perouter..."
+              sleep 1
+              WAITED=$((WAITED + 1))
+              if [ "$WAITED" -ge "$TIMEOUT" ]; then
+                echo "Timed out waiting for named netns after ${TIMEOUT}s"
+                exit 1
+              fi
+            done
+            GROUT_ARGS="--test-mode --socket-mode 0666"
```
rjarry, line 289: Could you make `--test-mode` conditional on some other setting? E.g. if hugepages are not available on the node.

charts/openperouter/templates/router.yaml
```diff
@@ -245,6 +260,78 @@ spec:
...
+        {{- with .Values.openperouter.grout.resources }}
+        resources:
+          {{- toYaml . | nindent 12 }}
+        {{- end }}
+        startupProbe:
+          exec:
+            command:
+              - /bin/sh
+              - -c
+              - "test -S ${GROUT_SOCK_PATH}"
```
rjarry: This isn't reliable, the socket could exist but grout may not be ready to accept API messages. Could we rely on systemd notify socket instead? Who will be responsible of starting that grout container?

charts/openperouter/templates/router.yaml
```diff
@@ -245,6 +260,78 @@ spec:
...
+          periodSeconds: 10
+        livenessProbe:
+          exec:
+            command:
+              - grcli
+              - -e
+              - --socket
+              - /var/run/grout/grout.sock
+              - graph
+              - show
```
rjarry: If `GROUT_SOCK_PATH` is already exported, no need to specify `--socket`. Also, you could use `grcli -e` (no other arg) for all probes (startup, readiness, liveness):

```console
$ grcli -e -s /non/existent < /dev/null
error: gr_connect: No such file or directory
$ grcli -e -s README.md < /dev/null
error: gr_connect: Connection refused
$ echo $GROUT_SOCK_PATH
/tmp/tmp.U2OKgkGEjh/grout.sock
$ grcli -e < /dev/null
$ echo $?
0
```

Upon connection, `grcli` sends a `GR_HELLO` message to negotiate the API version with `grout`. Even if no actual commands are executed. If it succeeds, it means `grout` is ready to accept further commands.

charts/openperouter/templates/router.yaml
```diff
@@ -245,6 +260,78 @@ spec:
...
+              if [ "$WAITED" -ge "$TIMEOUT" ]; then
+                echo "Timed out waiting for named netns after ${TIMEOUT}s"
+                exit 1
+              fi
+            done
+            GROUT_ARGS="--test-mode --socket-mode 0666"
+            {{- if eq .Values.openperouter.logLevel "debug" }}
+            GROUT_ARGS="${GROUT_ARGS} --verbose"
+            {{- end }}
+            exec nsenter --net=/var/run/netns/perouter grout ${GROUT_ARGS}
```
rjarry, line 293: Could you always redirect logging to syslog to benefit from level filtering?

```
exec nsenter --net=/var/run/netns/perouter grout --syslog ${GROUT_ARGS}
```

operator/bindata/deployment/openperouter/templates/router.yaml
```diff
@@ -45,7 +45,11 @@ data:
     # Check /etc/pam.d/frr if you intend to use "vtysh"!
     #
     vtysh_enable=yes
+    {{- if .Values.openperouter.grout.enabled }}
+    zebra_options="  -A 127.0.0.1 -s 90000000 --limit-fds 100000 -K 60 -M dplane_grout"
+    {{- else }}
```
rjarry: why is this duplicated in that other file?

internal/grout/grout_client.go
```diff
@@ -0,0 +1,127 @@
...
+	cmdArgs := append([]string{"-e", "-s", c.socketPath}, args...)
+
+	slog.DebugContext(ctx, "running grcli", "args", cmdArgs)
+	out, err := execCmd(ctx, "grcli", cmdArgs...)
+	output := strings.TrimSpace(string(out))
+	if err != nil {
+		return output, fmt.Errorf("grcli %s failed: %w, output: %s", strings.Join(args, " "), err, output)
+	}
+	return output, nil
+}
```
rjarry, line 174: Have you considered direct API socket access using the public headers?

internal/grout/grout_client.go
```diff
@@ -0,0 +1,127 @@
...
+	err := c.run(ctx, "address", "add", cidr, "iface", ifaceName)
+	if err != nil && strings.Contains(err.Error(), "already") {
+		slog.DebugContext(ctx, "address already assigned", "iface", ifaceName, "cidr", cidr)
+		return nil
+	}
+	return err
+}
+
+func (c *Client) getAddresses(ctx context.Context, ifaceName string) ([]string, error) {
+	out, err := c.runOutput(ctx, "address", "show", "iface", ifaceName)
```
rjarry, line 107: You may want to enable json output to avoid the awkward text parsing.

```console
# grcli -j address iface p0 | jq
[
  {
    "iface": "p0",
    "family": "ipv4",
    "address": "172.16.0.1/24"
  },
  {
    "iface": "p0",
    "family": "ipv6",
    "address": "fe80::d0e4:78ff:fe6b:720f/64"
  }
]
```

charts/openperouter/templates/router.yaml
zeeke: good point about the socketfile existance, but systemd is not available in the container. 
the kubelet invokes startupProbe, readinessProbe and livenessProbe to understand if the container is working properly.
changing to `grcli -e --socket ${GROUT_SOCK_PATH} graph show`. are there any other good command to understand if grout is ready for work?

charts/openperouter/templates/router.yaml
zeeke: I'd prefer to add the support to hugepages and cpu isolation in a subsequent PR

charts/openperouter/templates/router.yaml
zeeke: that way the standard output will be empty, right? If yes, that might make troubleshooting a little more difficult for the pod deployment.
Maybe we can add `--syslog` to the quadlet/systemd deployment files when I'll tackle that point. WDYT? would it be a fair compromise?

operator/bindata/deployment/openperouter/templates/router.yaml
zeeke: `operator/bindata/deployment/openperouter/templates/router.yaml` is generated (copied) from `charts/openperouter/templates/router.yaml`, if I'm not mistaken.
@fedepaol @maiqueb can you confirm?

internal/grout/grout_client.go
zeeke: Yes, that's part of the enhancement proposal, [milestone 5](https://github.com/openperouter/openperouter/blob/main/enhancements/grout-dataplane.md?plain=1#L243-L245). 
Other than including a C header, is implementing the communication protocol viable?

internal/grout/grout_client.go
zeeke: this makes definitely sense. I'll work on it

internal/grout/grout_client.go
rjarry: That is worth consideration. I'm not sure if it is safer or faster. It may require more work since you will need to deal with reconnecting the client on failure/restarts. And also, you expose the controller to potential memory access errors.

I had started a go library implementation here but never finished it:

https://github.com/rjarry/grout/tree/go/go

Let's put that aside for now :)

charts/openperouter/templates/router.yaml
rjarry: oh sorry, I hadn't understood that grout was started by the controller.

charts/openperouter/templates/router.yaml
rjarry: simply, `grcli -e` without any argument will be ok. Provided you have `GROUT_SOCK_PATH` in the environment and that stdin is `/dev/null`.

e2etests/tests/dump.go
```diff
@@ -74,6 +74,9 @@ func dumpFRRInfo(basePath, testName string, cs clientset.Interface, clabContaine
 	for name, exec := range executors {
 		func() {
 			dump := frr.RawDump(exec)
+			groutDump := frr.GroutDump(exec)
```
fedepaol: I'd be more deliberate and accept an external parameter to understand if this needs to be done.
There are also netlinks commands that might not be relevant in grout mode

internal/controller/routerconfiguration/grout_config.go
```diff
@@ -0,0 +1,77 @@
...
+		return fmt.Errorf("failed to convert config to host configuration: %w", err)
+	}
+
+	slog.InfoContext(ctx, "setting up underlay")
+	if err := grout.SetupUnderlay(ctx, groutClient, hostConfig.Underlay); err != nil {
+		return fmt.Errorf("failed to setup underlay: %w", err)
+	}
+
+	slog.InfoContext(ctx, "setting up passthrough")
+	if hostConfig.L3Passthrough != nil {
```
fedepaol, line 71: I think this must follow the error reporting structure we have in https://github.com/openperouter/openperouter/blob/main/internal/controller/routerconfiguration/host_config.go#L83

internal/controller/routerconfiguration/static_configuration_controller.go
```diff
@@ -75,7 +77,11 @@ func (r *StaticConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	updater := frrconfig.UpdaterForSocket(r.FRRReloadSocket, r.FRRConfigPath)
 
-	err = Reconcile(ctx, apiConfig, r.NodeIndex, r.LogLevel, r.FRRConfigPath, targetNS, updater, configureInterfaces)
+	hostConfigurator := configureInterfaces
+	if r.GroutEnabled {
```
fedepaol: we should avoid doing this every time and have the configurator a member of the reconciler

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -149,7 +151,11 @@ func (r *PERouterReconciler) reconcile(ctx context.Context, logger *slog.Logger)
 		return ctrl.Result{}, err
 	}
 
-	err = Reconcile(ctx, config, nodeIndex, r.LogLevel, r.FRRConfigPath, targetNS, updater, configureInterfaces)
+	hostConfigurator := configureInterfaces
```
fedepaol: same

internal/grout/grout_client.go
fedepaol, line 1: nit, the file name can be just client

internal/grout/grout_client.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if out == "" || out == "[]" {
```
fedepaol: won't this handled by the unmarshal below?

internal/grout/grout_client.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return addrs, nil
+}
+
+// portExists checks whether a port with the given name exists in grout.
+func (c *Client) portExists(ctx context.Context, name string) (bool, error) {
+	out, err := c.runOutput(ctx, "interface", "show", "name", name)
+	if err != nil {
+		// grcli returns an error when the interface doesn't exist
+		if strings.Contains(err.Error(), "No such") || strings.Contains(out, "No such") {
```
fedepaol, line 144: here it seems we are fishing for the error, and it seems a bit fragile. Can we agree on error codes with grout? Same for the other places where we parse the error.

internal/grout/helpers.go
```diff
@@ -0,0 +1,150 @@
...
+	"log/slog"
+	"net"
+
+	"github.com/openperouter/openperouter/internal/netnamespace"
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netns"
+)
+
+// assignIPsToLink assigns IPv4 and/or IPv6 addresses to a netlink interface.
+func assignIPsToLink(link netlink.Link, ipv4, ipv6 string) error {
```
fedepaol: there's already an assignIPsToInterface in hostnetork, makes sense to make it public instead of duplicating

internal/grout/helpers.go
```diff
@@ -0,0 +1,150 @@
...
+		if err := client.ensureAddress(ctx, portName, addr); err != nil {
+			return fmt.Errorf("failed to assign address %s to grout port %s: %w", addr, portName, err)
+		}
+	}
+	return nil
+}
+
+// moveLinkToHostNamespace moves a link from the given namespace to the current
+// (host) namespace. If the link is already in the host namespace, this is a no-op.
+func moveLinkToHostNamespace(ctx context.Context, name string, srcNS netns.NsHandle) error {
```
fedepaol: we have already a moveInterfaceToNamespace function, I'd try to understand if we can use that instead of duplicating (even if it's host -> ns, we'll have the other direction soon with https://github.com/openperouter/openperouter/pull/442)

internal/grout/helpers.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if err := netlink.LinkSetUp(link); err != nil {
+		return fmt.Errorf("failed to bring %s up: %w", name, err)
+	}
+
+	slog.DebugContext(ctx, "link moved to host namespace", "name", name)
+	return nil
+}
+
+func deleteAddressFromKernelInterface(ifaceName string, addr netlink.Addr) error {
```
fedepaol: I'd call this "removeAddressfromNetlinkInterface" . Also, it makes sense to move all the netlink operation to the hostnetwork package or to have a shared netlink package. For now I'd just move it. Also, please check we don't have it already.

internal/grout/helpers.go
```diff
@@ -0,0 +1,150 @@
...
+	if err := netlink.LinkSetHardwareAddr(link, mac); err != nil {
+		return fmt.Errorf("failed to set MAC on %s: %w", name, err)
+	}
+	if err := netlink.LinkSetUp(link); err != nil {
+		return fmt.Errorf("failed to bring %s up: %w", name, err)
+	}
+	return nil
+}
+
+func removeLinkByName(name string) error {
```
fedepaol: same requests as per the other netlink functions.

internal/grout/helpers.go
```diff
@@ -0,0 +1,150 @@
...
+		return fmt.Errorf("failed to find underlay interface %s: %w", ifaceName, err)
+	}
+	return netlink.AddrDel(link, &addr)
+}
+
+// setUniqueMAC assigns a random locally-administered unicast MAC to the named
+// link. TAP devices share the same MAC on both ends, which causes IPv6
+// DAD failures on the link-local address and prevents NDP from working.
+// Giving the host-side TAP its own MAC avoids both problems.
+func setUniqueMAC(name string) error {
```
fedepaol: let's make this a function of the link name instead of totally random if possible.

internal/grout/helpers.go
```diff
@@ -0,0 +1,150 @@
+// SPDX-License-Identifier:Apache-2.0
```
fedepaol: given the request below of moving / refactoring the majority of the functions, consider if it makes sense to move the leftovers directly in the bottom of the file where they are used. Otherwise, please rename this file in a more meaningful way.

internal/grout/passthrough.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to check grout port %s: %w", portName, err)
+	}
+
+	if !portExists {
+		if err := createPassthroughPort(ctx, client, portName, tapName, ns); err != nil {
+			return err
+		}
+	} else {
+		slog.InfoContext(ctx, "grout passthrough port already exists, skipping TAP setup", "port", portName)
```
fedepaol: I think it makes sense to always log, or even better we can have createPassthroughPort be idempotent and call it ensurePassthroughPort

internal/grout/passthrough.go
```diff
@@ -0,0 +1,96 @@
...
+	hostTap, err := netlink.LinkByName(tapName)
+	if err != nil {
+		return fmt.Errorf("host TAP %s not found after move: %w", tapName, err)
+	}
+	if err := assignIPsToLink(hostTap, params.HostVeth.HostIPv4, params.HostVeth.HostIPv6); err != nil {
+		return fmt.Errorf("failed to assign IPs to host TAP: %w", err)
+	}
+
+	// Assign IPs to the grout port "pt-ns" via grcli.
+	if err := assignIPsToGroutPort(ctx, client, portName, params.HostVeth.NSIPv4, params.HostVeth.NSIPv6); err != nil {
```
fedepaol: if the port exists but the ips change, we'll end up with double addresses right? If so, we should consider deleting if it's a possible scenario

internal/grout/passthrough.go
fedepaol: also, looking at `createPassthroughPort` implementation, it looks like the port might exist but not properly setup (ie if port exists but it's not on the proper namespace, we won't move it)

internal/grout/passthrough.go
```diff
@@ -0,0 +1,96 @@
...
+		return fmt.Errorf("failed to move TAP to host namespace: %w", err)
+	}
+
+	if err := setUniqueMAC(tapName); err != nil {
+		return fmt.Errorf("failed to set unique MAC on host TAP: %w", err)
+	}
+	return nil
+}
+
+func RemovePassthrough(ctx context.Context, client *Client, targetNS string) error {
```
fedepaol: what happens if the network namespace gets deleted? Will the taps persist? Or we might end up in a situation where grout thinks it has ports but the taps are not there anymore?

internal/grout/underlay.go
```diff
@@ -0,0 +1,162 @@
...
+
+		slog.InfoContext(ctx, "migrated underlay address to grout", "cidr", cidr, "iface", underlayPortName)
+	}
+	return nil
+}
+
+// RemoveUnderlay tears down the grout underlay port and resets the kernel
+// group IDs on moved NICs so HasUnderlayInterface returns false on the next
+// reconcile.
+func RemoveUnderlay(ctx context.Context, client *Client, targetNS string) error {
```
fedepaol: this is public, let's move it above

internal/hostnetwork/underlay.go
```diff
@@ -195,7 +195,7 @@ func HasUnderlayInterface(namespace string) (bool, error) {
 
 // findInterfacesInGroup returns a slice of interface names
 // for all interfaces in the namespace that belong to the specified group.
-func findInterfacesInGroup(ns netns.NsHandle, groupID uint32) ([]string, error) {
+func FindInterfacesInGroup(ns netns.NsHandle, groupID uint32) ([]string, error) {
```
fedepaol, line 242: let's file an issue to move netlink operations to a shared package. I prefer not to do it here as this is already complex enough

internal/webhooks/l2vni_webhook.go
```diff
@@ -63,6 +65,10 @@ func (v *L2VNIValidator) Handle(ctx context.Context, req admission.Request) (res
 		}
 	}
 
+	if v.groutEnabled && (req.Operation == v1.Create || req.Operation == v1.Update) {
```
fedepaol: don't we have this already in the validation functions?

operator/bindata/deployment/openperouter/templates/router.yaml
fedepaol: yep, this is the operator payload leveraging the helm charts

charts/openperouter/templates/controller.yaml
fedepaol, line 1: even if we don't generate an all-in-one for grout, I'd like to have a kubebuilder overlay too

internal/grout/grout_client.go
zeeke: that's look interesting!

Dockerfile.grout
```diff
@@ -0,0 +1,71 @@
...
+ARG GIT_COMMIT=dev
+ARG GIT_BRANCH=dev
+ARG TARGETOS
+ARG TARGETARCH
+
+WORKDIR $GOPATH/openperouter
+RUN --mount=type=cache,target=/go/pkg/mod/ \
+  --mount=type=bind,source=go.sum,target=go.sum \
+  --mount=type=bind,source=go.mod,target=go.mod \
+  go mod download -x
```
maiqueb, line 27: does this work w/ podman ?

Dockerfile.grout
```diff
@@ -0,0 +1,71 @@
...
+COPY --from=builder /go/openperouter/operatorbinary ./operator
+
+COPY operator/bindata bindata
+# Copy FRR startup configuration to the default location
+COPY systemdmode/frrconfig/daemons /etc/frr/daemons
+COPY systemdmode/frrconfig/vtysh.conf /etc/frr/vtysh.conf
+COPY systemdmode/frrconfig/frr.conf /etc/frr/frr.conf
+
+RUN ln -s /usr/libexec/frr /usr/lib/frr
+ADD https://raw.githubusercontent.com/FRRouting/frr/refs/tags/frr-10.5.0/docker/ubi8-minimal/docker-start /usr/lib/frr/docker-start
```
maiqueb: this version of FRR is different than the version we use for the kernel data plane ... 

Would we want to keep them aligned ?

e2etests/tests/dump.go
zeeke: ok, adding a --groutmode to the e2e test suite

internal/grout/grout_client.go
zeeke: I prefer grout_client.go:
the name of the file is not mentioned in any place in the code, but you can see it when you have multiple open tabs in the editor. having the `grout_` prefix helps browsing the codebase.

e2etests/tests/dump.go
fedepaol: I would call it --datapath kernel / grout vs "grout mode"

internal/grout/grout_client.go
zeeke: good point. @rjarry would the work in https://github.com/rjarry/grout/tree/go/go help for this matter?
or, any other ideas to improve this part?

charts/openperouter/README.md
```diff
@@ -36,6 +36,14 @@ Kubernetes: `>= 1.19.0-0`
 | openperouter.frr.reloader.resources | object | `{}` |  |
 | openperouter.frr.reloader.vtyshTimeout | string | `""` | Timeout for vtysh commands used in health checks. Increase under heavy VNI load. Defaults to 10s. |
 | openperouter.frr.resources | object | `{}` |  |
+| openperouter.grout.enabled | bool | `false` | Enable grout sidecar for DPDK-accelerated L3 routing |
```
fedepaol: I would have a openperouter.dataplane value that triggers the choice of kernel vs grout.

internal/grout/helpers.go
zeeke: ok, I'll wait for #442 and I'll rebase it

internal/grout/helpers.go
zeeke: I'll remove the file

internal/grout/passthrough.go
zeeke: which namespace? the TAP inteface stays on the host namespace, which can't be deleted.

Also, removing the `targetNS` argument from the function, as it's not used

internal/hostnetwork/underlay.go
zeeke: created:
- https://github.com/openperouter/openperouter/issues/530

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -19,10 +19,10 @@ import (
 )
 
 func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, nodeIndex int, logLevel, frrConfigPath, targetNamespace string, updater frr.ConfigUpdater, hostConfigurator HostConfigurator) error {
-	normalizeConfig(&apiConfig)
```
maiqueb, line 25: why did you move this ? 

isn't it problematic that first you validate the underlay, and only then normalize the struct where the underlays are ? is it possible for `normalizeConfig` to change the underlay configuration ?

internal/controller/routerconfiguration/static_configuration_controller.go
maiqueb: I agree.

internal/grout/grout_client.go
```diff
@@ -0,0 +1,134 @@
...
+	Address string `json:"address"`
+}
+
+// NewClient creates a new grout client pointing at the given UNIX socket.
+func NewClient(socketPath string) *Client {
+	return &Client{socketPath: socketPath}
+}
+
+func (c *Client) ensurePort(ctx context.Context, name, devargs string) error {
+	exists, err := c.portExists(ctx, name)
```
maiqueb, line 45: afaiu, this only check the port exists . Meaning, if something else (i.e. other than open pe) is holding that port, this function would not do anything.

internal/grout/grout_client_test.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		return nil, fmt.Errorf("unexpected command: [%s]", cmd)
+	}
+	return func() {
+		execCmd = original
+	}
+}
+
+const interfaceShowP0Output = `{"name":"p0","type":"port","id":2,"flags":["up","running","allmulti","tracing"],"mode":"VRF","domain":"main","mtu":1500,"speed":"unknown"}`
```
maiqueb: nit: can we get some format here ? it'd be easier to understand imho

internal/grout/grout_client_test.go
```diff
@@ -0,0 +1,152 @@
...
+				cmd: "grcli --err-exit --json --socket sock interface show name p0",
+				err: fmt.Errorf("error: command failed: No such device (ENODEV)"),
+			},
+			cmdCall{
+				cmd: "grcli --err-exit --json --socket sock interface add port p0 devargs net_tap0,remote=remote_i,iface=p0_tap",
+			})()
+
+		c := NewClient("sock")
+		err := c.ensurePort(context.Background(), "p0", "net_tap0,remote=remote_i,iface=p0_tap")
+		assert.NoError(t, err)
```
maiqueb: nit: this is opinionated.

maybe we can inline the calls ? The same applies to the other tests. 

i.e. 
```golang
assert.NoError(
    t, 
    NewClient("sock").ensurePort(
        context.Background(), 
        "p0", 
        "net_tap0,remote=remote_i,iface=p0_tap",
    ),
)
```

internal/grout/helpers.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		if err := netlink.AddrReplace(link, parsed); err != nil {
+			return fmt.Errorf("failed to assign address %s to %s: %w", addr, link.Attrs().Name, err)
+		}
+	}
+	return nil
+}
+
+// assignIPsToGroutPort assigns IPv4 and IPv6 addresses to a grout port via grcli.
+func assignIPsToGroutPort(ctx context.Context, client *Client, portName string, ipv4, ipv6 string) error {
```
maiqueb: nit: maybe make the input addresses variadic, and fail if that slice is empty ?

internal/grout/helpers.go
maiqueb: I agree. This should be scoped to netlink instead, and re-used whenever appropriate.

internal/grout/passthrough.go
maiqueb: +1 for having it idempotent

internal/grout/passthrough.go
```diff
@@ -0,0 +1,96 @@
...
+		return fmt.Errorf("failed to assign IPs to grout port: %w", err)
+	}
+
+	return nil
+}
+
+func createPassthroughPort(ctx context.Context, client *Client, portName, tapName string, ns netns.NsHandle) error {
+	_ = removeLinkByName(tapName)
+
+	devargs := fmt.Sprintf("net_tap1,iface=%s", tapName)
```
maiqueb: there's something I'm missing, apologies ...

how can we hard-code the `net_tapX` index to 1 ?

internal/grout/underlay.go
```diff
@@ -0,0 +1,162 @@
...
+
+func validateNoStaleUnderlays(ns netns.NsHandle, wanted []string) error {
+	existingIfaces, err := hostnetwork.FindInterfacesInGroup(ns, hostnetwork.UnderlayGroupID)
+	if err != nil {
+		return fmt.Errorf("failed to check existing underlay interfaces: %w", err)
+	}
+	for _, name := range existingIfaces {
+		if !slices.Contains(wanted, name) {
+			return hostnetwork.UnderlayExistsError(fmt.Sprintf(
+				"existing underlay found: %s, new inteUnderlayGroupIDrfaces are %v", name, wanted))
```
maiqueb: there's surely a typo here: `inteUnderlayGroupIDrfaces` probably you meant `underlayGroupIDInterfaces`

internal/grout/underlay.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := netlink.LinkSetUp(underlay); err != nil {
+		return fmt.Errorf("could not set link up for VRF %s: %v", underlay.Attrs().Name, err)
+	}
+
+	underlayAddrs, err := readUnderlayAddresses(underlayInterface)
+	if err != nil {
+		return fmt.Errorf("failed to read underlay interface addresses: %w", err)
+	}
+
+	devargs := fmt.Sprintf("net_tap0,remote=%s,iface=%s", underlayInterface, "tap_"+underlayInterface)
```
maiqueb: here you hard-code the tap index to 0. I assume this must happen *before* the passthrough port. 

Should we rely on order here ? could this be order independent ?

.github/workflows/ci.yaml
```diff
@@ -199,10 +225,11 @@ jobs:
       - build-kind-node-image
       - unit-tests
       - build-test-images
+      - build-grout-images
     strategy:
       fail-fast: false
       matrix:
-        deployment: [manifests, helm, operator, systemdmode, hostmode-boot]
+        deployment: [manifests, helm, operator, systemdmode, hostmode-boot, helm-grout]
```
maiqueb: so we're making a conscious choice and only testing grout on a single lane (newly added `helm-grout`), right ? 

Works for me.

e2etests/pkg/frr/grout_dump.go
```diff
@@ -0,0 +1,44 @@
...
+	out, err := exec.Exec("grcli", "--version")
+	if err != nil {
+		return "grcli not available"
+	}
+
+	var res strings.Builder
+	res.WriteString("Grout version:\n" + out + "\n\n")
+
+	commands := []command{
+		{desc: "grcli interface show", cmd: []string{"grcli", "interface", "show"}},
```
maiqueb, line 22: is there some fdb equivalent we can use ? 

maybe not for now, since you're only implementing for passthrough - but once we get to l2vnis, that'll be useful.

internal/webhooks/l2vni_webhook.go
zeeke: it got dropped during one of my rebases. I'll update the validation part

cmd/nodemarker/main.go
```diff
@@ -182,7 +185,7 @@ func main() {
 				logger.Error("unable to add v1alpha1 scheme", "error", err)
 			}
 
-			err := setupWebhook(mgr, logger)
+			err := setupWebhook(mgr, logger, args.groutEnabled)
```
maiqueb: can we try to come up with a way to setup the webhook without passing it a boolean ? e.g. use a builder, or multiple functions ?.. same for the `SetupL3VNI` / `SetupL2VNI` functions.

cmd/hostcontroller/main.go
```diff
@@ -88,14 +88,16 @@ type k8sModeParameters struct {
...
+	probeAddr       string
+	frrConfigPath   string
+	reloaderSocket  string
+	mode            string
+	ovsSocketPath   string
+	nodeName        string
+	namespace       string
+	logLevel        string
+	groutEnabled    bool
+	groutSocketPath string
```
maiqueb: do we really need these two flags ? i.e. couldn't we do the same thing with just the `groutSocketPath` variable ? 

I can't imagine why we'd pass the socket but want to have grout disabled ... or the other way around.

maiqueb (changes requested):
Really nice PR. 

Mostly I have some questions and opinionated nits. 

To move this forward, I would focus on:
- explaining the hard-coded tap indexes
- remove the boolean input args from the API
- have just a single input flag to the controller parameters

charts/openperouter/templates/controller.yaml
zeeke: sounds good to me, but I have to add a CI lane for that, otherwise the configuration could get stale. does it sound good?

charts/openperouter/README.md
zeeke: it'd diverge from the enhancement proposal. is it worth also updating that? I can also do it in a subsequent PR

Dockerfile.grout
zeeke: actually, it doesn't:
```podman build -f Dockerfile.grout .
...
[1/3] STEP 7/13: RUN --mount=type=cache,target=/go/pkg/mod/   --mount=type=bind,source=go.sum,target=go.sum   --mount=type=bind,source=go.mod,target=go.mod   go mod download -x
go: open /go/openperouter/go.mod: permission denied
Error: building at STEP "RUN --mount=type=cache,target=/go/pkg/mod/ --mount=type=bind,source=go.sum,target=go.sum --mount=type=bind,source=go.mod,target=go.mod go mod download -x": while running runtime: exit status 1
```

and it doesn't work for the main image too:
```
podman build -f Dockerfile .                                          
...
[1/2] STEP 7/13: RUN --mount=type=cache,target=/go/pkg/mod/   --mount=type=bind,source=go.sum,target=go.sum   --mount=type=bind,source=go.mod,target=go.mod   go mod download -x
go: open /go/openperouter/go.mod: permission denied
Error: building at STEP "RUN --mount=type=cache,target=/go/pkg/mod/ --mount=type=bind,source=go.sum,target=go.sum --mount=type=bind,source=go.mod,target=go.mod go mod download -x": while running runtime: exit status 1
```

I can open an issue for that and fix in a second PR. does it sound good?

charts/openperouter/README.md
maiqueb: I do prefer @fedepaol 's suggestion to what we agreed on the enhancement.

Dockerfile.grout
zeeke: Updating this specific line. What it really matters is the FRR version grout is built with, which is currently 10.6.
We should keep that aligned with what we have in Dockerfile, but that doesn't come for free.

charts/openperouter/README.md
fedepaol: I am lost. My suggestion is to diverge from the enhancement (which I think is fine and will always happen).
I would not change the enhancement because it's a snapshot of our decision.

I would not defer it as it's not a tremendous change.

internal/controller/routerconfiguration/reconcile.go
zeeke: that must be a leftover, reverting

Dockerfile.grout
maiqueb: yes, of course. I wasn't suggesting you should fix what's broken for podman img building in this PR ! 

Just want us to be aware that we can't use podman to build the images. It did work with buildah though: `buildah build -f Dockerfile .` 

I am OK with not even opening an issue (I wasn't aware it worked w/ buildah).

internal/grout/grout_client.go
zeeke: correct. I don't see a specific problem with that, do you?

Grout ports are like netlink interfaces. In the kernel datapath, if some other actor creates netlinks with names conflicting with the perouter, the controller fails. it doesn't change its behavior, right?

Dockerfile.grout
maiqueb: let's maybe open an issue for that ?

internal/grout/grout_client.go
maiqueb: yeah, you're right.

internal/grout/helpers.go
zeeke: I understand your point, but that follows the same pattern as:
```
func assignIPsToInterface(link netlink.Link, ipv4, ipv6 string) error {
```
I'd leave it with `ipv4, ipv6 string` until needed. WDYT?

maiqueb: it's a nit - I'm aware it was opinionated - no need to do anything.

internal/grout/passthrough.go
zeeke: you're right. I fixed with an incremental int.

internal/grout/underlay.go
zeeke: I moved to an incremental approach here. BTW, the number is not really important.
What matters here is the `net_tap` prefix, followed by a number.

.github/workflows/ci.yaml
zeeke: yes, that's my opinionated choice. I'm also adding a lane for the `grout` kustomize layer, but I wouldn't go by duplicating all the lanes. 

Maybe the systemd one in the future

e2etests/pkg/frr/grout_dump.go
zeeke: right. as of now `grcli route show` and `grcli nexthop show` do most of the job. When we'll deal with l2vni we'll have to dig into https://www.mankier.com/1/grcli-fdb

cmd/nodemarker/main.go
zeeke: I'm open to suggestions here. If we don't go with global variables or reading from environment variable around the code, I don't see how to remove this boolean (and I want to remove it!)

cmd/hostcontroller/main.go
zeeke: makes sense to me. @fedepaol any thoughts before I go with this change?

Dockerfile.grout
zeeke: sounds good!

zeeke:
addressed most of the comments.
@fedepaol @rjarry @maiqueb please take another look when you have time.

internal/controller/routerconfiguration/reconcile.go
maiqueb: this change was reverted ; resolving.

internal/grout/grout_client_test.go
maiqueb: tks for this. imho, it's a lot easier to understand this way.

internal/grout/grout_client_test.go
maiqueb: tks, resolving

internal/grout/passthrough.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// Assign IPs to the grout port "pt-ns" via grcli.
+	if err := ensurePortAddresses(ctx, client, portName, params.HostVeth.NSIPv4, params.HostVeth.NSIPv6); err != nil {
+		return fmt.Errorf("failed to ensure IPs to grout port: %w", err)
+	}
+
+	return nil
+}
+
+var tapIndex int = 0
```
maiqueb: let's say the controller restarts for some reason ...

won't this value reset ? could this be a problem ? (I'm aware you said the tap index was NOT a problem, I'm just making sure).

internal/grout/underlay.go
maiqueb: resolving

.github/workflows/ci.yaml
maiqueb: fair enough for me - I don't have a better suggestion.

e2etests/pkg/frr/grout_dump.go
maiqueb: perfect, that makes sense to me. 

Let's worry about it when we're implementing L2VNI support for grout integration

charts/openperouter/templates/controller.yaml
fedepaol: right. I theory we might even need one for operator plus grout. Now, the problem is that what costs a lot in terms of CI time is the deployment (running the e2es costs a lot too, but you can't get a discount from that).

Now, what we could do is to have one big lane where we cover everything functionally (meaning, every single test) and another one where we just test all the deployment methods / combos by deploying / undeploying (ie manifests first, charts and then operator) and we just run a subset of tests. But I guess this is material for a separate issue (cc @maiqueb ).

charts/openperouter/README.md
fedepaol: this is still unaddressed I guess?

cmd/hostcontroller/main.go
fedepaol: I would prefer having a "datapath" parameter with kernel as default / grout as other option instead of a raw flag. If we do that, we'll need the socket parameter anyways.
In general I am not a fan of inferring a behavior from a technical parameter, even if what @maiqueb says is totally legit, no one will mount the socket without using grout.

internal/grout/passthrough.go
zeeke: that's a good point. the number itself is not really meaningful but if it restarts from 0 and collides with the same number of another port, we have a problem:

```
grcli interface add port p0 devargs net_tap1,iface=x-p0
Created interface 2

grcli interface add port p1 devargs net_tap1,iface=x-p1
error: command failed: File exists (EEXIST)                 
```

as per my knowledge, we can go two way:
a. We save the counter to a file. the max available number should be big enough 
b. We go randomly between 0 and <max>; all the operations in the grout setup are supposed to be repeated in case of error.

@rjarry any suggestion on this matter? any best practice? if not, what's the max number we can set with `net_tapXXX` ?

cmd/hostcontroller/main.go
maiqueb: let's compromise and use some sort of enum instead of enabling grout.

cmd/nodemarker/main.go
maiqueb: we could maybe use a builder, which would select all the "setup CRD" functions you use grout helpers for. 

i.e. have a function pointer defaulting to setup kernel underlay, setup kernell2vni, etc. 

This `withGrout` option would change all these to use the functions you're providing. 

... I hope the function types are the same in both modes - I haven't checked. 

Let me loop in @RamLavi as well - he might think of an alternative for this.

charts/openperouter/README.md
zeeke: yes, I missed your answer, sorry. I'll address ASAP

internal/controller/routerconfiguration/host_config.go
```diff
@@ -24,9 +24,24 @@ type interfacesConfiguration struct {
...
+	Configure(ctx context.Context, config interfacesConfiguration) error
+	Validate(apiConfig conversion.APIConfigData) error
+}
+
+// KernelDatapathConfigurator configures the host via kernel netlink calls.
+// It supports the full API surface.
+type KernelDatapathConfigurator struct{}
+
+func (k *KernelDatapathConfigurator) Configure(ctx context.Context, config interfacesConfiguration) error {
+	return configureInterfaces(ctx, config)
```
fedepaol: can we move the configureInterfaces implementation here and avoid the wrapper?

cmd/hostcontroller/main.go
zeeke: ok, i'll go with 
```
--datapath kernel|grout
--grout-socket /var/run/grout/grout.sock
```

internal/controller/routerconfiguration/host_config.go
```diff
@@ -24,9 +24,24 @@ type interfacesConfiguration struct {
 	conversion.APIConfigData
 }
 
-// HostConfigurator applies host-level network configuration (interfaces, VNIs, sysctls).
-// Injected into Reconcile so tests can substitute a no-op implementation.
-type HostConfigurator func(ctx context.Context, config interfacesConfiguration) error
+// DatapathConfigurator abstracts host-level network configuration so the
+// reconciler can work with different datapaths (kernel netlink, grout/DPDK).
+type DatapathConfigurator interface {
```
fedepaol: either we merge host_config.go and grout_config.go or we move this somewhere else (maybe in reconcile.go?) as it's not host_config specific

internal/controller/routerconfiguration/host_config.go
```diff
@@ -24,9 +24,24 @@ type interfacesConfiguration struct {
...
+	Configure(ctx context.Context, config interfacesConfiguration) error
```
fedepaol: configure and validate can be two different interfaces, as they don't overlap

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -364,6 +367,11 @@ func (r *PERouterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		builder = builder.WatchesRawSource(source.Channel(r.TriggerChan, &handler.EnqueueRequestForObject{}))
 	}
 
+	r.datapathConfigurator = &KernelDatapathConfigurator{}
```
fedepaol: let's provide the datapath configurator directly from outside

internal/grout/passthrough.go
```diff
@@ -0,0 +1,232 @@
...
+
+		if err := setUniqueMAC(tapName); err != nil {
+			return fmt.Errorf("failed to set unique MAC on host TAP: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func RemovePassthrough(ctx context.Context, client *Client) error {
```
fedepaol: public, should go above

internal/grout/passthrough.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+	if err := netlink.LinkSetUp(link); err != nil {
+		return fmt.Errorf("failed to bring %s up: %w", name, err)
+	}
+	return nil
+}
+
+// moveLinkToHostNamespace moves a link from the given namespace to the current
+// (host) namespace. If the link is already in the host namespace, this is a no-op.
+func moveLinkToHostNamespace(ctx context.Context, name string, srcNS netns.NsHandle) error {
```
fedepaol, line 141: is this still used? Can we use MoveInterfaceToNamespace ?

internal/grout/passthrough.go
```diff
@@ -0,0 +1,232 @@
...
+	if err := ensureTapPortInHostNamespace(ctx, client, portName, tapName, "main", peRouterNs); err != nil {
+		return err
+	}
+
+	// Assign IPs to the host-side TAP (now in host namespace).
+	hostTap, err := netlink.LinkByName(tapName)
+	if err != nil {
+		return fmt.Errorf("host TAP %s not found after move: %w", tapName, err)
+	}
+	if err := hostnetwork.AssignIPsToInterface(hostTap, params.HostVeth.HostIPv4, params.HostVeth.HostIPv6); err != nil {
```
fedepaol: we should rename hostveth to something like hostLink or something, as we are now using hostVeth for taps

internal/grout/passthrough.go
```diff
@@ -0,0 +1,232 @@
...
+		return fmt.Errorf("failed to assign IPs to host TAP: %w", err)
+	}
+
+	// Assign IPs to the grout port "pt-ns" via grcli.
+	if err := ensurePortAddresses(ctx, client, portName, params.HostVeth.NSIPv4, params.HostVeth.NSIPv6); err != nil {
```
fedepaol: I know we asked for ensure, but maybe it's better to have simmetry with the previous one and call it something like assignIPsToGroutPort ?

internal/grout/passthrough.go
fedepaol: right, I guess we should fetch the highest currently existing and start from that one?
Also, how about those which gets deleted?

internal/grout/underlay.go
```diff
@@ -0,0 +1,346 @@
...
+
+		if err := client.ensureAddress(ctx, UnderlayPortNamePrefix+underlayInterface, cidr); err != nil {
+			return fmt.Errorf("failed to assign address %s to grout underlay port: %w", cidr, err)
+		}
+
+		if err := hostnetwork.DeleteAddressFromInterface(underlayInterface, addr); err != nil {
+			return fmt.Errorf("failed to remove address %s from underlay interface: %w", cidr, err)
+		}
+
+		if err := ensureKernelSubnetRoute("main", addr.IPNet.String()); err != nil {
```
fedepaol, line 192: I think we need some comments in why the routes are needed here, as it's not super clear to me.
Please comment the 3 steps here and in its symmetric one

internal/grout/underlay.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+	}
+
+	return nil
+}
+
+// RemoveUnderlay tears down the grout underlay ports and resets the kernel
+// group IDs on moved NICs so HasUnderlayInterface returns false on the next
+// reconcile.
+func RemoveUnderlay(ctx context.Context, client *Client, targetNS string) error {
```
fedepaol: please note we added bringing back to the original namespace too and not only deleting it in https://github.com/openperouter/openperouter/pull/442 , so we should support doing the same with grout

internal/grout/underlay.go
```diff
@@ -0,0 +1,346 @@
...
+		if err := netlink.AddrAdd(link, parsed); err != nil {
+			return fmt.Errorf("failed to assign address %s to %s: %w", cidr, linkName, err)
+		}
+		slog.InfoContext(ctx, "migrated address back to kernel interface", "cidr", cidr, "iface", linkName)
+	}
+
+	return nil
+}
+
+func ensureKernelSubnetRoute(ifaceName, addr string) error {
```
fedepaol, line 245: can we split this in a connectedRouteForAddress() which fetches the route , and another one that does the ensure part?

cmd/nodemarker/main.go
fedepaol: I think it's totally legit to split the datapathConfigurator in a configuratior and a datapathConfigValidator  interface and to inject the proper validator in the webhook package as a global.

internal/grout/underlay.go
```diff
@@ -0,0 +1,346 @@
...
+
+		if err := removeKernelSubnetRoute("main", addr); err != nil {
+			return fmt.Errorf("failed to remove kernel route for %s: %w", addr, err)
+		}
+	}
+
+	return nil
+}
+
+func readUnderlayAddresses(ifaceName string) ([]netlink.Addr, error) {
```
fedepaol: this should go to link_properties as addressesForInterface

internal/grout/underlay.go
```diff
@@ -0,0 +1,346 @@
...
+		if addr.IP.IsLinkLocalUnicast() {
+			continue
+		}
+		addrs = append(addrs, addr)
+	}
+
+	return addrs, nil
+}
+
+func setAddresses(ctx context.Context, linkName string, addrs []string) error {
```
fedepaol: we should reuse assignIPToInterface

internal/grout/underlay.go
```diff
@@ -0,0 +1,346 @@
...
+		return fmt.Errorf("failed to get link %s: %w", linkName, err)
+	}
+
+	for _, cidr := range addrs {
+		parsed, err := netlink.ParseAddr(cidr)
+		if err != nil {
+			return fmt.Errorf("failed to parse address %s: %w", cidr, err)
+		}
+
+		if parsed.IP.IsLinkLocalUnicast() {
```
fedepaol: this filter should happen outside, otherwise we are hiding the behavior

internal/grout/underlay.go
```diff
@@ -0,0 +1,346 @@
...
+	link, err := netlink.LinkByName(ifaceName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find underlay interface %s: %w", ifaceName, err)
+	}
+	all, err := netlink.AddrList(link, netlink.FAMILY_ALL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list addresses on %s: %w", ifaceName, err)
+	}
+	for _, addr := range all {
+		if addr.IP.IsLinkLocalMulticast() {
```
fedepaol: we should filter these outside

internal/grout/underlay.go
```diff
@@ -0,0 +1,346 @@
...
+	}); err != nil {
+		return fmt.Errorf("failed to add route for %s dev %s: %w", ipNet, ifaceName, err)
+	}
+
+	slog.Info("added kernel route for subnet", "cidr", addr, "src", srcAddr, "ipnet", ipNet, "iface", ifaceName)
+	return nil
+}
+
+func removeKernelSubnetRoute(ifaceName, addr string) error {
+	srcAddr, ipNet, err := net.ParseCIDR(addr)
```
fedepaol, line 291: same here, let's split

internal/webhooks/l2vni_webhook.go
```diff
@@ -77,6 +79,12 @@ func (v *L2VNIValidator) Handle(ctx context.Context, req admission.Request) (res
 			return admission.Denied(err.Error())
 		}
 	}
+
+	if v.groutEnabled && (req.Operation == v1.Create || req.Operation == v1.Update) {
```
fedepaol: instead of redoing the ifs, I'd move this below in validateL2VNICreate / Update, or maybe even better directly in the conversion.Validate** functions?

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -63,7 +63,10 @@ func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, nodeInde
 		RawFRRConfigs: apiConfig.RawFRRConfigs,
 	}
 
-	err = hostConfigurator(ctx, interfacesConfiguration{
+	err = datapathConfigurator.Validate(config)
```
fedepaol: we should do this before filtering imo

fedepaol (changes requested):

Dockerfile.grout
zeeke: raised: 
- https://github.com/openperouter/openperouter/issues/538

internal/controller/routerconfiguration/host_config.go
zeeke: moving the interface to reconcile.go, as it's where it is consumed.

internal/controller/routerconfiguration/host_config.go
zeeke: I'm not sure I get what you mean with "they don't overlap".

They are both consumed by the Reconcile function, which ask the interface which API are supported by the datapath configurator (Validate) and to apply the configuration (Configure). 
Splitting the interface would make the Reconcile function accept two arguments instead of one, allowing that to be called with inconsistent parameters (e.g., KernelValidate and GroutConfigure).

Am I missing something?

internal/grout/passthrough.go
rjarry: There is no maximum number for `net_tapXXX`. In fact, it does not need to be a number. It can be whatever string you want, as long as it fits into `GR_PORT_DEVARGS_SIZE` (128 characters).

It is completely independent from the Linux *netdev* name (which can be configured with the `iface=` devargs parameter). By default, the DPDK tap driver will set a literal `iface=dtap%d` which gets interpreted by Linux as "take the next X where dtapX does not exist yet".

Unfortunately, if we don't explicitly set `iface=`, there is no way to know the *netdev* name in advance. If you need to know it, you must set it.

internal/grout/passthrough.go
zeeke: I can make that function delegate to `MoveInterfaceToNamespace`.
Calling directly from `ensureTapPortInHostNamespace` makes the call site complexity hit the linter limit.

I'll propose a refactoring of the `MoveInterfaceToNamespace` in another PR, to make call site simpler

internal/grout/passthrough.go
zeeke: sounds good. 
`HostLink` is not a good name in my opinion, it'd refer a link in the host and references would become `params.HostLink.HostIPv4`, `params.HostLink.NSIPv4`, ..

What about `LinkPair`: `params.LinkPair.HostIPv4`, `params.LinkPair.NSIPv4` ?

internal/controller/routerconfiguration/host_config.go
fedepaol: see my comment about webhooks above. There's no reason to have a single interface representing both behaviors, as the validation part is totaly separated from the configuration part, because of this and for the convenience of not having to carry the configuration side of the interface to the webhook package, I think it's sane to split this in two.

internal/grout/passthrough.go
zeeke: What about the IP addresses that need to be deleted? should assignIPsToGrout delete them or just add ip address like AssignIPsToInterface?

internal/grout/passthrough.go
fedepaol: oof naming is hard. CrossNsIPs? It's ips after all

internal/grout/passthrough.go
fedepaol: what's _that function_? 

why can't we use MoveInterfaceToNamespace directly and get rid of moveLinkToHostNamespace?

internal/grout/passthrough.go
zeeke: I'd go with a random alphanumeric char of 8 character.s Looks simple and robust to me

internal/grout/passthrough.go
fedepaol: ah right, missed that part. Let's call this ensureGroutSideIPs or something like that then.

As a side note, why we don't have the problem of delting old ips on the host side?

internal/grout/passthrough.go
zeeke: this is a preview of how it would look like:
https://github.com/zeeke/openperouter/pull/10/changes/f373157ea3e5faf4ef2f34796cb45c9292345ece

That field also has JSON marker, so we have to pay a little more attention to renaming.

other suggestions:
- AddressPair `json:"addresspair"`
- AddrPair `json:"addrpair"`

internal/grout/passthrough.go
zeeke: that function = `moveLinkToHostNamespace`

>why can't we use MoveInterfaceToNamespace directly and get rid of moveLinkToHostNamespace?

because calling MoveInterfaceToNamespace needs more params, a scaffolding like this at minimum:
```golang
	srcNsHandle, err := netlink.NewHandleAt(srcNS)
	if err != nil {
		return fmt.Errorf("SetupUnderlay: failed to get netlink handle for namespace %s: %w", srcNS.String(), err)
	}
	defer srcNsHandle.Close()

	defaultNetNS, err := netns.Get()
	if err != nil {
		return fmt.Errorf("SetupUnderlay: failed to get netns handle for default namespace: %w", err)
	}
	defer func() {
		if err := defaultNetNS.Close(); err != nil {
			slog.Error("failed to close default namespace", "error", err)
		}
	}()

	defaultNetNSHandle, err := netlink.NewHandleAt(defaultNetNS)
	if err != nil {
		return fmt.Errorf("SetupUnderlay: failed to get netlink handle for default namespace: %w", err)
	}
	defer defaultNetNSHandle.Close()

   	if err := hostnetwork.MoveInterfaceToNamespace(ctx, name, srcNsHandle, defaultNetNSHandle, srcNS, 0); err != nil {
		return fmt.Errorf("failed to move interface %s to host namespace: %w", name, err)
	}
```

all the if and defer if breaks the linter complexity limit.

charts/openperouter/templates/controller.yaml
zeeke: sounds good, i'll takle this deployment test (smoke tests ??) in a subsequent PR

fedepaol: just file an issue and for now fill a full fledged ci lane (as we don't want to merge stuff that we are not sure it's working)

zeeke: opened
- https://github.com/openperouter/openperouter/issues/541

internal/grout/underlay.go
zeeke: yes, this function does the same.

Renamed to `RestoreUnderlay`

cmd/nodemarker/main.go
```diff
@@ -102,6 +104,8 @@ func main() {
 		"The service name used to generate the TLS cert's hostname")
 	flag.IntVar(&args.webhookPort, "webhook-port", 9443, "the port of the webhook service")
 	flag.StringVar(&args.webhookMode, "webhookmode", WebhookModeEnabled, "webhook mode: disabled, enabled, or webhookonly")
+	flag.BoolVar(&args.groutEnabled, "grout-enabled", false,
```
fedepaol: can we have the enum here too?

cmd/nodemarker/main.go
```diff
@@ -250,11 +254,15 @@ func setupCertRotation(notifyFinished chan struct{}, mgr manager.Manager, logger
 	return nil
 }
 
-func setupWebhook(mgr manager.Manager, logger *slog.Logger) error {
+func setupWebhook(mgr manager.Manager, logger *slog.Logger, groutEnabled bool) error {
 	logger.Info("webhooks enabled")
 
 	webhooks.Logger = logger
 	webhooks.WebhookClient = mgr.GetAPIReader()
+	webhooks.DatapathConfigValidator = &conversion.KernelDatapathConfigValidator{}
```
fedepaol: same here, or even better let's pass the validator from outside

cmd/nodemarker/main.go
zeeke: done

internal/grout/passthrough.go
fedepaol: makes sense. And I forgot, this whole ballet is really distracting :/ 
Thanks for the explanation

fedepaol:
there are still unaddressed comments I guess?

operator/api/v1alpha1/openperouter_types.go
```diff
@@ -58,6 +58,10 @@ type OpenPERouterSpec struct {
 	// +kubebuilder:validation:Minimum=1
 	// +kubebuilder:validation:Maximum=65535
 	HealthProbePort *int32 `json:"healthProbePort,omitempty"`
+	// groutEnabled enables the Grout DPDK dataplane sidecar in the router pods.
+	// When enabled, FRR uses the dplane_grout module for accelerated forwarding. (default: false)
+	// +optional
+	GroutEnabled *bool `json:"groutEnabled,omitempty"`
 }
```
zeeke, line 66: @maiqueb @fedepaol I guess you also want move to Datapath: Kernel/Grout here too. right?

maiqueb: yes - at least I'd like to.

zeeke:
> there are still unaddressed comments I guess?

waiting for the CI to be green. the main open point is https://github.com/openperouter/openperouter/pull/338/changes/7179c6c88c27ee649cf1edfed30943f02038195b :
 are we sure about the rename `veth --> crossNsIp` among field and json fields?

zeeke:
> > there are still unaddressed comments I guess?
> 
> waiting for the CI to be green. the main open point is [7179c6c](https://github.com/openperouter/openperouter/commit/7179c6c88c27ee649cf1edfed30943f02038195b) : are we sure about the rename `veth --> crossNsIp` among field and json fields?

CI is green
@maiqueb @rjarry @fedepaol please take another look when you have time. 
I'm still not 10% convinced about `veth --> crossNsIp` renaming, but if it's ok for you I can live with that

e2etests/tests/dump.go
```diff
@@ -42,7 +42,7 @@ func dumpIfFails(cs clientset.Interface, additionalNamespaces ...string) {
 	}
 }
 
-func dumpFRRInfo(basePath, testName string, cs clientset.Interface, clabContainers ...string) {
+func dumpFRRInfo(basePath, testName string, cs clientset.Interface, groutMode bool, clabContainers ...string) {
```
fedepaol: nit, can this be an enum too?

e2etests/tests/hostconfiguration.go
```diff
@@ -1357,19 +1357,19 @@ var _ = ginkgo.Describe("Router Host configuration", func() {
 })
 
 type l3passthroughParams struct {
-	TargetNS string `json:"targetns"`
-	HostVeth *veth  `json:"veth"`
+	TargetNS string      `json:"targetns"`
+	HostVeth *crossNsIPs `json:"crossNsIPs"`
```
fedepaol: nit: the field is still called hostVeth while it might not be a veth? hostLinkIPs? I don't know.
Naming this was hard the first time, it's becoming harder now that it can be two different things

internal/grout/passthrough.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	// Delete the grout port so it can be recreated on the next setup.
+	portName := hostnetwork.PassthroughNames.NamespaceSide
+	if err := client.deletePort(ctx, portName); err != nil {
+		return fmt.Errorf("RemovePassthrough: failed to delete grout port %s: %w", portName, err)
+	}
+	return nil
+}
+
+func makeTapRandomString() string {
```
fedepaol: nit, but only if you need to recommit, this should go below

internal/grout/underlay.go
```diff
@@ -0,0 +1,313 @@
...
+		if err := client.deleteAddress(ctx, underlayPortName, addr); err != nil {
+			return fmt.Errorf("failed to delete addresses on grout port %s: %w", underlayPortName, err)
+		}
+
+	}
+
+	return nil
+}
+
+func connectedRouteForAddress(ifaceName, addr string) (*netlink.Route, error) {
```
fedepaol: nit, this can go below

fedepaol (commented):
Just a few nits, and I agree, the link name is not the best... let's try to think if we can come up with a better one

e2etests/tests/hostconfiguration.go
maiqueb: yes, this name is pretty bad. 

@fedepaol suggestion is good enough IMHO - the struct is indeed for representing the host link - and it holds IPs.

e2etests/tests/dump.go
maiqueb: please... or an option ... (have arbitrary "dump" instructions in the dumping machine, have `withGrout` which adds dumpGrout() to the arbitrary slice)

e2etests/tests/hostconfiguration.go
zeeke: it took me a while to understand how these tests work, and I'm not getting why these structs have been used instead of `internal/hostnetwork/vni.go` ones.

If it's because that's an "internal" package to import, consider that the hostconfiguration.go file relies on a binary which comes from that package (`internal/hostnetwork/external_suite_test.go`) and makes strong assumptions about its content (e.g. Ginkgo test names).

Meanwhile, I'm renaming to CrossNsIPs

e2etests/tests/dump.go
zeeke: I refactored dumpFRRInfo
https://github.com/openperouter/openperouter/pull/338/changes/4438f7e3a2bee7399c58db0ec0363f325fc9a578

As it already had variadic arguments that prevented using the options pattern directly.

internal/controller/routerconfiguration/grout_config.go
maiqueb: Can we resolve this comment @zeeke / @fedepaol ?

internal/grout/grout_client.go
maiqueb: wait, I still don't see this in the code. 

Why was it resolved @fedepaol ¿

internal/grout/underlay.go
maiqueb: I think we can resolve this comment.

internal/grout/underlay.go
maiqueb: ditto

maiqueb (commented):
Nearly there. 

I just don't know if you've acted on a previous comment, then, from my perspective, LGTM.

internal/grout/grout_client.go
fedepaol: I don't know anymore, probably a mistake on my side. And I agree, @zeeke this makes a lot of sense

internal/controller/routerconfiguration/grout_config.go
zeeke: yes

internal/grout/grout_client.go
zeeke: runOutput function adds `--json` to each commands:

https://github.com/openperouter/openperouter/pull/338/changes#diff-fa94e2eb2ceb54a6ed841cdb0aae8d785841f1f5301411cdf770f7b332f563c4R165

```golang
// runOutput executes a grcli command and returns stdout and any error.
func (c *Client) runOutput(ctx context.Context, args ...string) (string, error) {
	cmdArgs := append([]string{"--err-exit", "--json", "--socket", c.socketPath}, args...)
    ...
```

when grout v0.17.0 will be release, we'll also have json formatted errors, which  wil improve this file further.

maiqueb: ah, OK. didn't expect it to be defined within the helper.

maiqueb (approved):
nice.

internal/grout/grout_client.go
fedepaol: can we file an issue to avoid forgetting this?

zeeke: right!
- https://github.com/openperouter/openperouter/issues/553

fedepaol:
@zeeke mind fixing the conflicts?

maiqueb:
@zeeke tests seem to be failing. 

Unless something major changes while fixing them, I am OK with the merge. 

Thanks for this very nice feature.

fedepaol:
Thanks!

PR #339, opened 2026-04-24
fix(e2e): handle neighbor entry GC race in bridge refresh test

fedepaol:

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
/kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

The IPv6 migration test timed out waiting for a STALE neighbor entry
because the kernel NDP state machine (STALE->DELAY->FAILED->GC) completed
in ~8 seconds, faster than the test polling interval via podman exec.

Introduce ErrNoNeighborEntry sentinel so callers can distinguish
"entry was GC'd" from other failures, and treat a missing entry as
success. Also verify the neighbor entry exists before pod deletion
to confirm it was learned and not simply never created.

Spotted in https://github.com/openperouter/openperouter/actions/runs/24882543505/job/72855481022?pr=310

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/tests/bridgerefresh.go
+++ b/e2etests/tests/bridgerefresh.go
@@ -4,6 +4,7 @@ package tests
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -295,6 +296,11 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
 					return checkType2RouteExists(cs, migratingPodIPOnly, vtepIPOnly, l2VNI)
 				}, 3*time.Minute, 5*time.Second).ShouldNot(HaveOccurred())
 
+				By("Verifying neighbor entry exists on the router before migration")
+				Eventually(func() error {
+					return checkNeighborExists(cs, migratingPodIPOnly, l2VNI, migratingPod.Spec.NodeName)
+				}, time.Minute, 2*time.Second).ShouldNot(HaveOccurred())
+
 				By("Deleting migrating pod (simulating pod eviction/migration)")
 				err = cs.CoreV1().Pods(testNamespace).Delete(
 					context.Background(),
@@ -316,7 +322,11 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
 				// This is important to trigger the deadlock situation described in https://github.com/FRRouting/frr/issues/14156
 				By("Waiting for neighbor entry to go STALE on the router on the same node")
 				Eventually(func() error {
-					return checkNeighborStale(cs, migratingPodIPOnly, l2VNI, migratingPod.Spec.NodeName)
+					err := checkNeighborStale(cs, migratingPodIPOnly, l2VNI, migratingPod.Spec.NodeName)
+					if errors.Is(err, ErrNoNeighborEntry) { // if the neighbor entry is expired no need to wait
+						return nil
+					}
+					return err
 				}, 3*time.Minute, 2*time.Second).ShouldNot(HaveOccurred())
 
 				By("Recreating migrating pod on node B (the other node) with same IP")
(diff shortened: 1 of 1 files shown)


e2etests/tests/bridgerefresh.go
```diff
@@ -316,7 +322,11 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
 				// This is important to trigger the deadlock situation described in https://github.com/FRRouting/frr/issues/14156
 				By("Waiting for neighbor entry to go STALE on the router on the same node")
 				Eventually(func() error {
-					return checkNeighborStale(cs, migratingPodIPOnly, l2VNI, migratingPod.Spec.NodeName)
+					err := checkNeighborStale(cs, migratingPodIPOnly, l2VNI, migratingPod.Spec.NodeName)
+					if errors.Is(err, ErrNoNeighborEntry) { // if the neighbor entry is expired no need to wait
+						return nil
+					}
+					return err
 				}, 3*time.Minute, 2*time.Second).ShouldNot(HaveOccurred())
```
maiqueb, line 330: ok, got it. 

can't we move this logic to the matcher instead ? like, the error should either: not have occurred, *or* match `ErrNoNeighborEntry` ? 

I think it would read better.

e2etests/tests/bridgerefresh.go
```diff
@@ -295,6 +296,11 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
 					return checkType2RouteExists(cs, migratingPodIPOnly, vtepIPOnly, l2VNI)
 				}, 3*time.Minute, 5*time.Second).ShouldNot(HaveOccurred())
 
+				By("Verifying neighbor entry exists on the router before migration")
+				Eventually(func() error {
+					return checkNeighborExists(cs, migratingPodIPOnly, l2VNI, migratingPod.Spec.NodeName)
+				}, time.Minute, 2*time.Second).ShouldNot(HaveOccurred())
```
maiqueb, line 302: isn't should succeed the same ? I think it reads better.

e2etests/tests/bridgerefresh.go
```diff
@@ -370,6 +380,28 @@ func checkType2RouteExists(cs clientset.Interface, podIP, vtepIP string, vni int
...
+	exec, err := openperouter.ExecutorForNode(currentRouters, nodeName)
+	if err != nil {
+		return err
+	}
+	out, err := exec.Exec("ip", "neigh", "show", podIP, "dev", bridgeDev)
+	if err != nil {
+		return fmt.Errorf("failed to check neighbor on router %s: %w", exec.Name(), err)
+	}
+	if strings.TrimSpace(out) == "" {
+		return fmt.Errorf("no neighbor entry for %s on %s in router %s", podIP, bridgeDev, exec.Name())
```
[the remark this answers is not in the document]
maiqueb: this could be interesting.

e2etests/tests/bridgerefresh.go
```diff
@@ -386,11 +418,10 @@ func checkNeighborStale(cs clientset.Interface, podIP string, vni int, nodeName
 	}
 	out = strings.TrimSpace(out)
 	if out == "" {
-		return fmt.Errorf("no neighbor entry for %s on %s in router %s", podIP, bridgeDev, exec.Name())
+		return fmt.Errorf("%w for %s on %s in router %s", ErrNoNeighborEntry, podIP, bridgeDev, exec.Name())
```
maiqueb, line 421: ditto

maiqueb (approved):
tks

e2etests/tests/bridgerefresh.go
maiqueb: > ok, got it.
> 
> can't we move this logic to the matcher instead ? like, the error should either: not have occurred, _or_ match `ErrNoNeighborEntry` ?
> 
> I think it would read better.

FWIW, I mean something like:


```golang
				By("Waiting for neighbor entry to go STALE on the router on the same node")
				Eventually(func() error {
					return checkNeighborStale(cs, migratingPodIPOnly, l2VNI, migratingPod.Spec.NodeName)
				}, 3*time.Minute, 2*time.Second).Should(
                    Or(
                        Not(HaveOccurred()),
                        MatchError(func(err error) bool {return errors.Is(err, ErrNoNeighborEntry)} ),
                    ),
                )
```

fedepaol:
closing as I don't remember this anymore. Will start from scratch if I we are hitting hte bug again

PR #159, opened 2025-11-06
Status CRD

phoracek:
**Is this a BUG FIX or a FEATURE ?**:

/kind feature

**What this PR does / why we need it**:

This PR introduces a status reporting system for OpenPERouter that provides real-time visibility into the configuration state of router resources on each node. The implementation adds a new RouterNodeConfigurationStatus CRD and associated infrastructure to track and report the health of Underlay, L2VNI, L3VNI, and L3Passthrough resource configurations.

Key components added:

1. **RouterNodeConfigurationStatus CRD**: A new Kubernetes resource that provides per-node status information including:
   - Failed resource details (kind, name, and error messages)
   - Kubernetes standard conditions (Ready/Degraded)
   - Last update timestamps
   - Node owner references for lifecycle management

2. **Centralized Status Management**: A new `StatusManager` component that:
   - Implements both `StatusReporter` and `StatusReader` interfaces
   - Maintains a thread-safe cache of failed resources
   - Uses controller-runtime event channels for real-time status updates
   - Provides aggregated status summaries for controllers

3. **Integration with Resource Validation**: Status reporting is integrated directly into validation methods for:
   - Underlay BGP session configuration
   - L2VNI bridge and gateway setup
   - L3VNI VRF and routing configuration
   - L3Passthrough route management

4. **Testing**: Full e2e test coverage including:
   - Automatic RouterNodeConfigurationStatus resource creation
   - Resource lifecycle management with owner references
   - Success and failure scenario validation
   - Status utility functions for easy test integration

This enhancement addresses the need for operators to have clear visibility into the configuration state of OpenPERouter resources, enabling better troubleshooting and monitoring of BGP/EVPN network configurations across Kubernetes clusters.

Implements #20.

**Special notes for your reviewer**:

I am now refining the last commit, to make sure it addressed all the comments.

After that I will update the website.

**Release note**:

```release-note
Introduce a new CR capturing the configuration status of each resource per node
```

PR #345, opened 2026-04-28
POC configuration resilience with per-resource quarantine

RamLavi:
kind/enhancement

Implement configuration resilience with per-resource quarantine, as described in the enhancement proposal:
https://github.com/openperouter/openperouter/pull/225

Today, a single invalid resource causes the entire reconcile loop to fail and tears down all configuration, including valid resources. This PR changes the controller to validate and process resources independently:

- Invalid resources are quarantined and reported
- Valid resources continue to be configured (FRR + host)

Behavior details:

- L3VNI validation failure:
  - Entire VRF is quarantined
  - Dependent L2VNIs are marked `DependencyFailed` and not configured

- L2VNI validation failure:
  - Isolated to the resource
  - Does not affect other L2VNIs

- Additional controller-only validations:
  - Cross-type VNI uniqueness
  - VRF dependency tracking

- Overlay teardown:
  - Scoped per-VRF
  - Failures (e.g. netlink errors) do not affect other VRFs

- Status reporting:
  - All failures are exposed via `RouterNodeConfigurationStatus`
  - Includes Ready/Degraded conditions and per-resource failure details

This improves robustness and prevents unrelated configuration from being disrupted by a single bad resource.

Depends on:
- https://github.com/openperouter/openperouter/pull/159 (Status CRD)
  - The `[DNM]` commit in this PR is a temporary copy and should be dropped once #159 merges

Related:
- https://github.com/openperouter/openperouter/pull/332 (L2VNI webhook validation)

- First commit (`[DNM]`) is temporary and will be removed after #159 merges

Test plan:

- [x] E2E: cross-type VNI conflict → L2VNI quarantined with `ValidationFailed`
- [x] E2E: invalid RT on L3VNI → VRF quarantined, L2VNI gets `DependencyFailed`
- [x] E2E: orphan L2VNI → `DependencyFailed`
- [x] E2E: recovery → fix issue, status clears, node becomes Ready
- [x] Existing e2e suite passes (no regressions)

**Release note**:

```release-note
Introduce per-resource configuration resilience with quarantine semantics.

Invalid L3VNI resources now quarantine their VRF and block dependent L2VNIs.
Invalid L2VNIs are isolated and do not affect other resources.

All failures are reported via RouterNodeConfigurationStatus with detailed conditions.

### CODE REVIEW

File: internal/controller/routerconfiguration/reconcile.go [hunk 1]
```diff
@@ -6,53 +6,147 @@ import (
...
-	}
+	validL3VNIs, validVRFs := validateL3VNIsWithQuarantine(apiConfig.L3VNIs, usedVNIs, quarantinedVRFs, &result)
+	validL2VNIs := validateL2VNIsWithQuarantine(apiConfig.L2VNIs, usedVNIs, quarantinedVRFs, validVRFs, &result)
+	validPassthrough := validatePassthroughsWithQuarantine(apiConfig.L3Passthrough, &result)
 
-	if err := conversion.ValidateVRFs(apiConfig.L2VNIs, apiConfig.L3VNIs); err != nil {
-		return fmt.Errorf("failed to validate VRFs: %w", err)
+	if err := conversion.ValidateVRFs(validL2VNIs, validL3VNIs); err != nil {
+		return result, fmt.Errorf("failed to validate VRFs: %w", err)
 	}
```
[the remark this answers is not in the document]
RamLavi: ValidateVRFs runs on the already-filtered validL3VNIs/validL2VNIs — resources that passed per-resource quarantine. My thinking is - if valid resources still trigger a VRF conflict (e.g. subnet overlap), that's a systemic configuration problem, not an isolated bad resource.

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -6,53 +6,147 @@ import (
...
 
-	if err := conversion.ValidatePassthroughs(apiConfig.L3Passthrough); err != nil {
-		return fmt.Errorf("failed to validate l3passthrough: %w", err)
+	if err := conversion.ValidateHostSessions(validL3VNIs, validPassthrough); err != nil {
+		return result, fmt.Errorf("failed to validate host sessions: %w", err)
 	}
```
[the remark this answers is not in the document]
RamLavi: Same logic: this validates the valid set. Host session conflicts between two valid L3VNIs would mean the inputs are inherently contradictory. This shouldn't be a problem

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -6,53 +6,147 @@ import (
...
+
+func validatePassthroughsWithQuarantine(passthroughs []v1alpha1.L3Passthrough, result *ReconcileResult) []v1alpha1.L3Passthrough {
+	if err := conversion.ValidatePassthroughs(passthroughs); err != nil {
+		for _, pt := range passthroughs {
+			result.AddFailure("L3Passthrough", pt.Name, v1alpha1.ValidationFailed, err.Error())
+		}
+		return nil
+	}
+	return passthroughs
 }
```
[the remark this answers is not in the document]
RamLavi: Passthrough validation is global by nature (e.g. interface name uniqueness), so it's not clear which specific resource to quarantine when there's a conflict. 
I think it's good enough for the PoC, we can refine later if a maintainer thinks otherwise.

RamLavi:
gemini-code-assist please comment back on the comments I replied on

RamLavi:
> Hello @RamLavi, I am happy to help! Could you please point me to the specific review comments you are referring to? Once you provide the links or the context, I will be able to review them and provide further feedback.

I mean all of the ones you commented on this PR, and I commented back.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/b5e68722140b6ff0b2459a56a23f61ee13aa7470..c21c4e1162e055650e451f704546cba66409fb24): Fix rebase conflict

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/c21c4e1162e055650e451f704546cba66409fb24..ece76b783f06b4e80aefe413d4df4952c1e5f36b): Refactor commit for commitlint

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/ca338a6131b07ca0df7177a41348a5d615c5d7c0..00a9e0ae98d44f13c462f15f941af51a868a2d44): Rebase

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/00a9e0ae98d44f13c462f15f941af51a868a2d44..07d1733e771fe6b8a42c266a73281ebcf0451ec5): Added missing rbac

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/07d1733e771fe6b8a42c266a73281ebcf0451ec5..4e7ebabfbd3b986c5c74f96c3020d5f70d4c156a): dynamically get webhooks

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/4e7ebabfbd3b986c5c74f96c3020d5f70d4c156a..98f447af3d3041f84a6b69b25a029cbd90e86d0d): Add commit to remove VRF and gateway IPs from disconnected L2VNI on existing tests

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/98f447af3d3041f84a6b69b25a029cbd90e86d0d..ce1e3f0dd0174804e307fe14e075011bee9ebf0d): Remove L3VNI from another unrelated tests that doesn't require it, and now is failing our current logic.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/ce1e3f0dd0174804e307fe14e075011bee9ebf0d..ac07fe43c05c24fd6efa5ee415755df44cae7fc1): Move stuff from commit to commit to improve readability

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/ac07fe43c05c24fd6efa5ee415755df44cae7fc1..f0375a6f959d8514936dacd94cdcecb0951e8167): Add MyNamespace to host-mode reconciler, so that ensureStatusCR won't fail on empty search

internal/controller/routerconfiguration/reconcile.go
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist is that answer OK for you ?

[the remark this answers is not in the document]
maiqueb: @RamLavi I think I agree with the bot. The goal of the enhancement is to have these VRFs quarantined, to prevent the whole resource reconciliation error. 

Or did I miss something ?

internal/controller/routerconfiguration/reconcile.go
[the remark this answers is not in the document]
maiqueb: passthroughs - like the overlays - are limited to a single CR instance.

They can't "conflict" with another resource. What can happen though, is they can be misconfigured.

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,163 @@
...
+	"github.com/openperouter/openperouter/api/v1alpha1"
+)
+
+const (
+	ConditionTypeReady    = "Ready"
+	ConditionTypeDegraded = "Degraded"
+
+	ReasonConfigurationSuccessful = "ConfigurationSuccessful"
+	ReasonConfigurationFailed     = "ConfigurationFailed"
+	ReasonUnderlayFailed          = "UnderlayFailed"
```
maiqueb, line 41: we'll probably require also the same thing but for passthrough, right ?

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -54,15 +54,23 @@ func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, underlay
...
-	if err := configureInterfaces(ctx, interfacesConfiguration{
+	ifConfig := interfacesConfiguration{
 		targetNamespace:    targetNamespace,
 		APIConfigData:      validConfig,
 		nodeIndex:          nodeIndex,
 		underlayFromMultus: underlayFromMultus,
-	}); err != nil {
+	}
+	hostConfig, err := setupHostPrerequisites(ctx, ifConfig)
+	if err != nil {
```
maiqueb, line 65: maybe use the underlay list itself as an attribute ?

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -54,15 +54,23 @@ func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, underlay
...
 	}
 
+	if len(validConfig.Underlays) > 0 {
+		if err := provisionOverlays(ctx, ifConfig, hostConfig); err != nil {
+			return result, fmt.Errorf("failed to provision overlays: %w", err)
+		}
+	}
+
 	return result, nil
 }
```
maiqueb, line 71: I really dislike this function's signature - especially its return. 

Could you add a struct that would "join" both return types ? then the caller would get the rightful attribute from the struct.

internal/controller/routerconfiguration/status_test.go
```diff
@@ -0,0 +1,117 @@
...
+			expectedReadyReason:  ReasonConfigurationFailed,
+			expectedReadyMessage: "2 resources failed (L2VNI/orphan, L3VNI/bad-vni)",
+		},
+		{
+			name:                 "reconcile hard error",
+			result:               ReconcileResult{},
+			reconcileErr:         fmt.Errorf("failed to reload frr config: exit status 1"),
+			expectedReady:        metav1.ConditionFalse,
+			expectedDegraded:     metav1.ConditionTrue,
+			expectedReadyReason:  ReasonConfigurationFailed,
```
maiqueb, line 76: I am not sure, but maybe we want to have `ReasonRouterConfigurationFailed` ? 

Or do you think it is obvious ?

internal/controller/routerconfiguration/status_test.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"fmt"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openperouter/openperouter/api/v1alpha1"
+)
+
+func findCondition(conditions []metav1.Condition, condType string) *metav1.Condition {
```
maiqueb: this should go in the bottom of the file

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,163 @@
...
+			Type:               ConditionTypeDegraded,
+			Status:             metav1.ConditionTrue,
+			Reason:             ReasonConfigurationFailed,
+			Message:            reconcileErr.Error(),
+			LastTransitionTime: now,
+		})
+		return
+	}
+
+	if !result.HasFailures() && !result.HasFailure("Underlay") {
```
maiqueb: please derive this name from the underlay CRD Kind. Can that be done ? 

If not, I'd say use a const. 

And, I guess we need something similar for passthrough.

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	if err := cl.Status().Update(ctx, status); err != nil {
+		return fmt.Errorf("failed to update RouterNodeConfigurationStatus %s: %w", nodeName, err)
+	}
+
+	return nil
+}
+
+func setConditions(status *v1alpha1.RouterNodeConfigurationStatus, result ReconcileResult, reconcileErr error) {
```
maiqueb: would it make sense to receive the timestamp as input ? 

that might simplify unit testing.

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,163 @@
...
+	oldStatus := status.Status.DeepCopy()
+
+	status.Status.FailedResources = result.FailedResources
+	setConditions(status, result, nil)
+
+	if equality.Semantic.DeepEqual(*oldStatus, status.Status) {
+		return nil
+	}
+
+	if err := cl.Status().Update(ctx, status); err != nil {
```
maiqueb: shouldn't we do this in a `retry` block ?

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,163 @@
...
+func updateStatus(ctx context.Context, cl client.Client, nodeName, namespace string, result ReconcileResult) error {
+	status := &v1alpha1.RouterNodeConfigurationStatus{}
+	if err := cl.Get(ctx, client.ObjectKey{Name: nodeName, Namespace: namespace}, status); err != nil {
+		return fmt.Errorf("failed to get RouterNodeConfigurationStatus %s: %w", nodeName, err)
+	}
+
+	oldStatus := status.Status.DeepCopy()
+
+	status.Status.FailedResources = result.FailedResources
+	setConditions(status, result, nil)
```
maiqueb: I think it would be neater if you had a function for creating a status CR from a `ReconcileResult`.

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,163 @@
...
+			Namespace: namespace,
+		},
+	}
+	if err := cl.Create(ctx, status); err != nil {
+		return fmt.Errorf("failed to create RouterNodeConfigurationStatus %s: %w", nodeName, err)
+	}
+	return nil
+}
+
+func updateStatus(ctx context.Context, cl client.Client, nodeName, namespace string, result ReconcileResult) error {
```
maiqueb, line 58: nit: turn this into a builder pattern, and provide a `withError` option ...

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,163 @@
...
+const (
+	ConditionTypeReady    = "Ready"
+	ConditionTypeDegraded = "Degraded"
+
+	ReasonConfigurationSuccessful = "ConfigurationSuccessful"
+	ReasonConfigurationFailed     = "ConfigurationFailed"
+	ReasonUnderlayFailed          = "UnderlayFailed"
+)
+
+func ensureStatusCR(ctx context.Context, cl client.Client, nodeName, namespace string) error {
```
maiqueb, line 44: this function is raceful IIUC. between toc (time of check) and tou (time of use - which in this instance is a create) something else might happen - e.g. the status be created. Can that happen ?...

If so, the CREATE would fail. Is that OK ? I think it could be, since at the end of the day, this function is about ensuring the CR exists. 

Maybe worth changing it to "try to create ; swallow the already exists error; return error that might be nil".

internal/controller/routerconfiguration/reconcile_result.go
```diff
@@ -0,0 +1,57 @@
...
+	r.FailedResources = append(r.FailedResources, v1alpha1.FailedResource{
+		Kind:    kind,
+		Name:    name,
+		Reason:  reason,
+		Message: msg,
+	})
+	if r.failedKinds == nil {
+		r.failedKinds = sets.New[string]()
+	}
+	r.failedKinds.Insert(kind)
```
maiqueb, line 53: this is because we treat underlay (and probably passthrought errors differently, right ?

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -10,29 +10,31 @@ import (
 	"github.com/openperouter/openperouter/internal/frr"
 )
 
-func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, underlayFromMultus bool, nodeIndex int, logLevel, frrConfigPath, targetNamespace string, updater frr.ConfigUpdater) error {
+func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, underlayFromMultus bool, nodeIndex int, logLevel, frrConfigPath, targetNamespace string, updater frr.ConfigUpdater) (ReconcileResult, error) {
+	var result ReconcileResult
+
 	if err := conversion.ValidateUnderlays(apiConfig.Underlays); err != nil {
-		return fmt.Errorf("failed to validate underlays: %w", err)
+		return result, fmt.Errorf("failed to validate underlays: %w", err)
```
maiqueb: nit: can we be explicit here and return a newly created empty result ? 

When I read this code, I kept looking for "when is the result actually filled" ... And did not find it, which was confusing.

internal/controller/routerconfiguration/static_configuration_controller.go
```diff
@@ -74,7 +74,7 @@ func (r *StaticConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	updater := frrconfig.UpdaterForSocket(r.FRRReloadSocket, r.FRRConfigPath)
 
-	err = Reconcile(ctx, apiConfig, false, r.NodeIndex, r.LogLevel, r.FRRConfigPath, targetNS, updater)
+	_, err = Reconcile(ctx, apiConfig, false, r.NodeIndex, r.LogLevel, r.FRRConfigPath, targetNS, updater)
```
maiqueb: I assume this `result` will be used for something in a follow-up commit, right ?

internal/conversion/validate_vni.go
```diff
@@ -40,11 +40,29 @@ func ValidateL3VNIsForNodes(nodes []corev1.Node, underlays []v1alpha1.L3VNI) err
 	return nil
 }
 
-// ValidateL3VNIs runs L3VNI specific validation.
+// ValidateL3VNI validates a single L3VNI's fields (VRF name, route targets).
+func ValidateL3VNI(l3Vni v1alpha1.L3VNI) error {
+	vni := vnisFromL3VNIs([]v1alpha1.L3VNI{l3Vni})[0]
```
maiqueb: this call is confusing. What is this doing ? Is it applying some defaults ?... Why don't we handle the l3vni directly ?

internal/conversion/validate_vni.go
```diff
@@ -63,34 +81,43 @@ func ValidateL2VNIsForNodes(nodes []corev1.Node, underlays []v1alpha1.L2VNI) err
...
-		if vni.Spec.HostMaster != nil {
-			if err := validateHostMaster(vni.Name, vni.Spec.HostMaster); err != nil {
-				return err
-			}
-		}
-		if len(vni.Spec.L2GatewayIPs) > 0 && !hasVRF(vni) {
-			return fmt.Errorf("l2gatewayips cannot be set without spec.vrf for vni %q", vni.Name)
+	if err := ValidateRouteTargets(vni); err != nil {
+		return fmt.Errorf("invalid route targets for vni %s: %w", vni.name, err)
+	}
```
maiqueb, line 93: hm, I'm a bit confused here ... I don't think an L2 VNI has RTs defined to import / export. This should be gone IIUC.

internal/conversion/validate_vni_test.go
```diff
@@ -13,6 +13,102 @@ import (
 	"k8s.io/utils/ptr"
 )
 
+func TestValidateL3VNI(t *testing.T) {
+	tests := []struct {
+		name    string
+		vni     v1alpha1.L3VNI
+		wantErr bool
```
maiqueb: would it make sense to have the attribute type as `error` ? that way we could match the actual error.

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -67,3 +72,71 @@ func underlayName(apiConfig conversion.APIConfigData) string {
...
+			continue
+		}
+		usedVRFs[l3.Spec.VRF] = l3.Name
+		usedVNIs[l3.Spec.VNI] = "L3VNI/" + l3.Name
+		valid = append(valid, l3)
+	}
+	return valid, usedVRFs
+}
+
+func validateL2VNIsWithQuarantine(l2VNIs []v1alpha1.L2VNI, usedVNIs map[uint32]string, quarantinedVRFs sets.Set[string], validVRFs map[string]string, result *ReconcileResult) []v1alpha1.L2VNI {
```
maiqueb: can we maybe simplify this signature ? could you pass as input a list of VNIs that we know are *not* quarantined ? that'd save an input parameter to this function, and also simplify it (we wouldn't need to check if it is quarantined here).

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -66,7 +66,9 @@ func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, underlay
 	}
 
 	if len(validConfig.Underlays) > 0 {
-		if err := provisionOverlays(ctx, ifConfig, hostConfig); err != nil {
+		overlayResult, err := provisionOverlays(ctx, ifConfig, hostConfig)
+		result.Merge(overlayResult)
```
maiqueb: ok we merge before checking the error on purpose - so we can bring along the failed kinds, and failed VNIs / underlay / passthroughs. 

Right ?

e2etests/tests/evpn_l2.go
```diff
@@ -381,10 +381,8 @@ var _ = Describe("Routes between bgp and the fabric - vtepInterface", func() {
 		// advertised by the network fabric
 		redistributeConnectedForLeafKind(nodes)
 
-		l2VniRedWithGateway := l2VniRed.DeepCopy()
-		l2VniRedWithGateway.Spec.VRF = ptr.To("red")
-		l2VniRedWithGateway.Spec.L2GatewayIPs = []string{"192.171.24.1/24"}
-		l2VniRedWithGateway.Spec.HostMaster = &v1alpha1.HostMaster{
+		l2VniRedVtep := l2VniRed.DeepCopy()
```
maiqueb: I think what made this VNI "red" was the VRF name. 

Please choose a different name for it . Yes, I'm aware it'll bloat the changes ... IMHO that's the price to pay for being explicit.

e2etests/tests/resiliency.go
```diff
@@ -0,0 +1,375 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "l2-conflict",
+			Namespace: openperouter.Namespace,
+		},
+		Spec: v1alpha1.L2VNISpec{
+			VNI: 300,
+		},
+	}
+
+	BeforeAll(func() {
```
maiqueb, line 90: don't you need to wait for the rollout of new router pods ? (since you've removed the existing underlay).

e2etests/tests/resiliency.go
```diff
@@ -0,0 +1,375 @@
...
+		}, 2*time.Minute, time.Second).ShouldNot(HaveOccurred())
+	})
+
+	AfterEach(func() {
+		dumpIfFails(cs)
+		Expect(Updater.CleanButUnderlay()).To(Succeed())
+	})
+
+	Context("when L3VNI and L2VNI have the same VNI number", func() {
+		It("should quarantine the conflicting L2VNI and configure the good resources", func() {
```
maiqueb, line 125: I'm a bit confused here ... Shouldn't the L3VNI be quarantined as well ? 

Like, it is relying on the L2 (since it doesn't have hostmaster set). It won't work when we provision it.

e2etests/tests/resiliency.go
maiqueb, line 1: I LOVE these tests. Only have a single question regarding the system behavior.

maiqueb (commented):
Some feedback for when you productify this PR. 

I think this meets the goals we had set. 

Please rebase the PR, so we can get results from CI.

internal/controller/routerconfiguration/status.go
RamLavi: I think you're right, better to exclude AlreadyExists err.
DONE

internal/controller/routerconfiguration/status.go
RamLavi: underlay is special because when it fails, everything else is skipped, so it gets its own dedicated condition reason and message ("VNI processing skipped").

internal/controller/routerconfiguration/reconcile_result.go
RamLavi: yes we want to find underlay Errors easily using the HasFailure() function.

internal/controller/routerconfiguration/status.go
RamLavi: AFAIU the enhancement doc only [defines](https://github.com/openperouter/openperouter/pull/225/changes#diff-1268f42294163078c6251982a20167009fe7ee0bd9760f631c225fac24100b31R245-R248) three condition reasons:  
  - ConfigurationSuccessful                                                                                                                                   
  - ConfigurationFailed                                                     
  - UnderlayFailed

Passthrough is different - if passthrough fails, it gets quarantined as a `FailedResource` just like any bad L3VNI or L2VNI, and the rest of the configuration proceeds normally. It falls into the generic "N resources failed".

internal/controller/routerconfiguration/status.go
RamLavi: OK I think I understand. will do

internal/controller/routerconfiguration/status.go
RamLavi: ACK

internal/controller/routerconfiguration/status.go
RamLavi: DONE

internal/controller/routerconfiguration/status.go
RamLavi: I inlined the setConditions into the buildStatus function ..

internal/controller/routerconfiguration/status_test.go
RamLavi: DONE

internal/controller/routerconfiguration/status_test.go
RamLavi: harmless, but imo it's redundant, package name is already routerconfiguration..

internal/controller/routerconfiguration/status.go
RamLavi: DONE for all types.

internal/controller/routerconfiguration/reconcile.go
RamLavi: yes. I'll return empty slice now, and fill it later in the commit that adds this logic.

internal/controller/routerconfiguration/static_configuration_controller.go
RamLavi: IIRC, the static controller shouldn't report per-resource status, so the result is intentionally discarded.

internal/conversion/validate_vni.go
RamLavi: oh wow missed that ugly thing. fixed.

internal/conversion/validate_vni.go
RamLavi: You're correct that L2VNIs don't have RTs.
However, this commit is a pure refactor extracting the per-item validators from the existing shared validateVNIs function, which already called ValidateRouteTargets for both types. Removing it here would be a behavior change. Happy to do it as a separate commit if you'd like.

internal/conversion/validate_vni_test.go
RamLavi: DONE

internal/controller/routerconfiguration/reconcile.go
RamLavi: maybe it's the comments relocation, but I'm not sure what you mean, can you clarify?

internal/controller/routerconfiguration/reconcile.go
RamLavi: yes.

internal/controller/routerconfiguration/reconcile.go
RamLavi: DONE. tell me what you think

e2etests/tests/evpn_l2.go
RamLavi: true. fixed

e2etests/tests/resiliency.go
RamLavi: This happens only at the start (BeforeAll) I agree a former suite could remove it, so better wait for rollout. added.

internal/conversion/validate_vni.go
maiqueb: all good, this is a poc. Consider it for the implementation PR please though.

e2etests/tests/resiliency.go
RamLavi: well, you're right that the L3VNI ends up with an empty VRF at runtime since its only L2VNI was quarantined. But detecting that would require reverse-dependency tracking ("does this L3VNI still have at least one connected L2VNI?"), which is a different scope than the current per-resource validation. We would need to think about it more carefully, if it's worth it.

internal/controller/routerconfiguration/reconcile.go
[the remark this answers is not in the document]
RamLavi: yeah you're right. I'll add a commit that handles this function

internal/controller/routerconfiguration/reconcile.go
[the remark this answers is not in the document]
RamLavi: @gemini-code-assist is that answer OK for you ?

internal/controller/routerconfiguration/status_test.go
maiqueb: ok then

internal/controller/routerconfiguration/status.go
qinqon: I don't think so, we should not block the goroutine of the Reconcile if possible, let's just return the errorr so it gets re-reconcile for example on conflicts.

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,163 @@
...
+			Message:            reconcileErr.Error(),
+			LastTransitionTime: now,
+		})
+		apimeta.SetStatusCondition(&status.Status.Conditions, metav1.Condition{
+			Type:               ConditionTypeDegraded,
+			Status:             metav1.ConditionTrue,
+			Reason:             ReasonConfigurationFailed,
+			Message:            reconcileErr.Error(),
+			LastTransitionTime: now,
+		})
```
qinqon: The transition time should be set the SetStatusCondition helper, so internally shuld calculate the now and set it, we are not going to hack the time any part in the code for this.

internal/conversion/validate_vni.go
```diff
@@ -40,11 +40,29 @@ func ValidateL3VNIsForNodes(nodes []corev1.Node, underlays []v1alpha1.L3VNI) err
 	return nil
 }
 
-// ValidateL3VNIs runs L3VNI specific validation.
+// ValidateL3VNI validates a single L3VNI's fields (VRF name, route targets).
+func ValidateL3VNI(l3Vni v1alpha1.L3VNI) error {
+	vni := vnisFromL3VNIs([]v1alpha1.L3VNI{l3Vni})[0]
+	if err := isValidInterfaceName(vni.vrfName); err != nil {
+		return fmt.Errorf("invalid vrf name for vni %s: %s - %w", vni.name, vni.vrfName, err)
```
qinqon: Let's not break the convention of using ":" to show the wrapped error.
```suggestion
		return fmt.Errorf("invalid vrf name for vni %q, vrf %q: %w", vni.name, vni.vrfName, err)
```

internal/conversion/validate_vni.go
```diff
@@ -40,11 +40,29 @@ func ValidateL3VNIsForNodes(nodes []corev1.Node, underlays []v1alpha1.L3VNI) err
...
+	}
+	if err := ValidateRouteTargets(vni); err != nil {
+		return fmt.Errorf("invalid route targets for vni %s: %w", vni.name, err)
```
qinqon:
```suggestion
		return fmt.Errorf("invalid route targets for vni %q: %w", vni.name, err)
```

internal/controller/routerconfiguration/status.go
RamLavi: perhaps we should inquire with @fedepaol

Hey Federico - according to the [enhancement](https://github.com/openperouter/openperouter/pull/225/changes#diff-1268f42294163078c6251982a20167009fe7ee0bd9760f631c225fac24100b31R245-R248) - the logic is that when underlay configuration fails, everything else is skipped.
Is that the same logic for passthrough? Right now, I've been treating it as a resource, meaning that if passthrough fails, it gets quarantined as a FailedResource just like any bad L3VNI or L2VNI, and the rest of the configuration proceeds normally. It falls into the generic "N resources failed".

Is this what you intended?

internal/conversion/validate_vni.go
```diff
@@ -63,34 +81,43 @@ func ValidateL2VNIsForNodes(nodes []corev1.Node, underlays []v1alpha1.L2VNI) err
...
-			}
+	}
+	if len(l2Vni.Spec.L2GatewayIPs) > 0 && !hasVRF(l2Vni) {
+		return fmt.Errorf("l2gatewayips cannot be set without spec.vrf for vni %q", l2Vni.Name)
+	}
+	if len(l2Vni.Spec.L2GatewayIPs) > 0 {
+		if _, err := ipfamily.ForCIDRStrings(l2Vni.Spec.L2GatewayIPs...); err != nil {
+			return fmt.Errorf("invalid l2gatewayips for vni %q = %v: %w", l2Vni.Name, l2Vni.Spec.L2GatewayIPs, err)
 		}
 	}
```
qinqon, line 106: Shouldn't we CEL inforce this things ?

internal/controller/routerconfiguration/host_config.go
```diff
@@ -68,56 +69,86 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
...
+func provisionOverlays(ctx context.Context, config interfacesConfiguration, hostConfig conversion.HostConfigData) (ReconcileResult, error) {
+	var result ReconcileResult
+
+	slog.InfoContext(ctx, "configure interface start", "namespace", config.targetNamespace)
+	defer slog.InfoContext(ctx, "configure interface end", "namespace", config.targetNamespace)
+
+	failedVRFs := sets.New[string]()
+	l3NameByVRF := map[string]string{}
+	for _, l3 := range config.L3VNIs {
+		l3NameByVRF[l3.Spec.VRF] = l3.Name
```
qinqon: Maybe we don't need `failedVRFs` we can just remove the entry from `l3NameByVRF` on error and later on so instead of 
```golang
failedVRFs.Has(vni.VRF) 
```
we just do
```golang
l3NameByVRF[vni.VRF]
```

internal/controller/routerconfiguration/host_config.go
```diff
@@ -68,56 +69,86 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
...
+			failedVRFs.Insert(vni.VRF)
+			continue
 		}
+		configuredL3VNIs = append(configuredL3VNIs, vni)
 	}
 
-	for _, vni := range hostConfig.L2VNIs {
+	var configuredL2VNIs []hostnetwork.L2VNIParams
+	for i, vni := range hostConfig.L2VNIs {
+		l2Name := config.L2VNIs[i].Name
```
qinqon: Are we sure the `i` from hostConfig.L2VNIs matches the `i` from config.L2VNIs ? do we need a similar map as L3VNI ?

internal/controller/routerconfiguration/host_config.go
```diff
@@ -68,56 +69,86 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
...
 		if err := hostnetwork.SetupL2VNI(ctx, vni); err != nil {
-			return fmt.Errorf("failed to setup vni: %w", err)
+			result.AddFailure("L2VNI", l2Name, v1alpha1.OverlayAttachmentFailed, err.Error())
+			continue
 		}
 		if err := bridgerefresh.StartForVNI(ctx, vni); err != nil {
-			return fmt.Errorf("failed to start bridge refresher for vni %d: %w", vni.VNI, err)
+			result.AddFailure("L2VNI", l2Name, v1alpha1.OverlayAttachmentFailed, err.Error())
+			continue
 		}
```
qinqon, line 136: Is it worth it a little haper to do addL2VNIFailure so we don't repeat ourselves with the "L2VN" and the OverlayAttachmentFailed ?

internal/controller/routerconfiguration/host_config.go
```diff
@@ -68,56 +69,86 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
...
+		configuredL3VNIs = append(configuredL3VNIs, vni)
 	}
 
-	for _, vni := range hostConfig.L2VNIs {
+	var configuredL2VNIs []hostnetwork.L2VNIParams
+	for i, vni := range hostConfig.L2VNIs {
+		l2Name := config.L2VNIs[i].Name
+		if failedVRFs.Has(vni.VRF) {
+			result.AddFailure("L2VNI", l2Name, v1alpha1.OverlayAttachmentFailed,
+				fmt.Sprintf("VRF %s failed netlink provisioning", vni.VRF))
```
qinqon:
```suggestion
				fmt.Sprintf("VRF %q failed netlink provisioning", vni.VRF))
```

internal/controller/routerconfiguration/host_config.go
```diff
@@ -68,56 +69,86 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
...
 		slog.InfoContext(ctx, "setting up L2VNI", "vni", vni.VNI)
 		if err := hostnetwork.SetupL2VNI(ctx, vni); err != nil {
-			return fmt.Errorf("failed to setup vni: %w", err)
+			result.AddFailure("L2VNI", l2Name, v1alpha1.OverlayAttachmentFailed, err.Error())
+			continue
 		}
 		if err := bridgerefresh.StartForVNI(ctx, vni); err != nil {
-			return fmt.Errorf("failed to start bridge refresher for vni %d: %w", vni.VNI, err)
+			result.AddFailure("L2VNI", l2Name, v1alpha1.OverlayAttachmentFailed, err.Error())
+			continue
```
qinqon, line 135: This is changing changing the behaviour, original we just return on fist error, not sure if calling StartForVNI make sense if SetupL2VNI has fail or if l3vni has failed.

internal/controller/routerconfiguration/host_config.go
```diff
@@ -68,56 +69,86 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
...
 		}
+		configuredL2VNIs = append(configuredL2VNIs, vni)
 	}
 
 	slog.InfoContext(ctx, "setting up passthrough")
 	if hostConfig.L3Passthrough != nil {
 		if err := hostnetwork.SetupPassthrough(ctx, *hostConfig.L3Passthrough); err != nil {
-			return fmt.Errorf("failed to setup passthrough: %w", err)
+			result.AddFailure("L3Passthrough", config.L3Passthrough[0].Name, v1alpha1.OverlayAttachmentFailed, err.Error())
```
qinqon: we are missing return here, is that ok ?

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -6,53 +6,147 @@ import (
...
+				continue
+			}
+		}
+		if err := conversion.ValidateL2VNI(l2); err != nil {
+			result.AddFailure("L2VNI", l2.Name, v1alpha1.ValidationFailed, err.Error())
+			continue
+		}
+		if existing, ok := usedVNIs[l2.Spec.VNI]; ok {
+			result.AddFailure("L2VNI", l2.Name, v1alpha1.ValidationFailed,
+				fmt.Sprintf("duplicate VNI %d, already used by %s", l2.Spec.VNI, existing))
```
qinqon, line 128: This is already check at ValidateL2VNIs, maybe we should use that ?

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -6,53 +6,147 @@ import (
...
+		if existing, ok := usedVNIs[l3.Spec.VNI]; ok {
+			result.AddFailure("L3VNI", l3.Name, v1alpha1.ValidationFailed,
+				fmt.Sprintf("duplicate VNI %d, already used by %s", l3.Spec.VNI, existing))
+			quarantinedVRFs.Insert(l3.Spec.VRF)
+			continue
+		}
+		usedVRFs[l3.Spec.VRF] = l3.Name
+		usedVNIs[l3.Spec.VNI] = "L3VNI/" + l3.Name
+		valid = append(valid, l3)
+	}
```
qinqon, line 107: We Have alrady a ValidateL3VNIs checing dup stuff, maybe we should move this logic we some extra stuff there so we have the dup check just in one place ?

internal/controller/routerconfiguration/status_test.go
qinqon: I think apimachinery has a helper for it or some other k8s golang util

```golang
import (
    "k8s.io/apimachinery/pkg/api/meta"
    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

func checkStatus(status *metav1.Status) {
    // Assuming 'status.Conditions' is a slice of metav1.Condition
    condition := meta.FindStatusCondition(status.Conditions, "Available")
    
    if condition != nil {
        if condition.Status == metav1.ConditionTrue {
            // Do something
        }
    }
}
```

qinqon (changes requested):
First pass

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/f0375a6f959d8514936dacd94cdcecb0951e8167..0962ec24dea99ac82f9f1fb27f6113ecf4432726): Rebase + address @maiqueb review comments

(Sorry, forgot to push the rebase separately)

e2etests/tests/resiliency.go
RamLavi: Actually - I was wrong. 
CI showed this actually times out when no prior underlay exists - CleanAll() has nothing to remove, so no rollout happens and the wait hangs for 2 minutes. 

Reverted to the standard pattern used by sessions.go / evpn_routes.go: 
- CleanAll() first, 
- capture routers after, 
- no rollout wait in BeforeAll. 
- The AfterAll rollout wait still covers the cleanup path.

internal/controller/routerconfiguration/status.go
RamLavi: well currently updateStatus caller only logs in case of error (since status update is best effort) - so if status errors on conflict, it will be just logged and silently swallowed. 
I think retryOnConflict is a good mitigation for this. if there is a conflict then most probably next retry will work. 
The upside of using retryOnConflict is that status reflects the current state immediately rather than waiting for the next reconcile trigger. I would vote keeping it.

internal/controller/routerconfiguration/reconcile.go
RamLavi: DONE

internal/controller/routerconfiguration/status.go
RamLavi: I mainly did it because I understood from @maiqueb that he wants me to unit test the timestamp. 
But it was a stretch so maybe I misinterpreted. @maiqueb can leave the TS to the SetStatusCondition helper?

internal/controller/routerconfiguration/status_test.go
RamLavi: DONE

internal/conversion/validate_vni.go
RamLavi: DONE

internal/conversion/validate_vni.go
RamLavi: DONE

internal/conversion/validate_vni.go
RamLavi: This is just a refactor of what we are already doing. I do this so that I could reuse it on next commit.

Moreover, the existing ValidateL3VNIs/ValidateL2VNIs only checks for duplicates within the same type. The reconcile's usedVNIs map on the next commit checks cross-type..

internal/controller/routerconfiguration/host_config.go
RamLavi: That's fair. DONE

RamLavi: oops. won't work. 
Disconnected L2VNIs (no VRF set) **also** pass through validation and reach this code. 
Their VRF name wouldn't be in l3NameByVRF, so deleting entries from the map would incorrectly skip them. 
failedVRFs only tracks explicitly failed VRFs, which is the right semantic here

internal/controller/routerconfiguration/host_config.go
RamLavi: APItoHostConfig iterates apiConfig.L2VNIs in order and appends each to res.L2VNIs in [host_conversion.go](https://github.com/openperouter/openperouter/blob/main/internal/conversion/host_conversion.go#L111-L136) - no filtering or reordering. The indices are guaranteed to match.

But it CAN change in the future - so maybe worth mapping?

internal/controller/routerconfiguration/host_config.go
RamLavi: DONE

internal/controller/routerconfiguration/host_config.go
RamLavi: I am not sure as well honestly. @maiqueb what are your thoughts?

internal/controller/routerconfiguration/host_config.go
RamLavi: it's used 3 times so I don't think it's worth it tbh..

internal/controller/routerconfiguration/host_config.go
RamLavi: Yes, intentionally no return. IIUC, a passthrough failure shouldn't block VNI cleanup - those are independent. (I could be wrong here, this is just my understanding of the enhancment)

internal/controller/routerconfiguration/reconcile.go
RamLavi: This is partly what I did with the ValidateL3VNI/ValidateL2VNI refactor - I extracted the most I could into the singular validators so they could be reused by both the plural functions (ValidateL3VNIs/ValidateL2VNIs) and the reconciler's quarantine functions.                                                                                

The thing the reconciler does that the plural functions don't is **cross-type** duplicate checking - i.e. an L3VNI and L2VNI can't share the same VNI number. 

That requires the shared usedVNIs map passed to both validateL3VNIsWithQuarantine and validateL2VNIsWithQuarantine.

internal/controller/routerconfiguration/reconcile.go
RamLavi: See [comment](https://github.com/openperouter/openperouter/pull/345/changes/e67bbc79908ccefed7662d334670b7e2524cb3c5#r3218497727)

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/0962ec24dea99ac82f9f1fb27f6113ecf4432726..971b2194f5fc5b3c7ff92fcdc8ef639fe3917c6e): Fix BeforeAll issue

internal/controller/routerconfiguration/host_config.go
maiqueb: TBH, given my lack of overall clarity in the passthrough use-cases, I'd be more comfortable if @fedepaol chimed in.

internal/controller/routerconfiguration/status.go
maiqueb: Tks. Let's resolve.

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,161 @@
...
+			Type:               ConditionTypeDegraded,
+			Status:             metav1.ConditionTrue,
+			Reason:             ReasonConfigurationFailed,
+			Message:            result.ReconcileError.Error(),
+			LastTransitionTime: now,
+		})
+		return s
+	}
+
+	if !result.HasFailures() && !result.HasFailure(KindUnderlay) {
```
maiqueb: nit: isn't this redundant ? Like, if it doesn't have failures, why do we have to check that we don't have underlay failures ?

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,161 @@
...
+		})
+		return s
+	}
+
+	names := make([]string, len(result.FailedResources))
+	for i, f := range result.FailedResources {
+		names[i] = f.Kind + "/" + f.Name
+	}
+	slices.Sort(names)
+	msg := fmt.Sprintf("%d resources failed (%s)", len(result.FailedResources), strings.Join(names, ", "))
```
maiqueb: nit: maybe turn this into an helper ?

internal/controller/routerconfiguration/status_test.go
```diff
@@ -0,0 +1,125 @@
...
+			expectedReadyReason:  ReasonConfigurationSuccessful,
+			expectedReadyMessage: "All configuration applied successfully",
+		},
+		{
+			name:                 "underlay failed",
+			result:               underlayResult,
+			expectedReady:        metav1.ConditionFalse,
+			expectedDegraded:     metav1.ConditionTrue,
+			expectedReadyReason:  ReasonUnderlayFailed,
+			expectedReadyMessage: "Underlay failed validation, existing FRR configuration left as-is",
```
maiqueb, line 59: @fedepaol in your opinion what should we do here ? 

I kind of remember us discussing that maybe this would make our lives harder, and that we should probably invalidate the FRR config as well (like, the underlay is broken, the FRR conf does not matter at this point - traffic won't flow).

internal/conversion/validate_vni_test.go
```diff
@@ -13,6 +13,108 @@ import (
...
+}
+
+func TestValidateL2VNI(t *testing.T) {
+	tests := []struct {
+		name           string
+		vni            v1alpha1.L2VNI
+		wantErrContain string
+	}{
+		{
+			name: "valid L2VNI",
```
maiqueb: I think we should have two valid L2 VNIs:
- one with VRF + gw IPs (maybe the GW IPs is going too far).
- a disconnected overlay (no VRF, no GW IPs)

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -67,3 +69,69 @@ func underlayName(apiConfig conversion.APIConfigData) string {
...
+				fmt.Sprintf("duplicate VRF %s, already used by L3VNI %s", l3.Spec.VRF, existing))
+			continue
+		}
+		if existing, ok := usedVNIs[l3.Spec.VNI]; ok {
+			result.AddFailure(KindL3VNI, l3.Name, v1alpha1.ValidationFailed,
+				fmt.Sprintf("duplicate VNI %d, already used by %s", l3.Spec.VNI, existing))
+			continue
+		}
+		usedVRFs[l3.Spec.VRF] = l3.Name
+		usedVNIs[l3.Spec.VNI] = KindL3VNI + "/" + l3.Name
```
maiqueb, line 105: nit: is the `KindL3VNI` really necessary ?

internal/controller/routerconfiguration/reconcile.go
```diff
@@ -19,30 +19,32 @@ func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, underlay
...
-		return result, fmt.Errorf("failed to validate l3vnis: %w", err)
-	}
+	usedVNIs := map[int32]string{}
 
-	if err := conversion.ValidateL2VNIs(apiConfig.L2VNIs); err != nil {
-		return result, fmt.Errorf("failed to validate l2vnis: %w", err)
-	}
+	l3Result := validateL3VNIsWithQuarantine(apiConfig.L3VNIs, usedVNIs, &result)
+	validL2VNIs := validateL2VNIsWithQuarantine(apiConfig.L2VNIs, usedVNIs, l3Result.ValidVRFs, &result)
+	validPassthrough := validatePassthroughsWithQuarantine(apiConfig.L3Passthrough, &result)
```
maiqueb: NIT: I think these functions all have the same "problem" ... They all perform validation of the resources, *plus* they mutate the result as a side-effect. 

I know this is a PoC, and we're after the functionality. But maybe we can explore something more explicit - either also return the failed entities, and merge them in the caller, or IDK, encapsulate these functions in a struct which embeds the result, and mutate it internally ?

internal/controller/routerconfiguration/reconcile.go
maiqueb, line 1: don't we need associated unit tests to assert these behaviors ? For instance, I'd like to see a cross resource validation at play somewhere (e.g. the same VNI used in an L3 and L2 VNI, etc).

internal/conversion/validate_vni.go
```diff
@@ -155,7 +154,15 @@ func ValidateVRFs(l2Vnis []v1alpha1.L2VNI, l3Vnis []v1alpha1.L3VNI) error {
...
-	// Make sure that there are no subnet overlaps in the VRFs.
+	for _, err := range ValidateVRFSubnets(l2Vnis, l3Vnis) {
+		return err
+	}
+	return nil
+}
+
+// ValidateVRFSubnets checks for subnet overlaps per VRF and returns a map of
+// VRF name to error for each VRF that has overlapping subnets.
+func ValidateVRFSubnets(l2Vnis []v1alpha1.L2VNI, l3Vnis []v1alpha1.L3VNI) map[string]error {
```
maiqueb, line 165: nit: do we need to export this function ? Will it be used outside the pkg ? 

While I get the scope of this very simple commit, maybe if you include also all the "users" of it, we would know without having to ask.

internal/controller/routerconfiguration/host_config.go
maiqueb: I think this is the goal of the enhancement: do not let an error in a particular VRF (or VNI) affect the others. 

This is desired IMHO.

internal/controller/routerconfiguration/status.go
qinqon: > well currently updateStatus caller only logs in case of error (since status update is best effort) - so if status errors on conflict, it will be just logged and silently swallowed. I think retryOnConflict is a good mitigation for this. if there is a conflict then most probably next retry will work. 

Ok, let's do it.

> The upside of using retryOnConflict is that status reflects the current state immediately rather than waiting for the next reconcile trigger. I would vote keeping it.

Well that depend on how you implementing, if there is a conflict error status should not be configured at all.

internal/controller/routerconfiguration/status.go
qinqon: You can do some trick to test timestamp without passing to the function, I think is better to to calculate inside of it.

internal/controller/routerconfiguration/host_config.go
qinqon: That's does not look robust, implementation chan change so order can change, let's map things up.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/971b2194f5fc5b3c7ff92fcdc8ef639fe3917c6e..33ab537e1a6d440f92e5790ff7edc25086a6c5b3): Rebase

internal/controller/routerconfiguration/status.go
RamLavi: yes this is silly. 
DONE

internal/controller/routerconfiguration/status.go
RamLavi: DONE

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,161 @@
...
+			LastTransitionTime: now,
+		})
+		apimeta.SetStatusCondition(&s.Conditions, metav1.Condition{
+			Type:               ConditionTypeDegraded,
+			Status:             metav1.ConditionTrue,
+			Reason:             ReasonConfigurationFailed,
+			Message:            result.ReconcileError.Error(),
+			LastTransitionTime: now,
+		})
+		return s
```
qinqon: I think is worth it to have have a pair of cunction to set segraded and ready since it do the swap between them like 
```golang
func setStatusConditionDegraded(reason, message string, lasTransitionTime metav1.Time)       
        apimeta.SetStatusCondition(&s.Conditions, metav1.Condition{
			Type:               ConditionTypeReady,
			Status:             metav1.ConditionFalse,
			Reason:             reason.
			Message:            message,
			LastTransitionTime: now,
		})
		apimeta.SetStatusCondition(&s.Conditions, metav1.Condition{
			Type:               ConditionTypeDegraded,
			Status:             metav1.ConditionTrue,
			Reason:             reason,
			Message:           message,
			LastTransitionTime: now,
		})
```

so you can do later on 
```golang
   setStatusConditionDegraded(ReasonConfigurationFailed, result.ReconcileError.Error(), now)
```
and
```golang
   setStatusConditionDegraded(ReasonUnderlayFailed, "Underlay failed validation, existing FRR configuration left", now)
```

same with Available this way you don't duplicate the true/false of degraded/avaiable

internal/controller/routerconfiguration/status.go
```diff
@@ -0,0 +1,161 @@
...
+		})
+		apimeta.SetStatusCondition(&s.Conditions, metav1.Condition{
+			Type:               ConditionTypeDegraded,
+			Status:             metav1.ConditionTrue,
+			Reason:             ReasonUnderlayFailed,
+			Message:            "Underlay failed validation, VNI processing skipped",
+			LastTransitionTime: now,
+		})
+		return s
+	}
```
qinqon, line 99: Does it make sense to have different messages here ?

internal/controller/routerconfiguration/status.go
qinqon: Not worth it, disregard

qinqon (changes requested):

internal/controller/routerconfiguration/status.go
RamLavi: @maiqueb thinking about it some more, I wonder why need to be testing the TS? AFAIU if left not set, apimeta.SetStatusCondition already sets the TS [internally](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go#L37-L39). 

Can we keep `LastTransitionTime` empty and keep the default behavior?

internal/controller/routerconfiguration/reconcile.go
RamLavi: right now error messages look like `duplicate VNI 300, already used by L3VNI/foo`
If we don't use the kind, then the error message will be ambiguous regarding the type. I think it's worth keeping.

internal/controller/routerconfiguration/status.go
maiqueb: nice suggestion IMHO

internal/controller/routerconfiguration/reconcile.go
RamLavi: I agree. For that, I need to add a "updater" function (similar to `updater frr.ConfigUpdater`) so that I could test the Reconcile function which is the only public function to unit test.
Will add it as a pre-commit so things stay sane to review

internal/controller/routerconfiguration/status.go
maiqueb: I was only trying to avoid having to sanitize the conditions in the tests because of the internally calculated time stamp. 

If you think that's too much work / the wrong direction, feel free to propose something different.

internal/controller/routerconfiguration/reconcile.go
RamLavi: DONE

internal/controller/routerconfiguration/status.go
RamLavi: OK since you don't feel very strong against it, I'll drop the TS checks. if we see we need it later - we'll add

internal/controller/routerconfiguration/status.go
RamLavi: it's not really important I think. for the sake of using the helper - I'll merge to one msg..

internal/controller/routerconfiguration/status.go
RamLavi: DONE

internal/conversion/validate_vni_test.go
RamLavi: DONE

internal/conversion/validate_vni.go
RamLavi: We need it exported because it is used by ValidateVRFs (called by the webhook), and for our reconcile. I'll make it more clear on the commit message

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/33ab537e1a6d440f92e5790ff7edc25086a6c5b3..ab8a2f5fb7615b23a62dc00c5dce85e73cbda55b): Address @maiqueb and @qinqon 's latest review cycles:

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/ab8a2f5fb7615b23a62dc00c5dce85e73cbda55b..a58460eedc1173e66b8c5a546ae9a21c04b8fe18): Fix lint issue

PR #350, opened 2026-04-30
fix: add link-local address to L2VNI bridge for IPv6 NDP

### LINKED ISSUE #342: flaky E2E test "Routes between bgp and the fabric should create two pods connected to the l2 overlay for dual stack" - both for linux and ovs bridges
## Description
                                                                                                                                                  
The E2E test `Routes between bgp and the fabric -- should create two pods connected to the l2 overlay -- OVS bridge autocreate for dual stack` is flaky. When it fails, IPv6 pod-to-host connectivity times out while IPv4 works fine through the same OVS bridge.
                                                                                                                                                
The failure manifests as curl timing out trying to reach a host on the red VRF over IPv6:                                                       
```
[FAILED] Timed out after 139.720s.                                                                                                              
The function passed to Eventually failed at e2etests/tests/validate_pods.go:24 with:                                                          
curl 192.171.24.2:8090 failed: curl: (28) Failed to connect to 192.171.24.2 port 8090 after 135539 ms: Operation timed out
```

### Failed runs                                           

- Nightly on main: https://github.com/openperouter/openperouter/actions/runs/24972862373                                                        
- https://github.com/openperouter/openperouter/actions/runs/24896636853
- https://github.com/openperouter/openperouter/actions/runs/24895248428

### Not only OVS
Please note that the *[nightly on main](**url**)* lane failed on linux bridge, not OVS bridge. Both bridge types seem to be affected - probably OVS is just more common. 

### Key findings
The problem is NOT in the OVS bridge. I proved this by pinging from the PE itself (inside perouter netns, bypassing OVS entirely): 
                                                                                                                                                  
  - ping -I red 192.168.20.2 → WORKS (0% loss)                                                                                                    
  - ping -I red 2001:db8:20::2 → FAILS (100% loss)                                                                                                
  - ping 2001:db8:20::1 (leafA gateway) → FAILS (100% loss)

maiqueb:

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design

/kind flake

> /kind failing
> /kind documentation
> /kind regression
> /kind example

The kernel's `ndisc_send_ns()` silently drops unicast Neighbor Solicitation probes when the outgoing interface has no link-local address. 

Since we set `addr_gen_mode=1` on PE bridges (suppressing automatic link-local generation), the L2VNI bridge (br-pe-110) has no link-local when IPv6 gateway IPs are configured. This causes NDP neighbor entries to cycle DELAY → PROBE → FAILED because the kernel never sends the unicast NS probe.

Without stable NDP entries, FRR cannot advertise Type-2 MAC+IP EVPN routes for pods' IPv6 addresses. 

Return traffic via ECMP uses the Type-5 prefix route, lands on the wrong PE, and is dropped - causing the dual-stack E2E test flake on first run.

Add an EUI-64 link-local address (derived from the bridge's fixed MAC) to L2VNI bridges that have IPv6 gateway IPs. This enables the kernel's native NDP to work correctly while keeping `addr_gen_mode=1` for EVPN compatibility.

Fixes: #342

**Release note**:

```release-note

```

### DIFF
--- a/internal/hostnetwork/vni.go
+++ b/internal/hostnetwork/vni.go
@@ -7,8 +7,10 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net"
 	"strings"
 
+	"github.com/openperouter/openperouter/internal/ipfamily"
 	"github.com/openperouter/openperouter/internal/netnamespace"
 	"github.com/openperouter/openperouter/internal/ovsmodel"
 	libovsclient "github.com/ovn-kubernetes/libovsdb/client"
@@ -203,6 +205,12 @@ func SetupL2VNI(ctx context.Context, params L2VNIParams) error {
 			// setting up the same mac address for all the nodes for distributed gateway
 			if err := ensureBridgeFixedMacAddress(bridge, params.VNI); err != nil {
 				return fmt.Errorf("failed to set bridge mac address %s: %v", name, err)
+			}
+
+			if hasIPv6Gateway(params.L2GatewayIPs) {
+				if err := ensureBridgeLinkLocal(bridge); err != nil {
+					return fmt.Errorf("failed to ensure link-local on bridge %s: %w", name, err)
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


maiqueb:
Closing it in lieu of #351 , which is more aligned to the FRR documentation.

PR #351, opened 2026-04-30
refactor: use functional options for bridge setup

### LINKED ISSUE #342: flaky E2E test "Routes between bgp and the fabric should create two pods connected to the l2 overlay for dual stack" - both for linux and ovs bridges

maiqueb:

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design

/kind flake

> /kind failing
> /kind documentation
> /kind regression
> /kind example

Make `setupBridge` accept variadic `BridgeOption` functions instead of hardcoding `setAddrGenModeNone`. `L3VNI` passes `WithAddrGenModeNone()` to suppress IPv6 link-local generation; L2VNI passes no options, allowing the kernel to generate a link-local address naturally. 

This fixes the dual-stack NDP flake (#342) where the kernel's `ndisc_send_ns()` silently dropped unicast NS probes on L2VNI bridges that had no link-local.

Fixes: #342 

**Release note**:

```release-note
Do not set AddrGenModeNone on L2 VNI bridges.
```

### DIFF
--- a/internal/hostnetwork/bridge.go
+++ b/internal/hostnetwork/bridge.go
@@ -13,18 +13,32 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
-// setup bridge creates the bridge if not exists, and it enslaves it to the provided
-// vrf.
-func setupBridge(params VNIParams, vrf *netlink.Vrf) (*netlink.Bridge, error) {
+// BridgeOption is a function applied to a bridge after creation but before link-up.
+type BridgeOption func(*netlink.Bridge) error
+
+// WithAddrGenModeNone returns a BridgeOption that sets addr_gen_mode=1 (none)
+// on the bridge, suppressing automatic IPv6 link-local address generation.
+// L3VNI bridges need this; L2VNI bridges must NOT use it because the kernel's
+// NDP state machine requires a link-local address for unicast NS probes.
+func WithAddrGenModeNone() BridgeOption {
+	return func(bridge *netlink.Bridge) error {
+		return setAddrGenModeNone(bridge)
+	}
+}
+
+// setupBridge creates the bridge if not exists, enslaves it to the provided
+// vrf, and applies any bridge options before bringing the link up.
+func setupBridge(params VNIParams, vrf *netlink.Vrf, opts ...BridgeOption) (*netlink.Bridge, error) {
 	name := BridgeName(params.VNI)
 	bridge, err := createBridge(name, vrf.Index)
 	if err != nil {
 		return nil, err
 	}
 
-	err = setAddrGenModeNone(bridge)
-	if err != nil {
-		return nil, fmt.Errorf("failed to set addr_gen_mode to 1 for %s: %w", bridge.Name, err)
+	for _, opt := range opts {
+		if err := opt(bridge); err != nil {
+			return nil, fmt.Errorf("failed to apply bridge option for %s: %w", bridge.Name, err)
+		}
 	}
 
 	err = linkSetUp(bridge)
--- a/internal/hostnetwork/bridge.go
+++ b/internal/hostnetwork/bridge.go
@@ -13,22 +13,9 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
-// BridgeOption is a function applied to a bridge after creation but before link-up.
-type BridgeOption func(*netlink.Bridge) error
-
-// WithAddrGenModeNone returns a BridgeOption that sets addr_gen_mode=1 (none)
-// on the bridge, suppressing automatic IPv6 link-local address generation.
-// L3VNI bridges need this; L2VNI bridges must NOT use it because the kernel's
-// NDP state machine requires a link-local address for unicast NS probes.
-func WithAddrGenModeNone() BridgeOption {
-	return func(bridge *netlink.Bridge) error {
-		return setAddrGenModeNone(bridge)
-	}
-}
-
 // setupBridge creates the bridge if not exists, enslaves it to the provided
 // vrf, and applies any bridge options before bringing the link up.
-func setupBridge(params VNIParams, vrf *netlink.Vrf, opts ...BridgeOption) (*netlink.Bridge, error) {
+func setupBridge(params VNIParams, vrf *netlink.Vrf, opts ...NetlinkOption) (*netlink.Bridge, error) {
 	name := BridgeName(params.VNI)
 	bridge, err := createBridge(name, vrf.Index)
 	if err != nil {
--- a/internal/hostnetwork/netlink.go
+++ b/internal/hostnetwork/netlink.go
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package hostnetwork
+
+import "github.com/vishvananda/netlink"
+
+// NetlinkOption is a function applied to a Link after creation but before link-up.
+type NetlinkOption func(link netlink.Link) error
--- a/internal/hostnetwork/vni.go
+++ b/internal/hostnetwork/vni.go
@@ -76,7 +76,7 @@ func (e NotRouterInterfaceError) Error() string {
 // VXLan interface, and moves the veth to the VRF corresponding
 // to the L3 routing domain, exposing it to the default host namespace.
 func SetupL3VNI(ctx context.Context, params L3VNIParams) error {
-	if err := setupVNI(ctx, params.VNIParams); err != nil {
+	if err := setupVNI(ctx, params.VNIParams, WithAddrGenModeNone()); err != nil {
 		return fmt.Errorf("SetupL3VNI: failed to setup VNI: %w", err)
 	}
 	slog.DebugContext(ctx, "setting up l3 VNI", "params", params)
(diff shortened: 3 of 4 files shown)


maiqueb:
Both this PR and #350 fix the same issue. Let's choose one of them to focus on. 

My vote's on this one.

internal/hostnetwork/bridge.go
```diff
@@ -13,18 +13,32 @@ import (
...
-// vrf.
-func setupBridge(params VNIParams, vrf *netlink.Vrf) (*netlink.Bridge, error) {
+// BridgeOption is a function applied to a bridge after creation but before link-up.
+type BridgeOption func(*netlink.Bridge) error
+
+// WithAddrGenModeNone returns a BridgeOption that sets addr_gen_mode=1 (none)
+// on the bridge, suppressing automatic IPv6 link-local address generation.
+// L3VNI bridges need this; L2VNI bridges must NOT use it because the kernel's
+// NDP state machine requires a link-local address for unicast NS probes.
+func WithAddrGenModeNone() BridgeOption {
```
fedepaol: if not used outside, let's not export it.

internal/hostnetwork/vni.go
```diff
@@ -76,7 +76,7 @@ func (e NotRouterInterfaceError) Error() string {
 // VXLan interface, and moves the veth to the VRF corresponding
 // to the L3 routing domain, exposing it to the default host namespace.
 func SetupL3VNI(ctx context.Context, params L3VNIParams) error {
-	if err := setupVNI(ctx, params.VNIParams); err != nil {
+	if err := setupVNI(ctx, params.VNIParams, WithAddrGenModeNone()); err != nil {
```
fedepaol: I find odd to pass the BridgeOption through so many layers, but all the other options are not great. 

- VNIParams is vni type agnostic so setting the flag there doesn't make much sense imo and we are delegating the decision to a higher level (the conversion function).

- The decision to set the property belongs to setupl3vni, but we can't either set it after calling setupVNI because at that point it's too late (I think?), so it must happen in setupVNI.

- splitting setupVNI (as I did in the past) would end up in removing that completely as setupBridge is right the middle instruction

One only comment is that withaddrgenmodenone is a flag in disguise, and until we don't need to tweak more bridge options, is maybe an overkill provided we give a good value / alias to the flag.

qinqon: This doesn't read well, the builder function "WithAddrGenModeNone" make sense if we have it under a package so we do `bridge.WithAddrGenModeNone` or we add Bridge to the name `WithBridgeAddrGenModeNone`

internal/hostnetwork/bridge.go
```diff
@@ -13,18 +13,32 @@ import (
...
+	return func(bridge *netlink.Bridge) error {
+		return setAddrGenModeNone(bridge)
+	}
```
qinqon: You don't need a closure here
```suggestion
	   return setAddrGenModeNone
	}
```

internal/hostnetwork/vni.go
maiqueb: While you're right, I still find this more readable. 

I'm personally good with the overkill, and like how it reads. 

And to be fair, odds are we'll need to tweak something else around these grounds in the future.

internal/hostnetwork/bridge.go
```diff
@@ -13,18 +13,32 @@ import (
...
+}
+
+// setupBridge creates the bridge if not exists, enslaves it to the provided
+// vrf, and applies any bridge options before bringing the link up.
+func setupBridge(params VNIParams, vrf *netlink.Vrf, opts ...BridgeOption) (*netlink.Bridge, error) {
```
qinqon: Not sure we need the Builder opt pattern here maybe just a list of functs is enough 
```suggestion
func setupBridge(params VNIParams, vrf *netlink.Vrf, opts ...func(netlink.Bridge) error) (*netlink.Bridge, error) {
```

this way we can just pass the function we have

```golang
setupVNI(ctx, params.VNIParams, setBridgeAddrGenModeNone)
```

qinqon (changes requested):
Using builder opt pattern for this is overkill, let's just pass a list of `func (netlink.Bridge) error`

internal/hostnetwork/bridge.go
maiqueb: it's much more readable if you ask me.

maiqueb:
> builder opt pattern for this is overkill, let's just pass a list of `func (netlink.Bridge) error`

I don't agree. those are bridge options. i.e. options that can ONLY be applied to a bridge. let's be explicit. if those functions operated on a netlink interface (meaning, not only a bridge, but a veth, a macvlan, a dummy, etc) I would think otherwise. Here, I think we should make as explicit as we can.

internal/hostnetwork/vni.go
fedepaol: With the premise that the current version is fine, 

> While you're right, I still find this more readable.
> 

I think it's equally readable provided we are good with naming. Also, when we go in functional / interface land we loose a bit control over behavior (meaning we need to jump from one place to the other) so it's actually a bit more complex to understand (talking in general, here I agree it's negligible).

> I'm personally good with the overkill, and like how it reads.
> 

Fair enough to me

> And to be fair, odds are we'll need to tweak something else around these grounds in the future.

For that we can still change if needed

maiqueb: I can put the bridge name there, sure. 

tbh, the way I envision this going forward is, we'll need to tweak the bridge, the vrf, and the vxlan we are creating. 

Thus, I think we'll have a common option struct, where we'll put the bridge options, the vrf options, and the vxlan options. Then, each respective setupBridge / setupVRF / setupVXLAN will receive their set of options. 

Of course all this is not needed right now, which is why I didn't add it.

maiqueb:
> > builder opt pattern for this is overkill, let's just pass a list of `func (netlink.Bridge) error`
> 
> I don't agree. those are bridge options. i.e. options that can ONLY be applied to a bridge. let's be explicit. if those functions operated on a netlink interface (meaning, not only a bridge, but a veth, a macvlan, a dummy, etc) I would think otherwise. Here, I think we should make as explicit as we can.

... which, I just realized is exactly the signature [setAddrGenModeNone](https://github.com/openperouter/openperouter/blob/44c9f9cef997bb5ac4395e81f4ff458dba10fff0/internal/hostnetwork/link_properties.go#L78) has.

internal/hostnetwork/vni.go
maiqueb: For instance, an L2VNI does NOT need to have a VRF. That's another obvious candidate for a builder option.

internal/hostnetwork/bridge.go
maiqueb: I've removed this in the latest push, since I'm passing the function `setAddrGenModeNone` directly as a netlink option.

internal/hostnetwork/bridge.go
maiqueb: Done. I've put this in a separate commit in case we don't want to go this direction. 

I'm happy to squash it though.

internal/hostnetwork/bridge.go
maiqueb: I'm doing that, but still defining a netlink option type, which, imho, makes it more readable.

fedepaol:
LGTM, feel free to merge when CI is happy

internal/hostnetwork/netlink.go
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package hostnetwork
+
+import "github.com/vishvananda/netlink"
+
+// NetlinkOption is a function applied to a Link after creation but before link-up.
+type NetlinkOption func(link netlink.Link) error
```
qinqon, line 8: nit: Maybe this should be just `LinkOption` so we can have similar in the future for bridges (netlink.Bridge)

qinqon (changes requested):
Just a nit not a must.

internal/hostnetwork/netlink.go
maiqueb: a bridge is a link.

qinqon (approved):
LGTM

PR #355, opened 2026-05-03
api: Node Router Status

ormergi:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
/kind documentation
> /kind regression
> /kind example

Implement [Status Reporting and Configuration Resilience](https://github.com/openperouter/openperouter/blob/main/enhancements/status-crd.md#status-reporting-and-configuration-resilience) enhancement , [phase 1](https://github.com/openperouter/openperouter/blob/main/enhancements/status-crd.md#phase-1-routernodeconfigurationstatus-crd-creation).

Introduce the node router status API, CRD and controller, enable inspecting openperouter state on a spesific node.

- This change extend the router config controllers (for both k8s and static modes) to manage the node-status resource. 
- The controllers create CRs with static status. Additional business logic will be added in follow-up commits, on this PR or follow-up one, depending on reviewers feedback.
- CRs are created in openperouter namespace, one CR per controller pods's node, CR name is the hosting node name.
- CRs have owner-reference for the controller hosting **node**.  

**Release note**:

```release-note
api: Introduce node router status API. 
Enable inspecting router configuration status on a spesific node via NodeRouterConfigurationStatus CRD.
```


api/v1alpha1/routernodeconfigurationstatus_types.go
```diff
@@ -0,0 +1,111 @@
...
+	ReasonConfigFailed     = "ConfigurationFailed"
+	ReasonUnderlayFailed   = "UnderlayFailed"
+)
+
+// FailedResource describes a single resource that could not be configured on the node.
+type FailedResource struct {
+	// Kind resource type name (e.g.: L3VNI, L2VNI).
+	// +kubebuilder:validation:Required
+	// + kubebuilder:validation:MinLength=1
+	// + kubebuilder:validation:MaxLength=253
```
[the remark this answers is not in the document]
ormergi: This is wrong, space is allowed.

e2etests/tests/node_router_status.go
```diff
@@ -0,0 +1,50 @@
...
+func normConditions(conditions []metav1.Condition) metav1.Condition {
+	var norm []metav1.Condition //nolint:prealloc
+	for i := range conditions {
+		c := conditions[i].DeepCopy()
+		c.LastTransitionTime = metav1.Time{}
+		c.ObservedGeneration = 0
+		norm = append(norm, *c)
+	}
+	return metav1.Condition{}
+}
```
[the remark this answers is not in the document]
ormergi: Fixed

internal/controller/routerstatus/reconciler.go
```diff
@@ -0,0 +1,78 @@
...
+				},
+			},
+		},
+	}
+
+	if err := r.Client.Create(ctx, rns); err != nil {
+		if !k8serrors.IsAlreadyExists(err) {
+			return ctrl.Result{}, err
+		}
+	}
```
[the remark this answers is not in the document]
ormergi: Fixed

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,78 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+
+	"github.com/openperouter/openperouter/api/v1alpha1"
+)
+
+// +kubebuilder:rbac:groups=openpe.openperouter.github.io,resources=routernodeconfigurationstatuses,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=openpe.openperouter.github.io,resources=routernodeconfigurationstatuses/status,verbs=get;update;patch
```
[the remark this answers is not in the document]
ormergi: Spaces are allowed

api/v1alpha1/routernodeconfigurationstatus_types.go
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist what do you think ? @ormergi 's saying it is allowed. 

I'll trust it when I see that e2e tests can run.

api/v1alpha1/routernodeconfigurationstatus_types.go
```diff
@@ -0,0 +1,115 @@
...
+type RouterNodeConfigurationStatusList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []RouterNodeConfigurationStatus `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&RouterNodeConfigurationStatus{}, &RouterNodeConfigurationStatusList{})
+}
```
maiqueb: nit: let's put this in the beginning of the file, after the imports.

api/v1alpha1/routernodeconfigurationstatus_types.go
```diff
@@ -0,0 +1,115 @@
...
+// FailedResourceKind
+// +kubebuilder:validation:Enum=Underlay;L2VNI;L3VNI;FrrConfiguration
+type FailedResourceKind string
+
+const (
+	FailedResourceKindUnderlay         FailedResourceKind = "Underlay"
+	FailedResourceKindL2VNI            FailedResourceKind = "L2VNI"
+	FailedResourceKindL3VNI            FailedResourceKind = "L3VNI"
+	FailedResourceKindFrrConfiguration FailedResourceKind = "FrrConfiguration"
+)
```
maiqueb: don't we have resource kind defined elsewhere ? It seems redundant - all in all, all kinds can fail, right ?

api/v1alpha1/routernodeconfigurationstatus_types.go
```diff
@@ -0,0 +1,115 @@
...
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name"`
+
+	// Reason describe the resource failure.
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=50
+	Reason string `json:"reason"`
```
maiqueb: isn't len 50 too short ? 

Really asking, not sure.

internal/controller/routerstatus/reconciler.go
```diff
@@ -0,0 +1,142 @@
...
+		return ctrl.Result{}, nil
+	}
+
+	nodeStatus := &v1alpha1.RouterNodeConfigurationStatus{
+		ObjectMeta: metav1.ObjectMeta{Name: localNode.Name},
+	}
+	_, err := ctrlutil.CreateOrPatch(ctx, r.client, nodeStatus, func() error {
+		nodeStatus.Status = newStatus()
+		return ctrlutil.SetOwnerReference(localNode, nodeStatus, r.client.Scheme())
+	})
```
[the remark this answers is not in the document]
ormergi: Done

internal/controller/routerstatus/controller.go
```diff
@@ -88,12 +96,17 @@ func (n *NodeStatusController) Reconcile(ctx context.Context, req ctrl.Request)
...
 
+	// set owner reference pointing to the hosting pod to have the CR garbage-collected
+	// when the hosting pod is deleted (due to controller placement change, node drain/taint/removal).
+	podRef := newPodOwnerReference(n.podName, n.podUID)
 	nodeStatus := &v1alpha1.RouterNodeConfigurationStatus{
-		ObjectMeta: metav1.ObjectMeta{Name: n.nodeName, Namespace: n.namespace},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            n.nodeName,
+			Namespace:       n.namespace,
+			OwnerReferences: []metav1.OwnerReference{podRef},
```
RamLavi: enhancement [says](https://github.com/openperouter/openperouter/blob/9bee02cde9e85b76b20d02871b68a3121dac3f23/enhancements/status-crd.md?plain=1#L232) owner reference should be the node, not the pod. 
We can arguably say this is better lifecycle choice (CR goes away when the controller pod goes) but I'm not sure why did we differ from the enhancement. Can you explain?

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// SetupWithManager registers NodeStatusController to controller-runtime manager.
+// Watches RouterNodeConfigurationStatus events of the hosting node.
+func (n *NodeStatusController) SetupWithManager(mgr ctrl.Manager) error {
+	hostingNodeStatusEvents := predicate.NewPredicateFuncs(func(object client.Object) bool {
+		if o, ok := object.(*v1alpha1.RouterNodeConfigurationStatus); ok {
+			return o.Name == n.nodeName
+		}
+		return true
```
RamLavi: shouldn't this be `false`?

ormergi: No, because it mean it will filter **out** all other events.
This function effects node-status CR events only.

api/v1alpha1/routernodeconfigurationstatus_types.go
```diff
@@ -0,0 +1,60 @@
...
+type RouterNodeConfigurationStatusList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+
+	Items []RouterNodeConfigurationStatus `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&RouterNodeConfigurationStatus{}, &RouterNodeConfigurationStatusList{})
+}
```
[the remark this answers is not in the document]
ormergi: Done

API-DOCS.md
```diff
@@ -56,6 +57,67 @@ _Appears in:_
...
+
+
+#### FailedResourceKind
+
+_Underlying type:_ _string_
+
+
+
+_Validation:_
+- Enum: [Underlay L2VNI L3VNI FrrConfiguration]
```
qinqon: Aren't we missing l3passthrough ?

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,166 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	"github.com/openperouter/openperouter/api/v1alpha1"
+)
+
+// +kubebuilder:rbac:groups=openpe.openperouter.github.io,resources=routernodeconfigurationstatuses,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=openpe.openperouter.github.io,resources=routernodeconfigurationstatuses/status,verbs=get;update;patch
+
+// NodeStatusController manage node router status API.
+type NodeStatusController struct {
```
qinqon: I am not sure we need a different controller, at the end we are going to reconciler the node status with the VNI resources, doing at separate controller we may be duplicating a lot of code, also a lot of failure are already detected there.

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,166 @@
...
+		return true
+	})
+
+	// setup channel for triggering initial reconcile iteration to initial CR created
+	initCh := make(chan event.GenericEvent, 1)
+	go func() {
+		initCh <- newEvent(n.nodeName, n.namespace)
+		close(initCh)
+	}()
+
```
qinqon: Don't we have a easier mechanism to create the CRD for the first time ? for example if we end up with just one controller since the PERouterReconciler is already reconciler on corev1.Node there is going to be always a first Reconcile on startup so we end up creating them, this is similar how kubernetes-nmstate daemonsets create the NodeNetworkState

internal/controller/routerstatus/controller.go
qinqon: > enhancement [says](https://github.com/openperouter/openperouter/blob/9bee02cde9e85b76b20d02871b68a3121dac3f23/enhancements/status-crd.md?plain=1#L232) owner reference should be the node, not the pod. 
> We can arguably say this is better lifecycle choice (CR goes away when the controller pod goes) but I'm not sure why did we differ from the enhancement. Can you explain?

Agree the owner has to be the node, the fact that the controller pod crashed can not end up with the CRD being deleted

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,166 @@
...
+			Namespace:       n.namespace,
+			OwnerReferences: []metav1.OwnerReference{podRef},
+		},
+	}
+
+	_, err := ctrlutil.CreateOrPatch(ctx, n.client, nodeStatus, func() error {
+		// TODO: update status according to router configuration resources
+		//  - Fetch the effective router configuration resources selecting the hosting node.
+		//  - Compose status according the effective resources state.
+		nodeStatus.Status = updateStatus(nodeStatus.Status)
```
qinqon: We have all that info inside the PERouterREconciler it will be just about updating the status resource, I think we should not create just another controller for this.

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,166 @@
...
+		Message: "All configuration applied successfully",
+	}
+	degraded := metav1.Condition{
+		Type:    v1alpha1.ConditionTypeDegraded,
+		Status:  metav1.ConditionFalse,
+		Reason:  v1alpha1.ConditionReasonConfigSuccessful,
+		Message: "All configurations are healthy",
+	}
+	return []metav1.Condition{ready, degraded}
+}
```
qinqon: Since this is all placeholder I am not sure if we want to state here "All good" or "Not implemented"

qinqon (changes requested):
First pass, let's clarify controller stuff before even continue

internal/controller/routerstatus/controller.go
ormergi: Good point.
I will change it to reflect unknown state.

internal/controller/routerstatus/controller.go
RamLavi: IIUC ,the only event sent here is RouterNodeConfigurationStatus:
```
func newEvent(name, namespace string) event.GenericEvent {
	return event.GenericEvent{Object: &v1alpha1.RouterNodeConfigurationStatus{
		ObjectMeta: metav1.ObjectMeta{
			Name:      name,
			Namespace: namespace,
		},
	}}
}
```
So you will **never** reach that `return true`.
The risk is future changes. If you add Watches() for other types - they will hit the "return true" for all events.. But I guess you can change that when you add future types..

RamLavi: That said — if the controller is added into PERouterReconciler (as Quique suggested) then this whole code will not be relevant anyways.

internal/controller/routerstatus/controller_test.go
```diff
@@ -0,0 +1,117 @@
...
+	It("should create initial CR", func() {
+		Eventually(func(g Gomega) {
+			key := client.ObjectKey{Name: testCtrlNode, Namespace: testNamespace}
+			actual := &v1alpha1.RouterNodeConfigurationStatus{}
+			g.Expect(testsClient.Get(context.Background(), key, actual)).To(Succeed())
+			sanitizeNodeStatus(actual)
+
+			expected := newTestCR(testCtrlNode, testNamespace)
+			res := equality.Semantic.DeepEqual(actual, expected)
+			g.Expect(res).To(BeTrueBecause("expected and actual CRs should be equal"))
```
RamLavi: if this fails we won't get a visible diff on the test output. Consider using Gomega's `To(Equal(...))`

internal/controller/routerstatus/controller_test.go
```diff
@@ -0,0 +1,117 @@
...
+		ctx, cancel := context.WithCancel(context.Background())
+		DeferCleanup(cancel)
+		go func() {
+			GinkgoRecover()
+			By("start manager")
+			Expect(mgr.Start(ctx)).To(Succeed())
+		}()
+	})
+
+	It("should create initial CR", func() {
```
RamLavi: Should we also check for "CR already exist", "CR recreated after deletion" cases?

api/v1alpha1/routernodeconfigurationstatus_types.go
```diff
@@ -0,0 +1,58 @@
...
+	)
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=routernodeconfigurationstatuses
+// +kubebuilder:resource:scope=Namespaced
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`,description=Ready
+// +kubebuilder:printcolumn:name="Degraded",type=string,JSONPath=`.status.conditions[?(@.type=="Degraded")].status`,description=Degraded
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
```
maiqueb, line 15: just curious: what is this column's purpose ? what do you plan to use this info for ?

internal/controller/routerstatus/controller.go
ormergi: In this scenario, controller crushed, we will end up with outdated CR doesnt reflect the actual state.

Having the onwer ref on the pod will make sure no outdated CR left dangling.
Once the controller recovers it will create the CR with reflecting the actual state.

This is also true for when the controller pod placement change (nodeSelector / node label chagnes, node deletion / drain) in a scenario the CR has node ref only the CR will left dangled.

Just one note about having node ref for node association, the CR name is as the node name so having the node ref is not much effective.

internal/controller/routerstatus/controller.go
ormergi: It will reach true on controller instances mathcing the node name.
Also note that e2e tests verifying CR creating are green. And I also verified each controller instance handle events for its hosting node only.

> The risk is future changes. If you add Watches() for other types - they will hit the "return true" for all events.. But I guess you can change that when you add future types..

This filtering doesnt contradicts with other Watch() options, because it return true by default meaning event will get handled. For node events it return true based on the mentioned condition.

internal/controller/routerstatus/controller.go
ormergi: The intention was to follow [#159](https://github.com/openperouter/openperouter/pull/159/changes/fb4fd41f720ba5f8850360a23dd5ecfa1d111809) example, where we introduce status-manager entity injected to the existing and introduced controllers:
- api config controller, report all types of events through the status-manager
- node-status controller, on events read the status-manager state and compose the node-status accordingly.

No code duplication whatsoever.
And the key benefit here is better modularity and proper separation for responsibilities.

WDYT?

internal/controller/routerstatus/controller.go
ormergi: I followed frr-k8s controller example, where a generic event is injected to kick-off initial CR creation.
This channel would later be used for communication between PERouterReconciler to the node-status controller.

internal/controller/routerstatus/controller.go
ormergi: The intention was to follow [#159](https://github.com/openperouter/openperouter/pull/159/changes/fb4fd41f720ba5f8850360a23dd5ecfa1d111809) and setup communication channel between PERouterREconciler and the node-status contorller to avoid double validations.

internal/controller/routerstatus/controller_test.go
ormergi: Initially I tired with Equal but it doesnt provide better results because it trim the output of the objects.
And I am not sure it does semantic equality check.

internal/controller/routerstatus/controller_test.go
ormergi: Looking at this again, I wonder if we even need this test at all, currently there is no much business logic to test.
This actually an integration test using envtest, its slow and has limitations because its not an actual cluster.
The added e2etests covers the scenarios you mentioned, and they are not expensive as no workloads are involved (pod/vm)

API-DOCS.md
ormergi: I am not sure tbh, there is nothing about it on the enhancment. 
Was it introduced later?

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil
+	})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func updateStatus(actualStatus v1alpha1.RouterNodeConfigurationStatusStatus) v1alpha1.RouterNodeConfigurationStatusStatus {
```
maiqueb: imho an "update" call returns an error. 

in this one, you're creating the status. this seems to be a factory function or something like that.

internal/controller/routerstatus/controller.go
```diff
@@ -56,13 +59,30 @@ func (n *NodeStatusController) SetupWithManager(mgr ctrl.Manager) error {
 		return true
 	})
 
+	// setup channel for triggering initial reconcile iteration to initial CR created
+	initCh := make(chan event.GenericEvent, 1)
+	go func() {
+		initCh <- newEvent(n.nodeName, n.namespace)
+		close(initCh)
+	}()
```
maiqueb: question, maybe I'm overthinking: shouldn't this be on the cmd/... ? i.e. whoever is instantiating the `NodeStatusController` is responsible for initializing it. 

Like, decouple creation of the object from starting it.

internal/controller/routerstatus/controller_test.go
```diff
@@ -0,0 +1,117 @@
...
+		By("create test namespace")
+		ns := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNamespace}}
+		Expect(testsClient.Create(context.Background(), ns)).To(Succeed())
+		DeferCleanup(func() {
+			Expect(testsClient.Delete(context.Background(), ns)).To(Succeed())
+		})
+
+		testController := routerstatus.NewNodeStatusController(
+			testsClient, slog.Default(),
+			testCtrlNode, testNamespace,
```
maiqueb: one entry per line please.

e2etests/tests/node_status.go
```diff
@@ -0,0 +1,168 @@
...
+		cs = k8sclient.New()
+
+		var err error
+		allNodes, err = k8s.GetNodes(cs)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(allNodes).NotTo(BeEmpty(), "should have at least one node")
+
+		controllerPods, err = k8s.PodsForLabel(cs, openperouter.Namespace, openperouter.ControllerLabelSelector)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(controllerPods).NotTo(BeEmpty(), "should have at least one controller pod")
```
maiqueb: aren't controller pods a daemonset ? shouldn't we assert we have as many controller pods as we have nodes ?

e2etests/tests/node_status.go
```diff
@@ -0,0 +1,168 @@
...
+
+		Eventually(assertNodeStatusesMatchExpected).
+			WithArguments(controllerNamespace, expectedNodeStatuses).
+			WithTimeout(1 * time.Minute).
+			WithPolling(1 * time.Second).
+			Should(Succeed())
+	})
+
+	It("should reconcile RouterNodeConfigurationStatus", func() {
+		expectedNodeStatuses := newNodeStatuses(controllerNodes, controllerPods, controllerNamespace)
```
maiqueb: IMHO it becomes hard to reason about the test because I have no idea what the `expectedNodeStatuses` are.  

I think we need to either shove these in front of the readers eyes, or have multiple properly named functions that create these status (e.g. `successfulStatus()` , `underlayFailedStatus()` etc).

e2etests/tests/node_status.go
```diff
@@ -0,0 +1,168 @@
...
+	})
+})
+
+func assertNodeStatusesMatchExpected(g Gomega, namespace string, expected []v1alpha1.RouterNodeConfigurationStatus) {
+	GinkgoHelper()
+	actual := &v1alpha1.RouterNodeConfigurationStatusList{}
+	g.Expect(Updater.Client().List(context.Background(), actual, &client.ListOptions{Namespace: namespace})).To(Succeed())
+
+	sanitizeNodeStatus(actual.Items)
+	g.Expect(actual.Items).To(ConsistOf(expected))
```
maiqueb: nit: maybe use WithTransform. 

But that SUPER debatable, and opinionated. You can tell me to buzz off.

internal/controller/routerstatus/controller.go
maiqueb: @qinqon are you worried about any particular duplication ? 

I think isolating the functionality between the multiple controllers is nice to have - assuming we won't have to replicate more than the typical controller-runtime boilerplate ... 

Having said that, we'd need for sure to see a list of which resources are reconciled by which controller to make an informed decision.

internal/controller/routerstatus/controller.go
```diff
@@ -88,12 +96,17 @@ func (n *NodeStatusController) Reconcile(ctx context.Context, req ctrl.Request)
 	logger.Info("start reconcile")
 	defer logger.Info("end reconcile")
 
+	// set owner reference pointing to the hosting pod to have the CR garbage-collected
+	// when the hosting pod is deleted (due to controller placement change, node drain/taint/removal).
+	podRef := newPodOwnerReference(n.podName, n.podUID)
```
maiqueb: can we be more specific ?... like, which pod ?

internal/controller/routerstatus/controller.go
maiqueb: > In this scenario, controller crushed, we will end up with outdated CR doesnt reflect the actual state.

How do you know ? The controller could crash without impacting the dataplane. Or **existing** control plane. Yes, new VNIs will **not** be applyed, but, whatever state the node had will be preserved.

> This is also true for when the controller pod placement change (nodeSelector / node label chagnes, node deletion / drain) in a scenario the CR has node ref only the CR will left dangled.

Hm ... the controller is a daemonset pod. meaning, you won't see controllers moving to other nodes or anything like that.

internal/controller/routerstatus/controller.go
maiqueb: this is debatable if you ask me. 

a simpler mechanism for the initial sync means that you **must** have 2 mechanisms: 
- one for the initial sync
- another for everything else

On paper, might be simpler, but I am inclined to say I find this approach (single mechanism) to be more elegant.

@qinqon do you foresee any trouble here ?

internal/controller/routerstatus/controller.go
maiqueb: why didn't you choose the router pod instead ? that's where both the data plane & control plane are running.

maiqueb (changes requested):
Some questions, and requesting some changes. 

Two things should be hashed out / defined ASAP:
- should we have a separate controller ? **I** think so. Unless ...
- should the owner of the status CR be the controller, the router, or the node ?... I think we should follow the enhancement and track the node.

internal/controller/routerstatus/controller.go
RamLavi: I agree with @qinqon here. Looking at what phases 2+ will require (computing status at the end of each reconcile based on which resources failed), the status CR creation and update logic belongs inside PERouterReconciler.
That reconciler already knows its node name and has all the context needed to create the status CR on first reconcile. A separate controller adds complexity like:
  - Channel-based [synthetic triggers](https://github.com/openperouter/openperouter/pull/355/changes/33dd7e3e8f79d9ca8c9193ba66eed57d626b5839#diff-a28e21dbf189d68ff35c19e9b68b68c23715b1501f558217cc02834a03ca2d6dR77)
  - [predicates](https://github.com/openperouter/openperouter/pull/355/changes/18853208d2c20e1d56487dcf7c953e1303c1e048#diff-a28e21dbf189d68ff35c19e9b68b68c23715b1501f558217cc02834a03ca2d6dR52)

Please consider this simpler approach: ensureStatusCR at the top of PERouterReconciler Reconcile() to guarantee the CR exists, then updateStatus at the bottom once we have reconcile results. This keeps everything in one place and maps directly to Phase 3's requirement of "status computed at the end of each reconcile".
                                                                            
For Phase 1, we'd just need ensureStatusCR creating the CR with empty status. The update logic comes in later phases.

ormergi:
Rebased - [push](https://github.com/openperouter/openperouter/compare/df7b80d9f89f61d4ead6f0d38305b4737ba08d15..52c4e5a19e09e8b539b71d81a51fc6afd0df4869)
Fix linter issues - [push](https://github.com/openperouter/openperouter/compare/52c4e5a19e09e8b539b71d81a51fc6afd0df4869..cc59fa36cd2dd65a120ab8a19331d431e4ef532a)

api/v1alpha1/routernodeconfigurationstatus_types.go
ormergi: Done

api/v1alpha1/routernodeconfigurationstatus_types.go
ormergi: All except for L3Passthrough its not mentioned in the enhancement.
Not sure if and how it should be handled.

api/v1alpha1/routernodeconfigurationstatus_types.go
ormergi: The condition Reason is pogrematic identifier CamelCase word.
The intention is to have some upper bond,  50 should be long enough. If not it can be bumped.

api/v1alpha1/routernodeconfigurationstatus_types.go
ormergi: printcolumn define `kubectl get` output for the reasorce, here we add Ready and Degraded (in addition to the default Age):
```
NAME                    READY     DEGRADED   AGE
pe-kind-control-plane   Unknown   Unknown    1h
pe-kind-worker          Unknown   Unknown    1h
```

internal/controller/routerstatus/controller.go
ormergi: The PR doesnt contain the desired business logic yet and the existing code doesnt produce errors.
Once the desired logic is added this function can change to return error.

internal/controller/routerstatus/controller.go
ormergi: Currently this is the only place that require initialization, I am not sure about the benefit of exporting it at the moment.

cmd/ instantiate the controller, register it to the controller-runtime manager (SetupWithManager) and start the manager.

The thing is controller-runtime doesnt seem to provide native solution / API for initialization phase.
Once the controller-runtime's manager starts (Start() called by cmd/..), the controller starts with clean workqueue.

internal/controller/routerstatus/controller.go
ormergi: I am going to removed this, we agreed no need for pod onwer ref. 
Node onwer ref should be sufficient since those pods are DameonSet and there will be single replica on each node.

internal/controller/routerstatus/controller.go
ormergi: Done

internal/controller/routerstatus/controller.go
ormergi: I missed that this is a deamonset, single replica per each node, no need for pod ref at all, I will change it to node owner ref.  Sorry about the noise.

internal/controller/routerstatus/controller.go
ormergi: I rather avoid one big controller that does everything, it tend to grow and become hard to maintain.
Having single controller per CRD is the best practice.

Take for example frr-k8s and metallb, similar and well established projects, provides configuration and node-status APIs, where wach CRD managed by different controllers, running on the same pod:
- metallb provides the ConfigurationState CRD for node status, managed by dedicated [controller](https://github.com/metallb/metallb/blob/main/internal/k8s/controllers/configuration_state_controller.go#L148-L149)
- frr-k8s provides the FRRNodeState CRD back for node status, managed by dedicated [controller](https://github.com/metallb/frr-k8s/blob/main/internal/controller/frrstate_controller.go#L140-L141)

The enhancement doesnt place hard requirement for which controller to manage the CRD, 
the same can be achieve the same with two controllers. The actual implementation can be opinionated.

Having said that I am not rejecting your suggestion it valuable and make sense, in this particular case I thin the project will benefit from having dedicated controller for node-state for the long term. I want to give it a trial.

As a side note I think we should switch mental mode from https://github.com/openperouter/openperouter/pull/345

internal/controller/routerstatus/controller.go
ormergi: Please see https://github.com/openperouter/openperouter/pull/355#discussion_r3215581563

internal/controller/routerstatus/controller_test.go
ormergi: Done

e2etests/tests/node_status.go
ormergi: I changed how expected objects created to reflect the expected status state

e2etests/tests/node_status.go
ormergi: Done

e2etests/tests/node_status.go
ormergi: Fixed

internal/controller/routerstatus/controller_test.go
ormergi: Done

internal/controller/routerstatus/controller.go
ormergi: The CR will have onwer ref to the node instead of pod.
Done

ormergi:
Resolve review comments - [diff](https://github.com/openperouter/openperouter/compare/cc59fa36cd2dd65a120ab8a19331d431e4ef532a..2bea8dee71e5e72530c8f2b2e7c658f577979990)
@qinqon @RamLavi @maiqueb PTAL

internal/controller/routerstatus/controller.go
qinqon: @maiqueb @ormergi I don't think this controller should re-validate Underlays/VNIs — the `PERouterReconciler` already does that work and is the only place that sees runtime errors (e.g. netlink failures, FRR reload issues). Duplicating validation here adds complexity without benefit.
The simplest approach is to call `Status().Update()` at the end of `PERouterReconciler.Reconcile()` directly.
If we want to keep a separate controller, I see two options:
1. **Status controller handles lifecycle only** (create on startup, recreate if deleted), while `PERouterReconciler` updates the status subresource directly. The status controller just ensures the CR exists.
3. **`PERouterReconciler` communicates with the status controller via a channel** (there's already a channel-based watcher in place). The reconciler pushes results through a `Reporter` struct and the status controller converts them into the CRD status. This is similar to what MetalLB does with `Layer2StatusChan` / `BGPStatusChan`.In either case, the status data should originate from `PERouterReconciler` — it's the single source of truth for which resources failed and why.

internal/controller/routerstatus/controller.go
qinqon: @ormergi @maiqueb since now the owner is the node, we should reconcile on node creation too thi smeans that at starup controller-runtime will issue "syntetic" reconciles of nodes and this will trigger the status CRD creation, meaning that we don't need the creation mechanism on this PR if we watch for node's creation (wich we should)

API-DOCS.md
qinqon: Nope, but if is not parth of enhacement maybe we should not implement it ? 

@maiqueb do you know why l3passthough was not part of the enhacement ?

internal/controller/routerstatus/reporter.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+}
+
+func (r *Reporter) GetChannel() chan event.GenericEvent {
+	return r.ch
+}
+
+func (r *Reporter) GetFailedResources() []v1alpha1.FailedResource {
+	return []v1alpha1.FailedResource{}
+}
```
qinqon: We are not using this at this PR, maybe this should be done by the follow up wire in PR ?

e2etests/pkg/openperouter/openperouter.go
```diff
@@ -17,6 +17,7 @@ import (
 const (
 	Namespace           = "openperouter-system"
 	routerLabelSelector = "app=router"
+	ControllerLabelSelector = "app=controller"
```
qinqon: We are not using this, if the e2e test runs towards a runOnMaster: false installation they will fail.

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,167 @@
...
+		close(initCh)
+	}()
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.RouterNodeConfigurationStatus{}).
+		WatchesRawSource(source.Channel(initCh, &handler.EnqueueRequestForObject{})).
+		WithEventFilter(hostingNodeStatusEvents).
+		Named("routernodeconfigurationstatus").
+		Complete(n)
+}
```
qinqon: We should watch node creation too here so we don't need the initialization channel, also is the natural way to do things.

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,167 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: n.nodeName, Namespace: n.namespace},
+	}
+	if _, err := ctrlutil.CreateOrUpdate(ctx, n.client, nodeStatus, func() error {
+		return ctrlutil.SetOwnerReference(node, nodeStatus, n.client.Scheme())
+	}); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// TODO: update status according to the node effective router configuration resources
+	//  Fetch configuration resources state from OpenPEController (via channel), compose status accordingly.
```
qinqon: This is weird, The channgel should not be used to "fetch" it should be use for the OpenPERouter controller to push status to the reconciler (adding a watcher for that channel).

Like are we going to have the OpenPERouter state in memory ? 

Not sure if all this is worth it.

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,167 @@
...
+		return ctrl.Result{}, nil
+	}
+	nodeStatus.Status = newStatus
+	if err := n.client.Status().Update(ctx, nodeStatus); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update status: %w", err)
+	}
+	logger.Info("node router status updated successfully", "node", n.nodeName)
+
+	return ctrl.Result{}, nil
+}
```
qinqon: I think we should use Server Side Apply, but maybe for phase 2 ? also we should introduce conflict resolutions here.

making it requebe without error.

qinqon (changes requested):

API-DOCS.md
maiqueb: I don't, but I suspect i totally forgot about that resource type.

internal/controller/routerstatus/controller.go
maiqueb: Spoke with @qinqon out of band, and we both concluded / agreed that a single reconcile loop that computes the status from the result of the reconcile is simpler / beneficial, since you don't have to come up w/ any scheme to pass the result of one reconcile loop to the other. 

@ormergi feel free to try to convince us otherwise, but, right now, we think it's more complex for no good reason.

internal/controller/routerstatus/reporter.go
ormergi: Sorry it was added by accident, I will remove it in next push.

internal/controller/routerstatus/controller.go
ormergi: This comment is bad and confusing I will remove it.

At fiest I tought its possible to piggy back the failed resources list over the generic event channel.
Turns out since it expect client.Object it trims the concrete type status making this impossible.

There are two options here:
1. Into middleman component, both OpenPERouter and node-status controller composite.
  On OpenPERouter reconcile it will notify about failing resources, and the middleman component to cache it.
  On node-status reconcile read the failed resources from the middleman (cached) and update status accordingly.
2. Have OpenPERouter update node-status `status.failedResources`, and the node-status controller in turn will update the conditions.

Option #2 is better the #1 because it doesnt require caching.
Although I am not sure if we will eventually some sort of local cache at some point.

WDYT?

internal/controller/routerstatus/controller.go
RamLavi: My 2cents: I think that since the nature of this CRD is to give a node scoped summary of the results of the PERouterReconciler controller, then it only makes sense to not separate.

ormergi: Thanks for the feedback, I appreciate it :pray: 

Just on last thing hear me out:
I think a good middle ground here would be option [#1](https://github.com/openperouter/openperouter/pull/355#discussion_r3217667307) - have the status controller handle the lifecycle only and the PERouterReconciler to update the status: 
create on startup, recreate if deleted, update condition according to failed-resources (and future input).

This approach saves the mechanics of moving the required info between the controllers whole allowing separating concepts.
This way we can balance between separation of responsibilities and complexity, having minimal footprint on PERouterReconciler.

WDYT?

maiqueb: I don't see how that's better. At the end of the day, all we need is a `defer updateStatus(reconcileResult)` in the reconcile loop.

I recommend trying the simplest approach first (which requires zero boilerplate) and only deviating from it if we encounter issues.

Separation of concerns isn't a strong enough argument for me - and to be honest, I don't think computing and updating the status of the reconcile loop is a different concern from the reconcile loop itself.

API-DOCS.md
qinqon: I see it's included at Ram's PR so I think it should be included here,

internal/controller/routerstatus/controller.go
qinqon: I don't think having two controllers communicating over channels complexity is really worth it to be honest.

internal/controller/routerstatus/controller.go
qinqon: Not worth it, I would start just implementint at the OpenPERouter controller, then we can see in a follow up if it's worth it to have different controller for that.

API-DOCS.md
ormergi: Done, in case we need it which we likely will.

internal/controller/routerstatus/controller.go
ormergi: FWIW I meant separation of concerns in a sense that the status in this case is a whole different CRD, thus should be handled by some other component, in the cost of complexity.

Fail enough, we should stick with simplest option. Thanks for the discussion.

Done
The new controller is removed, relevant logic moved to the configuration controller.

internal/controller/routerstatus/controller.go
ormergi: Done 
This change has been removed, CR will create because the controller watch nodes.

internal/controller/routerstatus/controller.go
ormergi: The PR changed to extend the  configuration controller instead.

internal/controller/routerstatus/controller_test.go
ormergi: Those tests were removed altogether in favor of e2e tests.

api/v1alpha1/routernodeconfigurationstatus_types.go
ormergi: I removed those types because they can be resolved via the CR TypeMeta.
The FailedResourceKind validation is updated to allow L3Passthrough.

internal/controller/routerstatus/controller.go
ormergi: This changes has been removed as the PR extend the exiting controller instead of adding new one.

internal/controller/routerstatus/controller.go
ormergi: Thanks for the feedback, I changed the PR to extend the exiting controller instead of adding new one.

internal/controller/routerstatus/controller.go
ormergi: This change was removed following pivot to extend existing controller.

e2etests/pkg/openperouter/openperouter.go
ormergi: Done

internal/controller/routerconfiguration/node_status.go
```diff
@@ -0,0 +1,110 @@
...
+	res, err := ctrlutil.CreateOrUpdate(ctx, r.Client, nodeStatus, func() error {
+		return ctrlutil.SetOwnerReference(node, nodeStatus, r.Scheme)
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create or update node status: %w", err)
+	}
+
+	r.Logger.Info("successfully reconciled node router status", "op:", res, "node", r.MyNode)
+
+	return nil
```
qinqon: This is not needed we should lazy create/update the resource at the updateNodeStatus  so we lazy create at the end of reconcile

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -84,6 +86,10 @@ func (r *PERouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 
 	ctx = context.WithValue(ctx, requestKey("request"), req.String())
 
+	if err := r.reconcileNodeStatus(ctx); err != nil {
```
qinqon: We don't need this to be here since we are going to use update the status just at the end, in fact let's just remove the function and we lazy create/update the CRD at the updateStatus function, so we nee just one "hook" for status at the Reconcile function.

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -122,16 +128,26 @@ func (r *PERouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 		}
 	}
-	if err != nil {
-		slog.Error("failed to configure the host", "error", err)
+	errs := []error{}
+	if reconcileErr != nil {
+		errs = append(errs, reconcileErr)
+	}
+
+	if perr := r.updateNodeStatusResource(ctx); perr != nil {
```
qinqon: Is this enough ? or it's better to put it on a "defer" at the beginning so we catch the returns and reflect on status ?

internal/controller/routerstatus/controller.go
```diff
@@ -0,0 +1,147 @@
...
+		client:    client,
+		logger:    logger,
+		nodeName:  nodeName,
+		namespace: namespace,
+	}
+}
+
+// SetupWithManager registers NodeStatusController to controller-runtime manager.
+// Watches RouterNodeConfigurationStatus events of the hosting node.
+func (n *NodeStatusController) SetupWithManager(mgr ctrl.Manager) error {
```
qinqon: Are we removing this controller ?

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/2bea8dee71e5e72530c8f2b2e7c658f577979990..ae062744a0fd934610984e3bff4d281a4854fb7a)): resolve review comments:
- Submerge the proposed controller logic to the existing configuration status
- Rm failed-resource kind constants (the can be resolved on runtime, we can add it upon necssasity)
- Update failed-resource kind validation to allow L3Passthough

qinqon (changes requested):

internal/controller/routerconfiguration/underlay_vni_controller.go
ormergi: The node-status updated will depend on the result of Reconsile() about (the one that actually configures the node) so I dont think we should more this to defer() at the moment.

internal/controller/routerstatus/controller.go
ormergi: Sure, it was removed on last push

internal/controller/routerconfiguration/underlay_vni_controller.go
ormergi: Done

internal/controller/routerconfiguration/node_status.go
ormergi: Done

ormergi:
[Rebase](https://github.com/openperouter/openperouter/compare/ae062744a0fd934610984e3bff4d281a4854fb7a..57b391a1adb8e29ca64611d5de77bd92523dbd60)

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/57b391a1adb8e29ca64611d5de77bd92523dbd60..2d185c5829c8eafa0273f36b344316be5e26a304): address https://github.com/openperouter/openperouter/pull/355#pullrequestreview-4272412570 
@qinqon PTAL

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -122,16 +124,26 @@ func (r *PERouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+
+	if err := r.reconcileNodeStatus(ctx); err != nil {
+		errs = append(errs, err)
+	}
+
+	if len(errs) > 0 {
+		err := errors.Join(errs...)
+		r.Logger.Error("failed to reconcile host configuration", "error", err)
 		return ctrl.Result{}, err
 	}
```
qinqon: with defer you don't need any of this, we will just do the best we can to update the status to whatever we have access and that's it, normally something like a Reporter struct at the reconciler struct or the like, we will see on follow up PRs.

qinqon (changes requested):

internal/controller/routerconfiguration/underlay_vni_controller.go
ormergi: Calling `reconcileNodeStatus()` on defer doesnt allow proper error handling (defer has no return value),
in a scenario of failure (create/update fail) it wont trigger reconciliation leaving the CR not exist/handled.

maiqueb: I'd say this is the only time I can think of a decent usage for named return parameter.

ormergi: Please dont go this path, it will make the controller code harder to understand I dont think its worth it.

Maybe I am missing something, its hard to tell because we dont have all the desired business logic.
Should like optimizations we can do later.

internal/controller/routerconfiguration/node_status.go
```diff
@@ -0,0 +1,95 @@
...
+	if equality.Semantic.DeepEqual(nodeStatus.Status, newStatus) {
+		return nil
+	}
+
+	newNodeStatus := nodeStatus.DeepCopy()
+	newNodeStatus.Status = newStatus
+
+	if err := r.Status().Patch(ctx, newNodeStatus, client.MergeFrom(nodeStatus)); err != nil {
+		return fmt.Errorf("failed to patch status: %w", err)
+	}
```
maiqueb, line 54: should we do this in a `retry` block ?

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -122,16 +124,26 @@ func (r *PERouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	if reconcileErr != nil {
+		errs = append(errs, reconcileErr)
+	}
+
+	if err := r.reconcileNodeStatus(ctx); err != nil {
+		errs = append(errs, err)
+	}
+
+	if len(errs) > 0 {
+		err := errors.Join(errs...)
```
maiqueb, line 145: @ormergi if we're only going to log the error, I wonder why you're taking this trouble. 

Couldn't we log the 2 errors separately ? 

just wondering, I don't have a strong opinion.

internal/controller/routerconfiguration/node_status.go
```diff
@@ -0,0 +1,95 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create or update node status: %w", err)
+	}
+
+	newStatus := updateStatus(nodeStatus.Status)
+
+	if equality.Semantic.DeepEqual(nodeStatus.Status, newStatus) {
+		return nil
+	}
```
qinqon, line 47: kwe don't need that MergeFrom return empty in this case so fhte Patch is a no-op

internal/controller/routerconfiguration/node_status.go
qinqon: I would just retry on conflicts for now, not sure we want to retry per se, also not sure we want to Reconcile on update status error (although we are suppose to be idempotent).

qinqon (changes requested):

internal/controller/routerconfiguration/node_status.go
maiqueb: fair enough, merge on conflicts seems the way to go. Tks for being more specific !

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -122,16 +124,26 @@ func (r *PERouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 		}
 	}
-	if err != nil {
-		slog.Error("failed to configure the host", "error", err)
+	errs := []error{}
+	if reconcileErr != nil {
+		errs = append(errs, reconcileErr)
+	}
+
+	if err := r.reconcileNodeStatus(ctx); err != nil {
```
qinqon, line 140: **Consider using `defer` with named return values for `reconcileNodeStatus`**

MetalLB uses `defer` so the status write always happens regardless of how the reconcile exits — no need to remember to call it before every `return`:

```go
func (r *PERouterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, retErr error) {
    defer func() {
        if err := r.reconcileNodeStatus(ctx); err != nil {
            r.Logger.Error("failed to update node status", "error", err)
            retErr = errors.Join(retErr, err)
        }
    }()

    // ... rest of reconcile logic ...
}
```

See MetalLB's [`config_controller.go:62-75`](https://github.com/metallb/metallb/blob/main/internal/k8s/controllers/config_controller.go#L62-L75) for reference.

internal/controller/routerconfiguration/underlay_vni_controller.go
qinqon, line 145: Joining the status write error into the return (triggering a full re-reconcile) is the right call. MetalLB does the same thing — the config reconcile is idempotent so the extra work is cheap, and it avoids the status CR staying permanently stale if you just log-and-forget.

qinqon (changes requested):
Let's follow metallb approach is not very complicated and make sense.

internal/controller/routerconfiguration/node_status.go
ormergi: The intention here is to avoid initiating redundant API calls, reducing failure points

internal/controller/routerconfiguration/underlay_vni_controller.go
ormergi: Note that the aggregating error is logged and returned.
We can log each error at the time it was detected (lines 137, 140).
I dont have strong opinion about it, lemee know if you prefer to log each error separately.

internal/controller/routerconfiguration/underlay_vni_controller.go
ormergi: @qinqon  Thanks for the reference 

I think this pattern is kind of missus of the defer mechanism, which should be use for resource release or handling panics - operations consist of almost zero logic. The 'reconcileNodeStatus' hold pretty complex logic.
In addition updating the node-status is fundamental part of the desired workflow, I think spreading it across the Reconcole method makes it harder to follow , reminds me of the goto pattern cons.

At the moment  we dont have all the desired business logic yet, and its hard to tell the full picture.
Would you please reconsider this change for a follow-up work, once we have phase2/3 bits?

RamLavi, line 140: do we want to fail the entire reconcile if the node-status patching failed? I tend to think we should retryOnconflict, and for other errs - only slog.. @ormergi wdyt?

qinqon: I don't agree with the perception, but I am ok if we do the `defer` at phase2 follow up.

qinqon: I think we should Reconcile should be idempotent

ormergi: In general I think the controller should use RetayOnConflict. The way I see it, using RetryOnConflict is patching an actual bad due to race or contradicting component.
As for now the only component that expected to patch the node-status is the configuration controller.
When patch fail it should return error to tigger reconciliation ,keeping the process idempotent.

internal/controller/routerconfiguration/node_status.go
ormergi: Please see https://github.com/openperouter/openperouter/pull/355#discussion_r3235192077

internal/controller/routerconfiguration/underlay_vni_controller.go
maiqueb: > In general I think the controller should use RetayOnConflict. The way I see it, using RetryOnConflict is patching an actual bad due to race or contradicting component. As for now the only component that expected to patch the node-status is the configuration controller. When patch fail it should return error to tigger reconciliation ,keeping the process idempotent.

It's not about what's "expected", it's about being resilient to unexpected things.

What exactly can go wrong with the retry on conflict ? Really trying to understand what you're trying to protect here.

ormergi: If something go wrong returning an error will trigger reconciliation and the the state will be resolved and it should be resilient.

RetryOnConflitct will silent potential races and unexpected patches/updates done by the controller in some different code path, or by other controllers.

ormergi: I think we should relay on native runtime mechanism,  instead of intervening and use RetryOnConflict in a middle of handling request.
The runtime will reenqueue gracefully with a backoff w/o blocking the controller workers allowing it to handle other requests.
Next time a conflict request is handled, it will use the latest object revision. 

Do we really need to intervent  in how the runtime handle conflicts?

RamLavi: I think this is a good use case for the defer, so I'll add it to the phase2, and continue the conversation there if you still have strong objection..

maiqueb: Fine, it's settled then from my perspective. 

Resolving this.

internal/controller/routerconfiguration/underlay_vni_controller.go
RamLavi: if both @qinqon and @ormergi agrees then I'll keep this approach on phase2

internal/controller/routerconfiguration/underlay_vni_controller.go
fedepaol: agree, -1 on named returns

internal/controller/routerconfiguration/node_status.go
```diff
@@ -31,14 +33,63 @@ func (r *PERouterReconciler) reconcileNodeStatus(ctx context.Context) error {
 	nodeStatus := &v1alpha1.RouterNodeConfigurationStatus{
 		ObjectMeta: metav1.ObjectMeta{Name: r.MyNode, Namespace: r.MyNamespace},
 	}
-	res, err := ctrlutil.CreateOrUpdate(ctx, r.Client, nodeStatus, func() error {
+	_, err := ctrlutil.CreateOrUpdate(ctx, r.Client, nodeStatus, func() error {
```
fedepaol, line 36: we should remove this from the first commit upfront

e2etests/tests/node_status.go
```diff
@@ -0,0 +1,156 @@
...
+	})
+
+	It("should create RouterNodeConfigurationStatus for each node running router", func() {
+		expected := newNodeStatuses(nodes, routerNamespace,
+			withConditions(newUnknownConditions()))
+
+		assertNodeStatusesMatchExpected(routerNamespace, expected)
+	})
+
+	It("should re-create RouterNodeConfigurationStatus when deleted", func() {
```
fedepaol: with this test the previous is not necessary

e2etests/tests/node_status.go
```diff
@@ -0,0 +1,156 @@
...
+		expected := newNodeStatuses(nodes, routerNamespace,
```
fedepaol: we always pass withConditions(newUnkonwnConditions), so imo it's an overkill. Let's just create them

e2etests/tests/node_status.go
```diff
@@ -0,0 +1,156 @@
...
+			withConditions(newUnknownConditions()))
+
+		assertNodeStatusesMatchExpected(routerNamespace, expected)
```
fedepaol: I'd simplify this by just looping over the nodes and ensuring we have a node status for each node (and no more) unless I am missing something

.github/workflows/ci.yaml
```diff
@@ -269,7 +269,7 @@ jobs:
         run: |
           if [ ${{ matrix.deployment }} = "systemdmode" ]; then
             export TEST_ARGS="--systemdmode"
-            export GINKGO_ARGS="--skip='Router Host configuration' --skip='North/south traffic after FRR container restart'"
+            export GINKGO_ARGS="--skip='Node Router Status' --skip='Router Host configuration' --skip='North/south traffic after FRR container restart'"
```
fedepaol: node status should work also on systemd mode, at least from when the apiserver is available

internal/controller/routerconfiguration/underlay_vni_controller.go
ormergi: This is a must, otherwise the CR will not be handled in case of an error.

e2etests/tests/node_status.go
ormergi: Although it may look overkill, the intention was to set some foundations for generating the expected CR as a reference, we will certainly need this for upcoming PRs. 
Lemme know if it make sense or you rather change it to what you suggest.

.github/workflows/ci.yaml
ormergi: I suppose we should move reconcileNodeStatus() call to the common reconcile.go.Reconcoile(), to have single entry point for the node-status then.

e2etests/tests/node_status.go
ormergi: The intention was to add a reference test assert on the expected CR state, currently the status is static (always have Unkown condition), once the additional business logic is added this test could evolve and replicated to cover addition scenarios.

ormergi: Lemme know if it make sense or you would rather change to what you suggested above.

ormergi:
[Rebase](https://github.com/openperouter/openperouter/compare/4acedd76af477306b6b149a95714d4c1c4522221..8989a33c3102cc3119c044c51d26d20b3191254e)

internal/controller/routerconfiguration/node_status.go
ormergi: Done

e2etests/tests/node_status.go
ormergi: Done

.github/workflows/ci.yaml
ormergi: Done

ormergi: Please ignore my last comments, I reverted these changes and the skip.

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/8989a33c3102cc3119c044c51d26d20b3191254e..4a936b437725f8c1d98a895a61445c837755ae01): resolve https://github.com/openperouter/openperouter/pull/355#pullrequestreview-4290264804 https://github.com/openperouter/openperouter/pull/355#pullrequestreview-4290491512

On hostmode the config controller would not pass down its namespace name, making the controller fail to create the node-status CR, it is fixed now and e2etests hostname job is green.

e2etest help job failed due to flake, not related to this PR changes https://github.com/openperouter/openperouter/issues/400

ormergi:
@qinqon @maiqueb @fedepaol could you please have another look?

internal/controller/routerconfiguration/underlay_vni_controller.go
maiqueb: > If something go wrong returning an error will trigger reconciliation and the the state will be resolved and it should be resilient.

True.

> 
> RetryOnConflitct will silent potential races and unexpected patches/updates done by the controller in some different code path, or by other controllers.

I don't think the mentioned different code path even exists, and that other controllers are also not interested in these CRDs. Still, the proper answer for the latter would be server side apply, which we have on the roadmap - #427 .

Having said that, I am OK with relying on a re-queued reconcile. If we see it's an issue, we can always change that.

qinqon (approved):
LGTM

e2etests/tests/node_status.go
maiqueb: I agree this might help out on https://github.com/openperouter/openperouter/pull/423. 

Still the point has been made: in this PR, it's borderline pointless.

internal/controller/routerconfiguration/node_status.go
maiqueb: @qinqon thoughts ?

internal/controller/routerconfiguration/node_status.go
maiqueb: I'm OK with relying on k8s.

internal/controller/routerconfiguration/underlay_vni_controller.go
maiqueb: ok we can keep it as is. if it proves hard to read, we can always change later

website/content/docs/configuration/node-status.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Node Status
+
+{{< hint warning >}}
+**Work in progress**: 
+{{< /hint >}}
+
+
+The node router configuration status can be viewed and monitored via the RouterNodeConfigurationStatus CRD.
+
```
maiqueb, line 18: nit: extra line. is there a reason for it ?

e2etests/tests/node_status.go
maiqueb: let's take @fedepaol 's suggestion. 

The next PR can turn this into something more useful.

maiqueb (changes requested):
some nits still need to be accounted, and we need to simplify the e2e tests.

e2etests/tests/node_status.go
maiqueb: having said that, to me, we should still test that:
- we have one status per node. This means that if something deletes the status, it'll be re-created by the controller
- asserting we have one status per node involves asserting the owner reference of the status CR

ormergi: I would rather have tests follow declarative approach (assert on desired state vs. spesific spec such as onwer ref) but I suppose we can improve later.
Done.

e2etests/tests/node_status.go
ormergi: Done

website/content/docs/configuration/node-status.md
ormergi: Removed

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/1387c5fb4d4f5e4bc2cdbde1fb8b8fc2b1fac069..3a4c18e2615ae12b8a7a0bd7020630ec58a6b556): resolve https://github.com/openperouter/openperouter/pull/355#pullrequestreview-4327464475

maiqueb (approved):
Thanks for this PR. 

I can proceed with the merge once CI finishes (assuming it's green - minus obvious flakes).

maiqueb:
Test    `[FAIL] North/south traffic after FRR container restart [It] should recover north/south traffic after FRR container restart` failed. 

I've never seen this flake before. Can you investigate ? @ormergi

ormergi:
Both failures seem unrelated to this PR changes.
I posted [this issue](https://github.com/openperouter/openperouter/issues/444) to track the `North/south traffic after FRR container restart [It] should recover north/south traffic after FRR container restart` flake with some initial insights.

PR #317, opened 2026-04-16
Implement router resiliency

maiqueb:
**Is this a BUG FIX or a FEATURE ?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug
> /kind cleanup

/kind feature

> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

**What this PR does / why we need it**:

**Special notes for your reviewer**:
Implements enhancement: https://github.com/openperouter/openperouter/blob/main/enhancements/router-resiliency.md

PoC'ed in PR: #271

**Release note**:

```release-note
Make the router resilient to data plane crashes.
```

PR #357, opened 2026-05-04
Router resilience no host networking router

qinqon:

/kind cleanup

Depends-on:
- https://github.com/openperouter/openperouter/pull/317

**Release note**:

```release-note

```


maiqueb:
the 3 e2e test lanes do not even manage to deploy the cluster @qinqon .

Try to understand why, so we can evaluate how much harder will this be. As I've said in #317 - I'd rather have this as a follow-up than waiting for this to be fixed.

qinqon:
> the 3 e2e test lanes do not even manage to deploy the cluster @qinqon .
> 
> Try to understand why, so we can evaluate how much harder will this be. As I've said in #317 - I'd rather have this as a follow-up than waiting for this to be fixed.

@maiqueb this is weird locally the cluster stats, I will investigate.

PR #442, opened 2026-05-20
fix: move underlay interface back to default netns on deletion

### LINKED ISSUE #441: PERouter does not move interface back to host namespace when underlay is deleted
```
[akaris@metallb openperouter ((7e867cb6...))]$ kubectl apply -f /tmp/underlay.yaml 
underlay.openpe.openperouter.github.io/underlay created
[akaris@metallb openperouter ((7e867cb6...))]$ kubectl delete -f /tmp/underlay.yaml 
underlay.openpe.openperouter.github.io "underlay" deleted from openperouter-system namespace
```

```
[akaris@metallb openperouter ((7e867cb6...))]$ cat /tmp/underlay.yaml 
kind: Underlay
metadata:
  name: underlay
spec:
  asn: 64514
  evpn:
    vtepCIDR:  100.65.0.0/24
  nics:
    - toswitch
  neighbors:
    - asn: 64512
      address: 192.168.11.2
```

Interface remains in the perouter namespace:
```
root@pe-kind-control-plane:/# ip netns exec perouter ip a
1: lo: <LOOPBACK> mtu 65536 qdisc noop state DOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
240: toswitch@if241: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9500 qdisc noqueue state UP group default 
    link/ether aa:c1:ab:d8:52:31 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 192.168.11.3/24 brd 192.168.11.255 scope global toswitch
       valid_lft forever preferred_lft forever
    inet6 fe80::a8c1:abff:fed8:5231/64 scope link 
       valid_lft forever preferred_lft forever
```

andreaskaris:
/kind bug

Move underlay interface back to default netns on deletion

Fixes https://github.com/openperouter/openperouter/issues/441

N/A

**Release note**:

```release-note
Fix underlay interface not being moved back to the default network namespace on underlay deletion. The interface is now restored with its original IP addresses and link-up state.
```

### DIFF
--- a/e2etests/pkg/openperouter/netns.go
+++ b/e2etests/pkg/openperouter/netns.go
@@ -10,7 +10,7 @@ import (
 	"github.com/openperouter/openperouter/e2etests/pkg/executor"
 )
 
-const namedNetns = "perouter"
+const NamedNetns = "perouter"
 
 // NamedNetnsExists checks whether /var/run/netns/perouter is present on nodeName.
 func NamedNetnsExists(nodeName string) (bool, error) {
@@ -23,7 +23,7 @@ func NamedNetnsExists(nodeName string) (bool, error) {
 	// Use exact name comparison to avoid "perouter" matching inside "openperouter".
 	for _, line := range strings.Split(out, "\n") {
 		fields := strings.Fields(line)
-		if len(fields) > 0 && fields[0] == namedNetns {
+		if len(fields) > 0 && fields[0] == NamedNetns {
 			return true, nil
 		}
 	}
@@ -34,7 +34,7 @@ func NamedNetnsExists(nodeName string) (bool, error) {
 // interface of the given link type (e.g. "vrf", "bridge", "vxlan").
 func NamedNetnsHasInterfaceType(nodeName, linkType string) (bool, error) {
 	exec := executor.ForContainer(nodeName)
-	out, err := exec.Exec("ip", "netns", "exec", namedNetns, "ip", "link", "show", "type", linkType)
+	out, err := exec.Exec("ip", "netns", "exec", NamedNetns, "ip", "link", "show", "type", linkType)
 	if err != nil {
 		return false, err
 	}
@@ -51,7 +51,7 @@ func DeleteNamedNetns(nodeName string) error {
 		log.Printf("pre-deletion of devices failed for %q, proceeding with netns delete: %v", nodeName, err)
 	}
 
-	_, err := e.Exec("ip", "netns", "delete", namedNetns)
+	_, err := e.Exec("ip", "netns", "delete", NamedNetns)
 	return err
 }
 
@@ -66,11 +66,11 @@ func UnderlayConfigured(nodeName string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	if !strings.Contains(out, namedNetns) {
+	if !strings.Contains(out, NamedNetns) {
 		return false, nil
 	}
 
-	out, err = exec.Exec("ip", "netns", "exec", namedNetns, "ip", "-o", "link", "show", "dev", "lo")
+	out, err = exec.Exec("ip", "netns", "exec", NamedNetns, "ip", "-o", "link", "show", "dev", "lo")
 	if err != nil {
 		return false, err
 	}
@@ -78,7 +78,7 @@ func UnderlayConfigured(nodeName string) (bool, error) {
 		return false, nil
 	}
 
-	out, err = exec.Exec("ip", "netns", "exec", namedNetns, "ip", "address", "show", "dev", "lo", "scope", "global")
+	out, err = exec.Exec("ip", "netns", "exec", NamedNetns, "ip", "address", "show", "dev", "lo", "scope", "global")
 	if err != nil {
 		return false, err
 	}
@@ -97,7 +97,7 @@ func UnderlayVethsExists(nodeName string) bool {
 		if _, err := exec.Exec("ip", "link", "show", iface); err == nil {
 			return true
 		}
-		if _, err := exec.Exec("ip", "netns", "exec", namedNetns, "ip", "link", "show", iface); err == nil {
+		if _, err := exec.Exec("ip", "netns", "exec", NamedNetns, "ip", "link", "show", iface); err == nil {
 			return true
 		}
 	}
@@ -107,7 +107,7 @@ func UnderlayVethsExists(nodeName string) bool {
 // deleteNetnsDevices lists all devices inside the perouter netns and deletes
 // them one by one, skipping loopback.
 func deleteNetnsDevices(e executor.Executor) error {
-	out, err := e.Exec("ip", "netns", "exec", namedNetns, "ip", "-o", "link", "show")
+	out, err := e.Exec("ip", "netns", "exec", NamedNetns, "ip", "-o", "link", "show")
 	if err != nil {
 		return err
 	}
@@ -121,7 +121,7 @@ func deleteNetnsDevices(e executor.Executor) error {
 		if name == "lo" {
 			continue
 		}
-		if out, err := e.Exec("ip", "netns", "exec", namedNetns, "ip", "link", "delete", name); err != nil {
+		if out, err := e.Exec("ip", "netns", "exec", NamedNetns, "ip", "link", "delete", name); err != nil {
 			if strings.Contains(out, "Cannot find device") {
 				continue
 			}
--- a/e2etests/tests/hostconfiguration.go
+++ b/e2etests/tests/hostconfiguration.go
@@ -1184,6 +1184,75 @@ var _ = ginkgo.Describe("Router Host configuration", func() {
 				}, underlayConfiguredTestSelector, p)
 			}
 		})
+
+		ginkgo.It("moves the underlay interface back to the default netns on deletion", func() {
+			nodes, err := k8s.GetNodes(cs)
+			Expect(err).NotTo(HaveOccurred())
+
+			underlayNIC := "toswitch1"
+
+			ginkgo.By(fmt.Sprintf("verifying %s exists in the default netns before creating the underlay", underlayNIC))
+			for _, node := range nodes {
+				Eventually(func() bool {
+					return openperouter.IsInterfaceInDefaultNetns(node.Name, underlayNIC)
+				}).WithTimeout(30 * time.Second).WithPolling(time.Second).Should(
+					BeTrueBecause("%s should be in the default netns on %s before underlay creation", underlayNIC, node.Name))
+			}
+
+			ginkgo.By(fmt.Sprintf("recording the %s IP addresses before creating the underlay", underlayNIC))
+			addrsBefore := map[string]string{}
+			for _, node := range nodes {
+				addrs, err := openperouter.InterfaceIPAddresses(node.Name, underlayNIC)
+				Expect(err).NotTo(HaveOccurred(), "failed to get %s addresses on %s", underlayNIC, node.Name)
+				addrsBefore[node.Name] = addrs
+			}
+
+			ginkgo.By("creating the underlay")
+			err = Updater.Update(config.Resources{
+				Underlays: []v1alpha1.Underlay{underlay},
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By(fmt.Sprintf("verifying %s moved into the perouter netns", underlayNIC))
+			for _, node := range nodes {
+				Eventually(func() bool {
+					return openperouter.IsInterfaceInNS(node.Name, underlayNIC, openperouter.NamedNetns)
+				}).WithTimeout(2 * time.Minute).WithPolling(time.Second).Should(BeTrueBecause(
+					"%s should be inside the perouter netns on %s after underlay creation", underlayNIC, node.Name))
+			}
+
+			ginkgo.By("deleting the underlay")
+			Expect(Updater.CleanAll()).To(Succeed())
+
+			ginkgo.By("waiting for all router pods to be ready after underlay deletion")
+			Eventually(func(g Gomega) {
+				pods, err := openperouter.RouterPods(cs)
+				g.Expect(err).NotTo(HaveOccurred())
+				for _, p := range pods {
+					g.Expect(k8s.PodIsReady(p)).To(BeTrueBecause("pod %s must be ready", p.Name))
+				}
+			}).WithTimeout(2 * time.Minute).WithPolling(time.Second).Should(Succeed())
+
+			ginkgo.By(fmt.Sprintf("verifying %s is back in the default netns after underlay deletion", underlayNIC))
+			for _, node := range nodes {
+				Eventually(func() bool {
+					return openperouter.IsInterfaceInDefaultNetns(node.Name, underlayNIC)
+				}).WithTimeout(2 * time.Minute).WithPolling(time.Second).Should(BeTrueBecause(
+					"%s should be back in the default netns on %s after underlay deletion", underlayNIC, node.Name))
+			}
+
+			ginkgo.By(fmt.Sprintf("verifying %s is UP and has the same IP addresses as before", underlayNIC))
+			for _, node := range nodes {
+				Expect(openperouter.InterfaceIsUp(node.Name, underlayNIC)).To(BeTrueBecause(
+					"%s should be UP on %s after underlay deletion", underlayNIC, node.Name))
+
+				addrsAfter, err := openperouter.InterfaceIPAddresses(node.Name, underlayNIC)
+				Expect(err).NotTo(HaveOccurred(), "failed to get %s addresses on %s after deletion", underlayNIC, node.Name)
+				Expect(addrsAfter).To(Equal(addrsBefore[node.Name]),
+					"%s IP addresses on %s should be preserved after underlay deletion; was: %q, is: %q",
+					underlayNIC, node.Name, addrsBefore[node.Name], addrsAfter)
+			}
+		})
 	})
 
 	ginkgo.Context("L3Passthrough", func() {
--- a/internal/hostnetwork/link_properties.go
+++ b/internal/hostnetwork/link_properties.go
@@ -13,7 +13,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/openperouter/openperouter/internal/netnamespace"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netlink/nl"
 	"github.com/vishvananda/netns"
@@ -173,67 +172,70 @@ func setNeighSuppression(link netlink.Link) error {
 	return nil
 }
 
-// moveInterfaceToNamespace takes the given interface and moves it into the given namespace.
-func moveInterfaceToNamespace(ctx context.Context, intf string, ns netns.NsHandle) error {
-	slog.DebugContext(ctx, "move intf to namespace", "intf", intf, "namespace", ns.String())
-	defer slog.DebugContext(ctx, "move intf to namespace end", "intf", intf, "namespace", ns.String())
+// moveInterfaceToNamespace takes the given interface and moves it from the given to the given namespace.
+func moveInterfaceToNamespace(ctx context.Context, intf string, fromHandle, toHandle *netlink.Handle,
+	toNS netns.NsHandle, groupID uint32) error {
+	slog.DebugContext(ctx, "move intf to namespace", "intf", intf, "toNS", toNS.String())
+	defer slog.DebugContext(ctx, "move intf to namespace end", "intf", intf, "toNS", toNS.String())
 
-	err := netnamespace.In(ns, func() error {
-		_, err := netlink.LinkByName(intf)
-		if err != nil {
-			return fmt.Errorf("moveInterfaceToNamespace: Failed to find link %s: %w", intf, err)
-		}
-		return nil
-	})
-	if err == nil {
-		slog.DebugContext(ctx, "intf is already in namespace", "intf", intf, "namespace", ns.String())
+	// Nothing to do for us if the interface is already inside the target namespace.
+	if _, err := toHandle.LinkByName(intf); err == nil {
+		slog.DebugContext(ctx, "intf is already in namespace", "intf", intf, "namespace", toNS.String())
 		return nil
 	}
-	var nsError netnamespace.SetNamespaceError
-	if errors.As(err, &nsError) {
-		return fmt.Errorf("moveInterfaceToNamespace: Failed to execute in ns %s: %w", intf, err)
-	}
 
-	link, err := netlink.LinkByName(intf)
+	// Get the link, store its current addresses so that we can later restore them (addresses are lost upon moving),
+	// and move the link to the target netns.
+	link, err := fromHandle.LinkByName(intf)
 	if err != nil {
 		return fmt.Errorf("moveInterfaceToNamespace: Failed to find link %s: %w", intf, err)
 	}
 
-	addresses, err := netlink.AddrList(link, netlink.FAMILY_ALL)
+	addresses, err := fromHandle.AddrList(link, netlink.FAMILY_ALL)
 	if err != nil {
 		return fmt.Errorf("moveInterfaceToNamespace: Failed to get addresses for intf %s: %w", link.Attrs().Name, err)
 	}
-
 	slog.DebugContext(ctx, "addresses before moving", "addresses", addresses)
-	err = netlink.LinkSetNsFd(link, int(ns))
-	if err != nil {
-		return fmt.Errorf("moveInterfaceToNamespace: Failed to move %s to network namespace %s: %w", link.Attrs().Name, ns.String(), err)
-	}
-	if err := netnamespace.In(ns, func() error {
-		nsLink, err := netlink.LinkByName(intf)
-		if err != nil {
-			return fmt.Errorf("moveInterfaceToNamespace: Failed to get link by name %s up in network namespace %s: %w", intf, ns.String(), err)
-		}
-		err = netlink.LinkSetUp(nsLink)
-		if err != nil {
-			return fmt.Errorf("moveInterfaceToNamespace: Failed to set %s up in network namespace %s: %w", nsLink.Attrs().Name, ns.String(), err)
-		}
 
-		slog.DebugContext(ctx, "restoring addresses in namespace", "addresses", addresses)
-		for _, a := range addresses {
-			slog.DebugContext(ctx, "restoring address in namespace", "address", a, "flags", a.Flags)
-			IFA_F_NOPREFIXROUTE := 0x200 //nolint:revive // matches kernel define
-			a.Flags &= ^IFA_F_NOPREFIXROUTE
-			slog.DebugContext(ctx, "restoring address in namespace after no prefix", "address", a, "flags", a.Flags)
-			err := netlink.AddrAdd(nsLink, &a)
-			if err != nil && !os.IsExist(err) {
-				return fmt.Errorf("moveNicToNamespace: Failed to add address %s to %s: %w", a, nsLink.Attrs().Name, err)
-			}
-		}
-		return nil
-	}); err != nil {
-		return err
+	err = fromHandle.LinkSetNsFd(link, int(toNS))
+	if err != nil {
+		return fmt.Errorf("moveInterfaceToNamespace: Failed to move %s to network namespace %s: %w", link.Attrs().Name, toNS.String(), err)
 	}
+
+	// After the move, get the link inside the new namespace, set the link up, and add the addresses again.
+	nsLink, err := toHandle.LinkByName(intf)
+	if err != nil {
+		return fmt.Errorf("moveInterfaceToNamespace: Failed to get link by name %s up in network namespace %s: %w", intf, toNS.String(), err)
+	}
+
+	slog.DebugContext(ctx, "restoring addresses in namespace", "addresses", addresses)
+	var errs []error
+	for _, a := range addresses {
+		slog.DebugContext(ctx, "restoring address in namespace", "address", a, "flags", a.Flags)
+		a.Flags &= ^unix.IFA_F_NOPREFIXROUTE
+		slog.DebugContext(ctx, "restoring address in namespace after no prefix", "address", a, "flags", a.Flags)
+		err := toHandle.AddrAdd(nsLink, &a)
+		if err != nil && !os.IsExist(err) {
+			errs = append(errs, fmt.Errorf("moveInterfaceToNamespace: Failed to add address %s to %s: %w", a, nsLink.Attrs().Name, err))
+			continue
+		}
+	}
+	if len(errs) != 0 {
+		return errors.Join(errs...)
+	}
+
+	// Set group ID only if not already set (idempotent)
+	if nsLink.Attrs().Group != groupID {
+		if err := toHandle.LinkSetGroup(nsLink, int(groupID)); err != nil {
+			return fmt.Errorf("moveInterfaceToNamespace: Failed to set group ID %d on interface %s: %w", groupID, intf, err)
+		}
+	}
+
+	err = toHandle.LinkSetUp(nsLink)
+	if err != nil {
+		return fmt.Errorf("moveInterfaceToNamespace: Failed to set %s up in network namespace %s: %w", nsLink.Attrs().Name, toNS.String(), err)
+	}
+
 	return nil
 }
 
--- a/internal/hostnetwork/underlay.go
+++ b/internal/hostnetwork/underlay.go
@@ -37,20 +37,30 @@ type UnderlayTunnelEndpointParams struct {
 func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
 	slog.DebugContext(ctx, "setup underlay", "params", params)
 	defer slog.DebugContext(ctx, "setup underlay done")
-	ns, err := netns.GetFromPath(params.TargetNS)
+
+	targetNetNS, err := netns.GetFromPath(params.TargetNS)
 	if err != nil {
 		return fmt.Errorf("setupUnderlay: Failed to find network namespace %s: %w", params.TargetNS, err)
 	}
 	defer func() {
-		if err := ns.Close(); err != nil {
+		if err := targetNetNS.Close(); err != nil {
 			slog.Error("failed to close namespace", "namespace", params.TargetNS, "error", err)
+		}
+	}()
+	defaultNetNS, err := netns.Get()
+	if err != nil {
+		return fmt.Errorf("SetupUnderlay: failed to get netns handle for default namespace: %w", err)
+	}
+	defer func() {
+		if err := defaultNetNS.Close(); err != nil {
+			slog.Error("failed to close default namespace", "error", err)
 		}
 	}()
 
 	// Check if there are existing underlay interfaces that aren't in the new list.
 	// This means the underlay configuration changed and requires rebuilding
 	// the network namespace.
-	existingIfaces, err := findInterfacesInGroup(ns, underlayGroupID)
+	existingIfaces, err := findInterfacesInGroup(targetNetNS, underlayGroupID)
 	if err != nil {
 		return fmt.Errorf("failed to check existing underlay interfaces: %w", err)
 	}
@@ -61,26 +71,20 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
 		}
 	}
 
+	targetNetNSHandle, err := netlink.NewHandleAt(targetNetNS)
+	if err != nil {
+		return fmt.Errorf("SetupUnderlay: failed to get netlink handle for namespace %s: %w", targetNetNS.String(), err)
+	}
+	defer targetNetNSHandle.Close()
+	defaultNetNSHandle, err := netlink.NewHandleAt(defaultNetNS)
+	if err != nil {
+		return fmt.Errorf("SetupUnderlay: failed to get netlink handle for default namespace: %w", err)
+	}
+	defer defaultNetNSHandle.Close()
+
 	for _, underlayInterface := range params.UnderlayInterfaces {
-		if err := moveInterfaceToNamespace(ctx, underlayInterface, ns); err != nil {
-			return err
-		}
-		if err := netnamespace.In(ns, func() error {
-			underlay, err := netlink.LinkByName(underlayInterface)
-			if err != nil {
-				return fmt.Errorf("failed to get underlay nic by name %s: %w", underlayInterface, err)
-			}
-			// Set group ID only if not already set (idempotent)
-			if underlay.Attrs().Group != underlayGroupID {
-				if err := netlink.LinkSetGroup(underlay, int(underlayGroupID)); err != nil {
-					return fmt.Errorf("failed to set group ID on underlay interface %s: %w", underlayInterface, err)
-				}
-			}
-			if err := netlink.LinkSetUp(underlay); err != nil {
-				return fmt.Errorf("could not set link up for VRF %s: %v", underlay.Attrs().Name, err)
-			}
-			return nil
-		}); err != nil {
+		if err := moveInterfaceToNamespace(ctx, underlayInterface, defaultNetNSHandle, targetNetNSHandle, targetNetNS,
+			underlayGroupID); err != nil {
 			return err
 		}
 	}
@@ -96,7 +100,7 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
 	if ip := params.TunnelEndpoint.IPv6CIDR; ip != "" {
 		vtepIPs = append(vtepIPs, ip)
 	}
-	if err := ensureLoopback(ctx, ns, vtepIPs...); err != nil {
+	if err := ensureLoopback(ctx, targetNetNS, vtepIPs...); err != nil {
 		return err
 	}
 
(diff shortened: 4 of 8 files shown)


internal/hostnetwork/link_properties.go
```diff
@@ -43,11 +41,16 @@ func assignIPToInterface(link netlink.Link, address string) error {
...
+func interfaceHasIPInNetns(handle netlink.Handle, link netlink.Link, address string) (bool, error) {
 	_, err := netlink.ParseAddr(address)
 	if err != nil {
 		return false, fmt.Errorf("interfaceHasIP: failed to parse address %s for interface %s: %w", address, link.Attrs().Name, err)
 	}
-	addresses, err := netlink.AddrList(link, netlink.FAMILY_ALL)
+	addresses, err := handle.AddrList(link, netlink.FAMILY_ALL)
 	if err != nil {
 		return false, fmt.Errorf("interfaceHasIP: failed to list addresses for interface %s: %w", link.Attrs().Name, err)
 	}
```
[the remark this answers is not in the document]
andreaskaris: done

internal/hostnetwork/underlay.go
```diff
@@ -155,38 +165,62 @@ func RemoveUnderlay(targetNS string) error {
...
+		// We can safely do this here as RemoveAllVNIs must have deleted all VNI configs before.
+		name := l.Attrs().Name
+		if name == "lo" || name == UnderlayLoopback {
+			continue
+		}
+		err = moveInterfaceToNamespace(context.TODO(), name, ns, defaultNS)
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
andreaskaris: done

internal/hostnetwork/link_properties.go
```diff
@@ -173,66 +183,67 @@ func setNeighSuppression(link netlink.Link) error {
...
-			slog.DebugContext(ctx, "restoring address in namespace after no prefix", "address", a, "flags", a.Flags)
-			err := netlink.AddrAdd(nsLink, &a)
-			if err != nil && !os.IsExist(err) {
-				return fmt.Errorf("moveNicToNamespace: Failed to add address %s to %s: %w", a, nsLink.Attrs().Name, err)
-			}
+	slog.DebugContext(ctx, "restoring addresses in namespace", "addresses", addresses)
+	for _, a := range addresses {
+		slog.DebugContext(ctx, "restoring address in namespace", "address", a, "flags", a.Flags)
+		IFA_F_NOPREFIXROUTE := 0x200 //nolint:revive // matches kernel define
+		a.Flags &= ^IFA_F_NOPREFIXROUTE
```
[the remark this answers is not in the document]
andreaskaris: I'm not touching this, it was like this before

internal/hostnetwork/underlay.go
```diff
@@ -155,38 +166,60 @@ func RemoveUnderlay(targetNS string) error {
...
+			return fmt.Errorf("failed to remove underlay marker from %s: %w", l.Attrs().Name, err)
 		}
-		return nil
-	})
+
+		if err = moveInterfaceToNamespace(ctx, l.Attrs().Name, ns, defaultNS); err != nil {
+			return err
+		}
+		break
+	}
```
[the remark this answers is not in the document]
andreaskaris: It was like that before, it's really not a lot of efficiency gains for the few links, and I want interfaceHasIP and interfaceHasIPInNetns to behave the same way (so that it won't confuse someone using them) and interfaceHasIP is called in several other locations

e2etests/pkg/openperouter/netns.go
```diff
@@ -67,3 +69,56 @@ func UnderlayVethExists(nodeName string) bool {
 	}
 	return false
 }
+
+// UnderlayVethInDefaultNetns checks whether the toswitch interface exists
+// in the default network namespace on nodeName.
+func UnderlayVethInDefaultNetns(nodeName string) bool {
```
maiqueb: let's prefix this predicate with `Is...` so we know this returns a boolean without having to check it's signature.

e2etests/pkg/openperouter/netns.go
```diff
@@ -67,3 +69,56 @@ func UnderlayVethExists(nodeName string) bool {
...
+	exec := executor.ForContainer(nodeName)
+	_, err := exec.Exec("ip", "link", "show", "toswitch")
+	return err == nil
+}
+
+// UnderlayVethInPerouterNetns checks whether the toswitch interface exists
+// inside the perouter network namespace on nodeName.
+func UnderlayVethInPerouterNetns(nodeName string) bool {
```
maiqueb: ditto

e2etests/pkg/openperouter/netns.go
```diff
@@ -67,3 +69,56 @@ func UnderlayVethExists(nodeName string) bool {
...
+	var addrs []string
+	for _, line := range strings.Split(out, "\n") {
+		m := addrRegexp.FindStringSubmatch(line)
+		if len(m) < 2 {
+			continue
+		}
+		addr := m[1]
+		if strings.HasPrefix(addr, "inet6 fe80:") || strings.HasPrefix(addr, "inet 169.254.") {
+			continue
+		}
```
maiqueb: hm, can't we rely on the address scope instead ? handling this by the known prefixes is cumbersome. 

I think we can `ip addr show scope global` for it.

e2etests/pkg/openperouter/netns.go
```diff
@@ -67,3 +69,56 @@ func UnderlayVethExists(nodeName string) bool {
...
+
+// UnderlayVethIsUp checks whether the toswitch interface in the default netns
+// on nodeName has state UP.
+func UnderlayVethIsUp(nodeName string) bool {
+	exec := executor.ForContainer(nodeName)
+	out, err := exec.Exec("ip", "link", "show", "toswitch")
+	if err != nil {
+		return false
+	}
+	return strings.Contains(out, "state UP")
```
maiqueb: you can query directly for `ip link show toswitch up` . no need for `strings.Contains`

e2etests/tests/hostconfiguration.go
```diff
@@ -1190,6 +1190,73 @@ var _ = ginkgo.Describe("Router Host configuration", func() {
...
+		ginkgo.It("moves the underlay interface back to the default netns on deletion", func() {
+			nodes, err := k8s.GetNodes(cs)
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By("verifying toswitch exists in the default netns before creating the underlay")
+			for _, node := range nodes {
+				Eventually(func() bool {
+					return openperouter.UnderlayVethInDefaultNetns(node.Name)
+				}).WithTimeout(30*time.Second).WithPolling(time.Second).Should(BeTrue(),
+					"toswitch should be in the default netns on %s before underlay creation", node.Name)
```
maiqueb: nit: maybe you could explore using `BeTrueBecause`: https://pkg.go.dev/github.com/onsi/gomega#BeTrueBecause

A bit more explicit IMHO. 

100 % opinionated.

internal/hostnetwork/link_properties.go
```diff
@@ -43,14 +41,26 @@ func assignIPToInterface(link netlink.Link, address string) error {
 
 // interfaceHasIP tells if the given link has the provided ip.
 func interfaceHasIP(link netlink.Link, address string) (bool, error) {
+	return interfaceHasIPInNetns(nil, link, address)
```
maiqueb: not extremely fond of this ... but I can compromise

internal/hostnetwork/underlay.go
```diff
@@ -155,38 +166,60 @@ func RemoveUnderlay(targetNS string) error {
...
+		return fmt.Errorf("RemoveUnderlay: failed to get netlink handle for namespace %s: %w", targetNS, err)
+	}
+	defer handle.Close()
+
+	lound, err := handle.LinkByName(UnderlayLoopback)
+	if err != nil {
+		if !errors.As(err, &netlink.LinkNotFoundError{}) {
+			return fmt.Errorf("failed to find %s: %w", UnderlayLoopback, err)
+		}
+	} else if err := handle.LinkDel(lound); err != nil {
```
maiqueb: we don't need the else, do we ?

internal/hostnetwork/underlay.go
```diff
@@ -155,38 +166,60 @@ func RemoveUnderlay(targetNS string) error {
...
-			}
+		if err := handle.AddrDel(l, addr); err != nil {
+			return fmt.Errorf("failed to remove underlay marker from %s: %w", l.Attrs().Name, err)
 		}
-		return nil
-	})
+
+		if err = moveInterfaceToNamespace(ctx, l.Attrs().Name, ns, defaultNS); err != nil {
+			return err
+		}
```
maiqueb: what if we fail to move the interface back to the ns, but succeeded at removing the IP from it ? 

we'll never again be able to reconcile it, right ? 

what happens if we switch this around (i.e. first move the iface, only then remove the marker ? )

internal/hostnetwork/underlay_test.go
```diff
@@ -170,6 +171,72 @@ var _ = Describe("Underlay configuration should work when", func() {
...
+	It("RemoveUnderlay should move the underlay interface back to the default namespace", func() {
+		params := UnderlayParams{
+			UnderlayInterface: underlayTestInterface,
+			EVPN: &UnderlayEVPNParams{
+				VtepIP: "192.168.1.1/32",
+			},
+			TargetNS: underlayTestNSPath(),
+		}
+		err := SetupUnderlay(context.Background(), params)
+		Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 170: please inline, which enables you to use positive logic - i.e. `Expect To Succeed`

internal/hostnetwork/underlay_test.go
```diff
@@ -170,6 +171,72 @@ var _ = Describe("Underlay configuration should work when", func() {
...
+				hasIP, err = interfaceHasIP(link, underlayInterfaceSpecialAddr)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(hasIP).To(BeTrue(), "interface should have %s while in the target namespace", underlayInterfaceSpecialAddr)
+
+				return nil
+			})
+		}, 30*time.Second, 1*time.Second).Should(Succeed())
+
+		err = RemoveUnderlay(context.TODO(), underlayTestNSPath())
+		Expect(err).NotTo(HaveOccurred())
```
maiqueb: ditto

internal/hostnetwork/underlay_test.go
```diff
@@ -170,6 +171,72 @@ var _ = Describe("Underlay configuration should work when", func() {
...
+
+		By("verifying the loopback was deleted from the target namespace")
+		Eventually(func(g Gomega) {
+			_ = netnamespace.In(testNs, func() error {
+				_, err := netlink.LinkByName(UnderlayLoopback)
+				g.Expect(err).To(HaveOccurred(), "loopback should be deleted from the target namespace")
```
maiqueb: should we assert the error thrown ? i.e. `MatchError(LinkNotFoundError)` 
https://github.com/vishvananda/netlink/blob/99e979749d3eaa40d3e329beb91544abf518a9e8/link.go#L1486

maiqueb (changes requested):
A few suggestions / questions, but overall, I think this is needed - now that we no longer destroy the pod on underlay removal, the interface should be moved back to the node's net namespace.

e2etests/pkg/openperouter/netns.go
andreaskaris: done

e2etests/pkg/openperouter/netns.go
andreaskaris: done

e2etests/pkg/openperouter/netns.go
andreaskaris: cool I didn't know this and neither did Claude

e2etests/tests/hostconfiguration.go
andreaskaris: done

e2etests/pkg/openperouter/netns.go
andreaskaris: done

internal/hostnetwork/underlay.go
andreaskaris: don't think so

internal/hostnetwork/underlay.go
andreaskaris: well given that IP addresses must be explicitly restored, then let's just do:
```
// moveInterfaceToNamespace takes the given interface and moves it from the given to the given namespace.
// Addresses in slice excludeAddresses will be dropped duringthe move.
func moveInterfaceToNamespace(ctx context.Context, intf string, fromNS, toNS netns.NsHandle, excludeAddresses []netlink.Addr) error {
```

and instruct the move function not to move that address? see next push

internal/hostnetwork/link_properties.go
andreaskaris: let's compromise ;)

internal/hostnetwork/underlay_test.go
andreaskaris: I left it as is because I got a nil pointer exception with MatchError.

Claude suggested this here:
```
			var linkNotFoundErr *netlink.LinkNotFoundError
			g.Expect(errors.As(err, &linkNotFoundErr)).To(BeTrue(), "loopback should be deleted from the target namespace")
```
But I find errors.As difficult to understand .. maybe it's just me, lmk and I'll change it

internal/hostnetwork/link_properties.go
maiqueb: deal.

internal/hostnetwork/underlay_test.go
maiqueb: that's ugly, let's not. 

What NPE do you get ? match error is supposed to fail is the expected error is nil ...

maiqueb: FWIW, I was expecting something like:

```go
Expect(err).To(MatchError(&netlink.LinkNotFoundError{}))
```

andreaskaris: I think it has more to do with a missing value while printing. the error isn't nil as I'm keeping the nil check before and then just add:

```
				Expect(err).To(MatchError(&netlink.LinkNotFoundError{}))
```

```
Underlay configuration should work when [It] RemoveUnderlay should move the underlay interface back to the default namespace
/home/akaris/development/openperouter/fix-namespace-issues/internal/hostnetwork/underlay_test.go:175

  [PANICKED] Test Panicked
  In [It] at: /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/internal/async_assertion.go:333 @ 05/21/26 15:45:45.7

  runtime error: invalid memory address or nil pointer dereference

  Full Stack Trace
    github.com/onsi/gomega/internal.(*AsyncAssertion).buildActualPoller.func3.1()
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/internal/async_assertion.go:333 +0x31f
    panic({0x15b4e80?, 0x2014c00?})
        /home/akaris/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.3.linux-amd64/src/runtime/panic.go:860 +0x13a
    github.com/vishvananda/netlink.(*LinkNotFoundError).Error(0xc0002c8670)
        <autogenerated>:1 +0x35
    github.com/onsi/gomega/format.Object({0x15a58a0, 0xc0002c8670}, 0x1)
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/format/format.go:263 +0x158
    github.com/onsi/gomega/format.Message({0x160d5a0, 0xc0002c8630}, {0x1707def, 0xe}, {0xc000586e18, 0x1, 0xc0001fcbf8?})
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/format/format.go:132 +0x19e
    github.com/onsi/gomega/matchers.(*MatchErrorMatcher).FailureMessage(0xc0001fcbd0, {0x160d5a0, 0xc0002c8630})
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/matchers/match_error_matcher.go:78 +0x188
    github.com/onsi/gomega/internal.(*Assertion).match(0xc0004ce300, {0x1790530, 0xc0001fcbd0}, 0x1, {0x0, 0x0, 0x0})
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/internal/assertion.go:101 +0x1a8
    github.com/onsi/gomega/internal.(*Assertion).To(0xc0004ce300, {0x1790530, 0xc0001fcbd0}, {0x0, 0x0, 0x0})
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/internal/assertion.go:62 +0xfb
    github.com/openperouter/openperouter/internal/hostnetwork.init.func3.9.3.1()
        /home/akaris/development/openperouter/fix-namespace-issues/internal/hostnetwork/underlay_test.go:218 +0x263
    github.com/openperouter/openperouter/internal/netnamespace.In(0x3b, 0xc000587130)
        /home/akaris/development/openperouter/fix-namespace-issues/internal/netnamespace/netnamespace.go:46 +0x216
    github.com/openperouter/openperouter/internal/hostnetwork.init.func3.9.3({0x179e5c0, 0xc0004ce180})
        /home/akaris/development/openperouter/fix-namespace-issues/internal/hostnetwork/underlay_test.go:215 +0x8e
    reflect.Value.call({0x156f8a0?, 0xc0002c8550?, 0x45c6c5?}, {0x16f4715, 0x4}, {0xc0004d0180, 0x1, 0x139c125?})
        /home/akaris/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.3.linux-amd64/src/reflect/value.go:586 +0x1232
    reflect.Value.Call({0x156f8a0?, 0xc0002c8550?, 0x1?}, {0xc0004d0180, 0x1, 0x1})
        /home/akaris/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.3.linux-amd64/src/reflect/value.go:369 +0xb6
    github.com/onsi/gomega/internal.(*AsyncAssertion).buildActualPoller.func3()
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/internal/async_assertion.go:337 +0x23b
    github.com/onsi/gomega/internal.(*AsyncAssertion).match(0xc0002b8150, {0x17903e0, 0x2075240}, 0x1, {0x0, 0x0, 0x0})
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/internal/async_assertion.go:410 +0x299
    github.com/onsi/gomega/internal.(*AsyncAssertion).Should(0xc0002b8150, {0x17903e0, 0x2075240}, {0x0, 0x0, 0x0})
        /home/akaris/go/pkg/mod/github.com/onsi/gomega@v1.39.1/internal/async_assertion.go:145 +0xc5
    github.com/openperouter/openperouter/internal/hostnetwork.init.func3.9()
        /home/akaris/development/openperouter/fix-namespace-issues/internal/hostnetwork/underlay_test.go:221 +0x7f6
```

maiqueb: ah ... got it. `netlink.LinkNotFoundError` has an embedded private error. This is - afaiu - a no go.

This would work though:
```
		Eventually(func() error {
			return netnamespace.In(testNs, func() error {
				_, err := netlink.LinkByName(UnderlayLoopback)
				return err
			})
		}, 30*time.Second, 1*time.Second).Should(MatchError("Link not found"))
```

It's not as expressive, but I think it's better than what you have now. 
... which is preferable to what claude suggested.

maiqueb: > doesn't work though, the field is unexported

Yes, that's what I meant with "embedded private error". Java terminology (yikes!)

OK, I get it ... it's used elsewhere. 

... but real ugly. OK for me. As long as we match the error is the expected one and not just an error, I'm OK.

andreaskaris: done

andreaskaris:
@fedepaol PTAL

qinqon:
@andreaskaris can you rebase the PR ? now the underlay is filtered by group id instead of special address

internal/hostnetwork/underlay.go
```diff
@@ -147,29 +158,58 @@ func RemoveUnderlay(targetNS string) error {
...
+		return fmt.Errorf("failed to list links in default namespace: %w", err)
+	}
+	for _, l := range links {
+		if l.Attrs().Group != underlayGroupID {
+			continue
+		}
+		if err := netlink.LinkSetGroup(l, 0); err != nil {
+			return fmt.Errorf("failed to reset group ID on %s: %w", l.Attrs().Name, err)
+		}
+	}
```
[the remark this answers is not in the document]
andreaskaris: yes, but if moveInterfaceToNamespace fails somewhere, and we have to retry the operation, we will not be able to identify the interfaces any more and they will be stuck in the namespace

andreaskaris:
@qinqon @maiqueb ready for review

internal/hostnetwork/link_properties.go
```diff
@@ -43,14 +41,26 @@ func assignIPToInterface(link netlink.Link, address string) error {
...
+}
+
+// interfaceHasIPInNetns tells if the given link has the provided ip, inside the provided netns handle.
+func interfaceHasIPInNetns(handle *netlink.Handle, link netlink.Link, address string) (bool, error) {
```
qinqon: I only see one call to interfaceHasIPInNetns and it's passing "nil" in the handle, why do we introduce the handle concept ?

internal/hostnetwork/link_properties.go
```diff
@@ -43,14 +41,26 @@ func assignIPToInterface(link netlink.Link, address string) error {
...
 		return false, fmt.Errorf("interfaceHasIP: failed to parse address %s for interface %s: %w", address, link.Attrs().Name, err)
 	}
-	addresses, err := netlink.AddrList(link, netlink.FAMILY_ALL)
+
+	var addresses []netlink.Addr
+	if handle == nil {
+		addresses, err = netlink.AddrList(link, netlink.FAMILY_ALL)
+	} else {
+		addresses, err = handle.AddrList(link, netlink.FAMILY_ALL)
+	}
```
qinqon: We should not duplicate the call we should pass a defualt handle implementation

internal/hostnetwork/link_properties.go
```diff
@@ -173,66 +183,67 @@ func setNeighSuppression(link netlink.Link) error {
...
-		_, err := netlink.LinkByName(intf)
-		if err != nil {
-			return fmt.Errorf("moveInterfaceToNamespace: Failed to find link %s: %w", intf, err)
-		}
-		return nil
-	})
-	if err == nil {
-		slog.DebugContext(ctx, "intf is already in namespace", "intf", intf, "namespace", ns.String())
-		return nil
+	fromHandle, err := netlink.NewHandleAt(fromNS)
```
qinqon: Why do we use Handle concept instead of namespace.In ? using this introduce complexity, and we have to lock the thread <- that's already implemented at the netnamespace.In

This should be implemented as 
```golang
runtime.LockOSThread()
defer runtime.UnlockOSThread()

curNS, err := netns.Get()
if err != nil {
    return fmt.Errorf("failed to get current namespace: %w", err)
}
defer curNS.Close()

fromHandle, err := netlink.NewHandleAtFrom(fromNS, curNS)
if err != nil { ... }
defer fromHandle.Close()

toHandle, err := netlink.NewHandleAtFrom(toNS, curNS)
if err != nil { ... }
defer toHandle.Close()
```

internal/hostnetwork/underlay.go
```diff
@@ -147,29 +162,58 @@ func RemoveUnderlay(targetNS string) error {
...
 
-	return netnamespace.In(ns, func() error {
-		lound, err := netlink.LinkByName(UnderlayLoopback)
-		if err != nil {
-			if !errors.As(err, &netlink.LinkNotFoundError{}) {
-				return fmt.Errorf("failed to find %s: %w", UnderlayLoopback, err)
-			}
-		} else if err := netlink.LinkDel(lound); err != nil {
-			return fmt.Errorf("failed to delete %s: %w", UnderlayLoopback, err)
+	handle, err := netlink.NewHandleAt(ns)
```
qinqon: ditto we need the LockOSThread thingy.

qinqon (changes requested):
I think we need to block the thread before creating handles from the namespace and use them.

internal/hostnetwork/link_properties.go
andreaskaris: I had that first, but claude pushed me towards this >_>
anyway, that entire function makes no sense

internal/hostnetwork/link_properties.go
andreaskaris: yeah makes no sense - it's a vestige from before the rebase, but even then I should have executed it simply with namespace.In

internal/hostnetwork/link_properties.go
andreaskaris: Looking at the netlink library's NewHandleAt, it uses netns.None() ("None gets an empty (closed) NsHandle.")

```
// NewHandleAt returns a netlink handle on the network namespace
// specified by ns. If ns=netns.None(), current network namespace
// will be assumed
func NewHandleAt(ns netns.NsHandle, nlFamilies ...int) (*Handle, error) {
	return newHandle(ns, netns.None(), nlFamilies...)
}

// NewHandleAtFrom works as NewHandle but allows client to specify the
// new and the origin netns Handle.
func NewHandleAtFrom(newNs, curNs netns.NsHandle) (*Handle, error) {
	return newHandle(newNs, curNs)
}

func newHandle(newNs, curNs netns.NsHandle, nlFamilies ...int) (*Handle, error) {
	h := &Handle{sockets: map[int]*nl.SocketHandle{}}
	fams := nl.SupportedNlFamilies
	if len(nlFamilies) != 0 {
		fams = nlFamilies
	}
	for _, f := range fams {
		s, err := nl.GetNetlinkSocketAt(newNs, curNs, f)
```

andreaskaris: The library at some point does lock for us, and if netns.None() is set for current, it'll actually get the new ns inside the lock. So your suggestion isn't needed, or am I reading that wrong?

```
// GetNetlinkSocketAt opens a netlink socket in the network namespace newNs
// and positions the thread back into the network namespace specified by curNs,
// when done. If curNs is close, the function derives the current namespace and
// moves back into it when done. If newNs is close, the socket will be opened
// in the current network namespace.
func GetNetlinkSocketAt(newNs, curNs netns.NsHandle, protocol int) (*NetlinkSocket, error) {
	c, err := executeInNetns(newNs, curNs)
	if err != nil {
		return nil, err
	}
	defer c()
	return getNetlinkSocket(protocol)
}

func executeInNetns(newNs, curNs netns.NsHandle) (func(), error) {
	var (
		err       error
		moveBack  func(netns.NsHandle) error
		closeNs   func() error
		unlockThd func()
	)
	restore := func() {
		// order matters
		if moveBack != nil {
			moveBack(curNs)
		}
		if closeNs != nil {
			closeNs()
		}
		if unlockThd != nil {
			unlockThd()
		}
	}
	if newNs.IsOpen() {
		runtime.LockOSThread()
		unlockThd = runtime.UnlockOSThread
		if !curNs.IsOpen() {
			if curNs, err = netns.Get(); err != nil {
				restore()
				return nil, fmt.Errorf("could not get current namespace while creating netlink socket: %v", err)
			}
			closeNs = curNs.Close
		}
```

qinqon: After checking the NewHandleAt I see that the  runtime.LockOSThread is already done there, so you are right you can resolve this.

internal/hostnetwork/underlay.go
qinqon: nonsense we can solve this.

internal/hostnetwork/link_properties.go
andreaskaris: I find working with the handles less complex in this context than doing netnamespace.In with a context, due to the back and forth switches between the namespaces

internal/hostnetwork/link_properties.go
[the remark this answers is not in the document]
andreaskaris: done

internal/hostnetwork/underlay.go
andreaskaris: same as above

andreaskaris: thanks, though, for pointing it out. I just blindly assumed stuff without checking!

internal/hostnetwork/underlay.go
```diff
@@ -134,35 +138,56 @@ func ensureLoopback(ctx context.Context, ns netns.NsHandle, vtepIP string) error
...
+			slog.Error("failed to close namespace", "namespace", fromNetNSPath, "error", err)
 		}
 	}()
-
-	return netnamespace.In(ns, func() error {
-		if err := clearNonDefaultLoopbackIPs(loopbackName); err != nil {
-			return err
+	defaultNetNS, err := netns.Get()
+	if err != nil {
+		return fmt.Errorf("RemoveUnderlay: failed to get netlink handle for default namespace: %w", err)
```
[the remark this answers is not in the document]
andreaskaris: done

internal/hostnetwork/underlay.go
```diff
@@ -60,26 +70,20 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
-			if err := netlink.LinkSetUp(underlay); err != nil {
-				return fmt.Errorf("could not set link up for VRF %s: %v", underlay.Attrs().Name, err)
-			}
-			return nil
-		}); err != nil {
+		if err := moveInterfaceToNamespace(ctx, underlayInterface, defaultNetNSHandle, targetNetNSHandle, targetNetNS,
+			underlayGroupID); err != nil {
 			return err
 		}
 	}
```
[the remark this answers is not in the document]
andreaskaris: hallucination?

internal/hostnetwork/underlay.go
```diff
@@ -134,35 +138,56 @@ func ensureLoopback(ctx context.Context, ns netns.NsHandle, vtepIP string) error
...
 		}
-		for _, l := range links {
-			if l.Attrs().Group == underlayGroupID {
-				if err := netlink.LinkSetGroup(l, 0); err != nil {
-					return fmt.Errorf("failed to reset group ID on %s: %w", l.Attrs().Name, err)
-				}
-			}
+		if err = moveInterfaceToNamespace(ctx, l.Attrs().Name, fromNetNSHandle, defaultNetNSHandle, defaultNetNS, 0); err != nil {
+			return err
 		}
```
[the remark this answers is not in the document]
andreaskaris: already done by the function itself

internal/hostnetwork/link_properties.go
```diff
@@ -173,67 +171,65 @@ func setNeighSuppression(link netlink.Link) error {
...
-		}
-		return nil
-	})
-	if err == nil {
-		slog.DebugContext(ctx, "intf is already in namespace", "intf", intf, "namespace", ns.String())
+	// Nothing to do for us if the interface is already inside the target namespace.
+	if _, err := toHandle.LinkByName(intf); err == nil {
+		slog.DebugContext(ctx, "intf is already in namespace", "intf", intf, "namespace", toNS.String())
 		return nil
 	}
```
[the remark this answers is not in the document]
andreaskaris: I absolutely agree with this, but it's out of scope for this PR - the problem is that it's likely more complex than that, as IP addresses might also be missing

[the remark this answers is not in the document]
andreaskaris: Given that this potential issue wasn't introduced here, deferring to issue: https://github.com/openperouter/openperouter/issues/470

internal/hostnetwork/link_properties.go
maiqueb: We should agree in a particular format and try to stick with it. 

I'm not favoring one over the other - I'm simply saying we should try to use one of them, unless the situation demands using the other alternative. 

We should also define that in the AGENTS.md, to ensure that generated AI code follows the communities guidelines.

internal/hostnetwork/underlay.go
```diff
@@ -134,35 +138,56 @@ func ensureLoopback(ctx context.Context, ns netns.NsHandle, vtepIP string) error
...
+		if l.Attrs().Group != underlayGroupID {
+			continue
 		}
-		for _, l := range links {
-			if l.Attrs().Group == underlayGroupID {
-				if err := netlink.LinkSetGroup(l, 0); err != nil {
-					return fmt.Errorf("failed to reset group ID on %s: %w", l.Attrs().Name, err)
-				}
-			}
+		if err = moveInterfaceToNamespace(ctx, l.Attrs().Name, fromNetNSHandle, defaultNetNSHandle, defaultNetNS, 0); err != nil {
```
maiqueb: what about the "other" interfaces ? they'll be stranded until the "first" one succeeds.

Shouldn't we continue, and aggregate errors ?

internal/hostnetwork/link_properties.go
```diff
@@ -173,67 +171,65 @@ func setNeighSuppression(link netlink.Link) error {
...
+	}
+
+	slog.DebugContext(ctx, "restoring addresses in namespace", "addresses", addresses)
+	for _, a := range addresses {
+		slog.DebugContext(ctx, "restoring address in namespace", "address", a, "flags", a.Flags)
+		a.Flags &= ^unix.IFA_F_NOPREFIXROUTE
+		slog.DebugContext(ctx, "restoring address in namespace after no prefix", "address", a, "flags", a.Flags)
+		err := toHandle.AddrAdd(nsLink, &a)
+		if err != nil && !os.IsExist(err) {
+			return fmt.Errorf("moveInterfaceToNamespace: Failed to add address %s to %s: %w", a, nsLink.Attrs().Name, err)
```
maiqueb: same here; what if we fail on one address ? what about the others ? 

shouldn't we aggregate errors and return them ?

internal/hostnetwork/link_properties.go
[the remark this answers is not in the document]
maiqueb: I agree with the bot (and you ...) but I'm OK deferring to a follow up. 
Tks for creating the issue.

maiqueb:
@andreaskaris could you resume work on this PR ? Do you need help ?

internal/hostnetwork/underlay.go
andreaskaris: yep, done next push

internal/hostnetwork/link_properties.go
andreaskaris: So that function is used to move the interface into the namespace, and out of it, on setup and restore. That part used to be +/- like that before, meaning it failed on the first one without trying the others. I changed it now to your suggestion.

andreaskaris:
Rebased, applied the 2 suggestions, but didn't run it locally through tests - waiting for CI to confirm that everything is o.k.

maiqueb (approved):
Thank you.

I'd like @fedepaol to review as well.

andreaskaris:
The CI failure is due to a crash in Zebra, probably the same known one:
```
STEP: checking reachability from hostARed to firstPod @ 06/15/26 17:17:48.542
  STEP: trying to hit 192.171.24.2 from 192.168.20.2 @ 06/15/26 17:17:48.542
• [30.509 seconds]
------------------------------
Routes between bgp and the fabric with Underlay in ipv4 should create two pods connected to the l2 overlay OVS bridge autocreate for dual stack
/home/runner/work/openperouter/openperouter/e2etests/tests/evpn_l2.go:276
  STEP: setting redistribute connected on leaves @ 06/15/26 17:18:04.816
  STEP: creating the pods @ 06/15/26 17:18:06.107
  STEP: removing the default gateway via the primary interface @ 06/15/26 17:18:11.658
  STEP: waiting for BGP sessions to establish on both nodes before traffic check @ 06/15/26 17:18:12.238
  STEP: checking reachability from firstPod to secondPod @ 06/15/26 17:18:12.338
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:12.339
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:16.492
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:20.656
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:24.813
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:28.975
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:33.196
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:37.356
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:41.517
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:45.678
  STEP: trying to hit 192.171.24.3 from 192.171.24.2 @ 06/15/26 17:18:49.836
  [FAILED] in [It] - /home/runner/work/openperouter/openperouter/e2etests/tests/evpn_l2.go:216 @ 06/15/26 17:18:52.996




ted_to_the_l2_overlay_OVS_bridge_autocreate_for_dual_stack]$ grep 'signal 6' -R -C1
openperouter-system_router-fstxc_pods_logs.log-2026/06/15 17:17:36.105 BGP: [JSYCM-MV07M] 0:Recv MACIP Del f 0x0 MAC 00:f3:00:00:00:6f IP 192.171.24.1 VNI 110 seq 0 state 1 ESI 00:00:00:00:00:00:00:00:00:00
openperouter-system_router-fstxc_pods_logs.log:ZEBRA: Received signal 6 at 1781543856 (si_addr 0x6400000047, PC 0x7f39a217acaf); aborting...
openperouter-system_router-fstxc_pods_logs.log-2026/06/15 17:17:36.105 BGP: [KPK7M-B9FAF] BGP announcing evpn route [2]:[0]:[48]:[00:00:00:00:00:00]:[32]:[192.171.24.1](VRF default) with dest 0x7f79d3089010 and flags 0x400 to zebra
--
openperouter-system_router-fstxc_pods_logs.log-2026/06/15 17:17:50.410 ZEBRA: [Y6R2N-EF2N4] interface pe-110 is being deleted from the system
openperouter-system_router-fstxc_pods_logs.log:ZEBRA: Received signal 6 at 1781543870 (si_addr 0x64000026b2, PC 0x7f2b56a42caf); aborting...
openperouter-system_router-fstxc_pods_logs.log-ZEBRA: in thread dplane_thread_loop scheduled from zebra/zebra_dplane.c:6835 dplane_provider_work_ready()
```

e2etests/pkg/openperouter/netns.go
```diff
@@ -139,3 +141,53 @@ func ifaceName(line string) (string, error) {
 	}
 	return strings.SplitN(parts[1], "@", 2)[0], nil
 }
+
+// IsUnderlayVethInDefaultNetns checks whether the interface exists
+// in the default network namespace on nodeName.
+func IsUnderlayVethInDefaultNetns(nodeName, intf string) bool {
```
fedepaol: Let's rename this "IsInterfaceInDefaultNs"

e2etests/pkg/openperouter/netns.go
```diff
@@ -139,3 +141,53 @@ func ifaceName(line string) (string, error) {
...
+	exec := executor.ForContainer(nodeName)
+	_, err := exec.Exec("ip", "link", "show", intf)
+	return err == nil
+}
+
+// IsUnderlayVethInPerouterNetns checks whether the interface exists
+// inside the perouter network namespace on nodeName.
+func IsUnderlayVethInPerouterNetns(nodeName, intf string) bool {
```
fedepaol: I's either call this isInterfaceInRouterNS OR export the router namespace as constant and convert this to isInterfaceInNS (and maybe factor with the one above with default namespace and logic against the "default" constant

e2etests/pkg/openperouter/netns.go
```diff
@@ -139,3 +141,53 @@ func ifaceName(line string) (string, error) {
...
+	exec := executor.ForContainer(nodeName)
+	_, err := exec.Exec("ip", "netns", "exec", namedNetns, "ip", "link", "show", intf)
+	return err == nil
+}
+
+var addrRegexp = regexp.MustCompile(`\s(inet6?\s+\S+)`)
+
+// UnderlayVethAddresses returns the non-link-local IP addresses assigned to the
+// interface in the default netns on nodeName, sorted and newline-joined.
+func UnderlayVethAddresses(nodeName, intf string) (string, error) {
```
fedepaol: same, let's talk about interfaces.
Also, not sure it makes sense to have it here in the netns.go

e2etests/pkg/openperouter/netns.go
fedepaol: and maybe move it to a netlink package if we don't have it already?

e2etests/pkg/openperouter/netns.go
```diff
@@ -139,3 +141,53 @@ func ifaceName(line string) (string, error) {
...
+		addr := m[1]
+		addrs = append(addrs, addr)
+	}
+	sort.Strings(addrs)
+	return strings.Join(addrs, "\n"), nil
+}
+
+// UnderlayVethIsUp checks whether the interface in the default netns
+// on nodeName has state UP.
+func UnderlayVethIsUp(nodeName, intf string) bool {
```
fedepaol: same considerations as above

internal/hostnetwork/link_properties.go
```diff
@@ -173,67 +172,70 @@ func setNeighSuppression(link netlink.Link) error {
 	return nil
 }
 
-// moveInterfaceToNamespace takes the given interface and moves it into the given namespace.
-func moveInterfaceToNamespace(ctx context.Context, intf string, ns netns.NsHandle) error {
-	slog.DebugContext(ctx, "move intf to namespace", "intf", intf, "namespace", ns.String())
-	defer slog.DebugContext(ctx, "move intf to namespace end", "intf", intf, "namespace", ns.String())
+// moveInterfaceToNamespace takes the given interface and moves it from the given to the given namespace.
+func moveInterfaceToNamespace(ctx context.Context, intf string, fromHandle, toHandle *netlink.Handle,
```
fedepaol, line 176: let's call to / from Handle sourceNamespace / targetNamespace (with or without handle, I don't have a strong opinion)

internal/hostnetwork/underlay.go
```diff
@@ -142,35 +146,58 @@ func ensureLoopback(ctx context.Context, ns netns.NsHandle, vtepIPs ...string) e
 // RemoveUnderlay removes the underlay state from the named network namespace:
 // it resets the group ID on underlay NICs (so HasUnderlayInterface
 // returns false on the next reconcile) and clears all IP addresses from the VTEP loopback (lo).
-func RemoveUnderlay(targetNS string) error {
-	ns, err := netns.GetFromPath(targetNS)
+func RemoveUnderlay(ctx context.Context, fromNetNSPath string) error {
```
fedepaol: should we call this RestoreUnderlay? Or, ReleaseUnderlay?

internal/hostnetwork/underlay.go
```diff
@@ -142,35 +146,58 @@ func ensureLoopback(ctx context.Context, ns netns.NsHandle, vtepIPs ...string) e
...
 
-		links, err := netlink.LinkList()
-		if err != nil {
-			return fmt.Errorf("failed to list links: %w", err)
+	fromNetNSHandle, err := netlink.NewHandleAt(fromNetNS)
+	if err != nil {
+		return fmt.Errorf("RemoveUnderlay: failed to get netlink handle for namespace %s: %w", fromNetNSPath, err)
+	}
+	defer fromNetNSHandle.Close()
+	defaultNetNSHandle, err := netlink.NewHandleAt(defaultNetNS)
```
fedepaol, line 211: this block _just to fetch the handles_ is annoying, but I am not sure how we can improve as we need the defers.
Maybe a wrapped removeUnderlay which does the work while the external one does the preamble so the separation is clearer?

internal/hostnetwork/underlay.go
```diff
@@ -240,13 +267,13 @@ func findUnderlayMTU(ns netns.NsHandle) (int, error) {
 	return minMTU, nil
 }
 
-func clearNonDefaultLoopbackIPs(intf string) error {
-	lo, err := netlink.LinkByName(intf)
+func clearNonDefaultLoopbackIPs(handle *netlink.Handle, intf string) error {
```
fedepaol: `handle` makes little sense here. Let's call it interfaceNSHandle or just nsHandle

fedepaol (commented):
mostly my usual naming idiosyncrasies , looks good overall

internal/hostnetwork/underlay.go
andreaskaris: done

internal/hostnetwork/underlay.go
andreaskaris: done

internal/hostnetwork/link_properties.go
andreaskaris: done

e2etests/pkg/openperouter/netns.go
andreaskaris: done

e2etests/pkg/openperouter/netns.go
andreaskaris: done

e2etests/pkg/openperouter/netns.go
andreaskaris: done

e2etests/pkg/openperouter/netns.go
andreaskaris: done .. moved to file interface.go

fedepaol: err... I think this wasn't done?

internal/hostnetwork/underlay.go
andreaskaris: done next push

e2etests/pkg/openperouter/netns.go
andreaskaris: sorry I pushed it all now

fedepaol:
@andreaskaris @maiqueb I guess now that we have this mechanism at hand, we could avoid killing the network namespace all together.

I am going to merge this and file an issue.

fedepaol:
Filed https://github.com/openperouter/openperouter/issues/531

PR #469, opened 2026-06-03
Mirror static resources to k8s api

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

When the k8s api is accessible, we might want to mirror and reconcile the static configurations to actual resources.
This gives us visibility and also allows the webhooks to avoid conflicting resources.

**Release note**:

```release-note
Mirror configuration consumed by static files to k8s resources for better visibility.
```

### DIFF
--- a/Claude.md
+++ b/Claude.md
@@ -23,6 +23,10 @@ Run `make test` to execute unit tests.
 - **Prerequisites:** Requires active deployment
 - **Before running:** Always verify cluster is running and deployment version is current
 - **Validation only:** You can run the command to check if e2e test code compiles, but confirm with the user before executing against a cluster
+- **Filtering tests:** Use `GINKGO_ARGS` with `--focus` to run specific tests, and `TEST_ARGS` for deployment mode:
+  ```
+  KUBECONFIG=bin/kubeconfig TEST_ARGS="--systemdmode" GINKGO_ARGS='--focus="Mirror static config"' make e2etests
+  ```
 
 **Development environment**
 
--- a/cmd/hostcontroller/main.go
+++ b/cmd/hostcontroller/main.go
@@ -265,8 +265,9 @@ func runK8sConfigReconcilerHostMode(ctx context.Context,
 		RouterHealthCheckPort: hostModeParams.routerHealthCheckPort,
 	}
 
-	// Create trigger channel for file watcher
+	// Create trigger channels for both controllers
 	triggerChan := make(chan event.GenericEvent, 1)
+	mirrorTriggerChan := make(chan event.GenericEvent, 1)
 
 	apiReconciler := &routerconfiguration.PERouterReconciler{
 		Client:          mgr.GetClient(),
@@ -305,15 +306,32 @@ func runK8sConfigReconcilerHostMode(ctx context.Context,
 		return fmt.Errorf("unable to start FRR restart watcher: %w", err)
 	}
 
-	// Setup file watcher to trigger API reconciler on static file changes
-	fw, err := filewatcher.New(hostModeParams.configurationDir, triggerChan, logger)
-	if err != nil {
-		return fmt.Errorf("unable to create file watcher for API reconciler: %w", err)
+	mirrorController := &routerconfiguration.MirrorController{
+		Client:      mgr.GetClient(),
+		Scheme:      mgr.GetScheme(),
+		Logger:      logger,
+		MyNode:      args.nodeName,
+		MyNamespace: args.namespace,
+		ConfigDir:   hostModeParams.configurationDir,
+		TriggerChan: mirrorTriggerChan,
 	}
 
-	if err := fw.Start(ctx); err != nil {
-		return fmt.Errorf("unable to start file watcher for API reconciler: %w", err)
+	if err := mirrorController.SetupWithManager(mgr); err != nil {
+		return fmt.Errorf("unable to create mirror controller: %w", err)
 	}
+
+	// Setup file watcher to trigger controllers on static file changes
+	filesChangedChan := make(chan event.GenericEvent, 1)
+	watcher, err := filewatcher.New(hostModeParams.configurationDir, filesChangedChan, logger)
+	if err != nil {
+		return fmt.Errorf("unable to create file watcher: %w", err)
+	}
+
+	if err := watcher.Start(ctx); err != nil {
+		return fmt.Errorf("unable to start file watcher: %w", err)
+	}
+
+	go fanOut(ctx, filesChangedChan, triggerChan, mirrorTriggerChan)
 
 	setupLog.Info("starting manager")
 	if err := mgr.Start(ctx); err != nil {
@@ -400,6 +418,8 @@ func runStaticConfigReconciler(ctx context.Context,
 		FRRReloadSocket: args.reloaderSocket,
 		RouterProvider:  staticRouterProvider,
 		ConfigDir:       hostModeParams.configurationDir,
+		MyNode:          args.nodeName,
+		MyNamespace:     args.namespace,
 	}
 	if err = staticReconciler.SetupWithManager(mgr); err != nil {
 		return fmt.Errorf("unable to create controller: %w", err)
@@ -571,4 +591,20 @@ func overrideHostMode(args *parameters, nodeConfig static.NodeConfig) error {
 	}
 	setupLog.Info("nodename not provided, using hostname", "nodename", args.nodeName)
 	return nil
+}
+
+func fanOut(ctx context.Context, in <-chan event.GenericEvent, outs ...chan<- event.GenericEvent) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case evt, ok := <-in:
+			if !ok {
+				return
+			}
+			for _, out := range outs {
+				out <- evt
+			}
+		}
+	}
 }
--- a/e2etests/tests/static_mirror.go
+++ b/e2etests/tests/static_mirror.go
@@ -1,0 +1,546 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package tests
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/openperouter/openperouter/api/v1alpha1"
+	"github.com/openperouter/openperouter/e2etests/pkg/config"
+	"github.com/openperouter/openperouter/e2etests/pkg/infra"
+	"github.com/openperouter/openperouter/e2etests/pkg/k8s"
+	"github.com/openperouter/openperouter/e2etests/pkg/k8sclient"
+	"github.com/openperouter/openperouter/e2etests/pkg/openperouter"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	clientset "k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	staticSourceLabelKey   = "openperouter.github.io/source"
+	staticSourceLabelValue = "static"
+)
+
+// staticUnderlayYAML is a minimal underlay for static mirroring tests.
+var staticUnderlayYAML = fmt.Sprintf(`underlays:
+  - asn: %d
+    nics:
+      - toswitch
+    neighbors:
+      - asn: 64512
+        address: "192.168.11.2"
+    tunnelEndpoint:
+      cidrs:
+        - "100.65.0.0/24"
+`, infra.Underlay.Spec.ASN)
+
+var staticL3VNIYAML = `l3vnis:
+  - vrf: mirror-red
+    vni: 8100
+    hostsession:
+      asn: 64514
+      hostasn: 64515
+      localcidr:
+        ipv4: "192.169.80.0/24"
+`
+
+var staticL3VNIUpdatedYAML = `l3vnis:
+  - vrf: mirror-red
+    vni: 8200
+    hostsession:
+      asn: 64514
+      hostasn: 64515
+      localcidr:
+        ipv4: "192.169.80.0/24"
+`
+
+var _ = Describe("Mirror static config to Kubernetes", Label("systemdmode"), Ordered, func() {
+	var cs clientset.Interface
+	var configPods []*corev1.Pod
+	var nodes []corev1.Node
+	var cl client.Reader
+
+	validateL3VNIs := func(expectedVNI int32) func() error {
+		return func() error {
+			l3vniList := &v1alpha1.L3VNIList{}
+			err := cl.List(context.Background(), l3vniList,
+				client.InNamespace(openperouter.Namespace),
+				client.MatchingLabels{staticSourceLabelKey: staticSourceLabelValue})
+			if err != nil {
+				return fmt.Errorf("failed to list l3vnis: %w", err)
+			}
+			if len(l3vniList.Items) < len(nodes) {
+				return fmt.Errorf("expected at least %d mirrored l3vnis, got %d",
+					len(nodes), len(l3vniList.Items))
+			}
+			for _, v := range l3vniList.Items {
+				if v.Spec.NodeSelector == nil {
+					return fmt.Errorf("mirrored L3VNI %s has nil node selector", v.Name)
+				}
+				if v.Spec.VNI != expectedVNI {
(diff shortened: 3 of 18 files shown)


cmd/hostcontroller/main.go
```diff
@@ -98,6 +98,27 @@ type parameters struct {
 	logLevel       string
 }
 
+// fanOut fans out events from in channel to multiple out channels.
+// It runs until ctx is cancelled or in channel is closed.
+func fanOut(ctx context.Context, in <-chan event.GenericEvent, outs ...chan<- event.GenericEvent) {
```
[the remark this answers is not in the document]
maiqueb: bot's right. move this to the bottom.

maiqueb: so it echoes whatever arrives at the in channel over to the variadic output channels, right ?

cmd/hostcontroller/main.go
```diff
@@ -98,6 +98,27 @@ type parameters struct {
...
+		case <-ctx.Done():
+			return
+		case evt, ok := <-in:
+			if !ok {
+				return
+			}
+			for _, out := range outs {
+				select {
+				case out <- evt:
+				default:
```
maiqueb: hm ... why do you need this select, case , default ? couldn't you just `out <- evt` ?

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		return ctrl.Result{}, err
+	}
+
+	if err := r.mirrorAndClean(ctx, &v1alpha1.L3VNIList{}, toObjects(staticConfig.L3VNIs), logger, "l3vni"); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.mirrorAndClean(ctx, &v1alpha1.L2VNIList{}, toObjects(staticConfig.L2VNIs), logger, "l2vni"); err != nil {
+		return ctrl.Result{}, err
+	}
```
maiqueb, line 63: so if something goes wrong with the underlay, we halt. That I understand. 

But shouldn't we try to follow what we've done in the config resiliency here ? i.e. provision per VRF, or, at the very least, try to keep provisioning all resources, and aggregate errors ?

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		return fmt.Errorf("failed to list mirrored %ss: %w", kind, err)
+	}
+
+	existing := extractItems(listObj)
+	existingByName := map[string]client.Object{}
+	for _, item := range existing {
+		existingByName[item.GetName()] = item
+	}
+
+	desiredNames := map[string]bool{}
```
maiqueb: nit: afaiu, we don't need a `map[string]bool` - a `map[string]struct{}` would suffice.

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	for _, item := range existing {
+		if desiredNames[item.GetName()] {
+			continue
+		}
+		logger.Info("deleting stale mirrored resource", "kind", kind, "name", item.GetName())
+		if err := r.Delete(ctx, item); err != nil {
+			return fmt.Errorf("failed to delete stale %s %s: %w", kind, item.GetName(), err)
+		}
```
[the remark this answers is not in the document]
maiqueb: oh, good bot ! 

nice catch.

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+			return err
+		}
+	}
+
+	for _, item := range existing {
+		if desiredNames[item.GetName()] {
+			continue
+		}
+		logger.Info("deleting stale mirrored resource", "kind", kind, "name", item.GetName())
+		if err := r.Delete(ctx, item); err != nil {
```
maiqueb: should we try to keep on, and aggregate errors ? i.e. should we fail the entire reconcile cycle if one resource failed ?

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		existingByName[item.GetName()] = item
+	}
+
+	desiredNames := map[string]bool{}
+	for _, d := range desired {
+		desiredNames[d.GetName()] = true
+		if ex, ok := existingByName[d.GetName()]; ok && resourceMatchesDesired(ex, d) {
+			continue
+		}
+		if _, err := r.createOrUpdateMirrored(ctx, d, logger, kind); err != nil {
```
maiqueb, line 122: should we try to keep on, and aggregate errors ? i.e. should we fail the entire reconcile cycle if one resource failed ?

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	return nil
+}
+
+// createOrUpdateMirrored creates or updates a single mirrored K8s resource.
+// It uses type assertion to copy the spec from the desired object.
+func (r *MirrorController) createOrUpdateMirrored(ctx context.Context, desired client.Object, logger *slog.Logger, kind string) (controllerutil.OperationResult, error) {
+	obj := newEmptyObject(desired)
+	obj.SetName(desired.GetName())
```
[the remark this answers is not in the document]
maiqueb: hm, I'd rather protect that in the caller. @gemini-code-assist what do you think about that ?

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		dst.(*v1alpha1.Underlay).Spec = s.Spec
+	case *v1alpha1.L3VNI:
+		dst.(*v1alpha1.L3VNI).Spec = s.Spec
+	case *v1alpha1.L2VNI:
+		dst.(*v1alpha1.L2VNI).Spec = s.Spec
+	case *v1alpha1.L3Passthrough:
+		dst.(*v1alpha1.L3Passthrough).Spec = s.Spec
+	case *v1alpha1.RawFRRConfig:
+		dst.(*v1alpha1.RawFRRConfig).Spec = s.Spec
+	}
```
maiqueb, line 187: shouldn't we deep-copy that ?

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func resourceMatchesDesired(existing, desired client.Object) bool {
+	if !reflect.DeepEqual(existing.GetLabels(), desired.GetLabels()) {
+		return false
+	}
+	switch e := existing.(type) {
+	case *v1alpha1.Underlay:
+		return reflect.DeepEqual(e.Spec, desired.(*v1alpha1.Underlay).Spec)
```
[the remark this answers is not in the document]
maiqueb: good bot ! @fedepaol please take this suggestion into account.

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil
+	}
+
+	items := make([]client.Object, itemsField.Len())
+	for i := 0; i < itemsField.Len(); i++ {
+		items[i] = itemsField.Index(i).Addr().Interface().(client.Object)
+	}
+	return items
+}
+
```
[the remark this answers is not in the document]
maiqueb: I find this more readable as well. And it looks like it might work.

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+// SetupWithManager registers the mirror controller with the manager.
+func (r *MirrorController) SetupWithManager(mgr ctrl.Manager) error {
+	// Predicate that matches only resources with the static source label for this node
```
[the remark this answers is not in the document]
maiqueb: yeap, this is true.

internal/controller/routerconfiguration/mirror_controller_test.go
```diff
@@ -0,0 +1,576 @@
...
+			}
+
+			mc := newMirrorController(t, dir, append([]client.Object{testNode()}, tt.preExisting...)...)
+			reconcile(t, mc)
+
+			// Check initial state
+			assertExpectedResources(t, mc, tt.expectedCRs)
+
+			// Handle external deletion test case
+			if tt.name == "recreates externally deleted resource" {
```
maiqueb: this is super ugly. I mean ... we have a table, and we have an if checking if we're handling a particular test ... 

Could we isolate this test from the table ? I think that's way cleaner.

internal/controller/routerconfiguration/mirror_controller_test.go
```diff
@@ -0,0 +1,576 @@
...
+				"openpe_l2vni.yaml": `l2vnis:
+  - vni: 300
+    vxlanport: 4789
+`,
+			},
+			expectedCRs: expectedResources{
+				underlays: 1,
+				l3vnis:    2,
+				l2vnis:    1,
+			},
```
maiqueb, line 91: remind me ... why don't we validate the L3VNIs or the L2VNIs ?

internal/controller/routerconfiguration/mirror_controller_test.go
```diff
@@ -0,0 +1,576 @@
...
+			updateFiles: map[string]string{
+				"openpe_l3vni.yaml": `l3vnis:
+  - vrf: red
+    vni: 100
+`,
+			},
+			updatedCRs: &expectedResources{
+				l3vnis: 1,
+			},
```
maiqueb, line 152: hm, maybe I'm being a bit pedantic, but, shouldn't we ensure we deleted the *right* VNI / VRF ? ...

e2etests/tests/static_mirror.go
```diff
@@ -0,0 +1,500 @@
...
+				podConfigMount, staticUnderlayYAML))
+			Expect(err).NotTo(HaveOccurred())
+
+			_, err = execInConfigPod(pod, fmt.Sprintf("cat > %s/openpe_l3vni.yaml << 'EOF'\n%s\nEOF",
+				podConfigMount, staticL3VNIYAML))
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		By("waiting for mirrored resources to appear with static source label")
+		Eventually(validateL3VNIs(8100), "60s", "2s").Should(Succeed())
```
maiqueb, line 236: why don't we validate the underlay ?

e2etests/tests/static_mirror.go
```diff
@@ -0,0 +1,500 @@
...
+		}
+
+		By("waiting for new mirrored L2VNI to appear")
+		Eventually(validateL2VNIs(8300), "30s", "2s").Should(Succeed())
+	})
+
+	It("deletes mirrored resource when static file is removed", func() {
+		By("removing L2VNI static file from all nodes")
+		for _, pod := range configPods {
+			_, err := execInConfigPod(pod, fmt.Sprintf("rm -f %s/openpe_l2vni.yaml", podConfigMount))
```
maiqueb, line 292: where was this added ? In the previous test ? 

If so, I don't like it. While these tests are `Ordered` I think they should be able to run independently. 

Let's provision these things in a before each block please.

internal/controller/routerconfiguration/mirror_controller_test.go
maiqueb, line 1: I think this file got into the wrong commit. It should be squashed into the previous commit.

Claude.md
```diff
@@ -23,6 +23,10 @@ Run `make test` to execute unit tests.
 - **Prerequisites:** Requires active deployment
 - **Before running:** Always verify cluster is running and deployment version is current
 - **Validation only:** You can run the command to check if e2e test code compiles, but confirm with the user before executing against a cluster
+- **Filtering tests:** Use `GINKGO_ARGS` with `--focus` to run specific tests, and `TEST_ARGS` for deployment mode:
+  ```
+  KUBECONFIG=bin/kubeconfig TEST_ARGS="--systemdmode" GINKGO_ARGS='--focus="Mirror static config"' make e2etests
+  ```
```
maiqueb, line 29: this is totally unrelated to this PR if you ask me, but sure. let's let it in

maiqueb (changes requested):
you're missing the sign off on all commits except the last.

cmd/hostcontroller/main.go
fedepaol: right, it's a way to send the same event to multiple consumers

cmd/hostcontroller/main.go
fedepaol: it's a safeguard to avoid blocking if the receiving channel is full (which shouldn't happen)

internal/controller/routerconfiguration/mirror_controller.go
fedepaol: we are not provisioning anything here, just mirroring a configuration that is being handled elsewhere.

My reasoning was that here we are just mirroring so the type of error we can get here is just a temporary one, but you are right and in theory (albeit very unlikely) we might even end up with a configuration where the webhook is blocking the mirroring because it's clashing with what comes from the api. I'll collect the errors.

internal/controller/routerconfiguration/mirror_controller_test.go
fedepaol: what do you mean?

maiqueb: I only see you actually validate the underlay in these tests - not the L2VNIs or L3VNIs. 

Hence the question.

e2etests/tests/static_mirror.go
```diff
@@ -0,0 +1,500 @@
...
+			Spec: v1alpha1.L3VNISpec{
+				VRF: "conflict-vrf",
+				VNI: 8200,
+				HostSession: &v1alpha1.HostSession{
+					ASN:     64514,
+					HostASN: new(int64(64518)),
+					LocalCIDR: v1alpha1.LocalCIDRConfig{
+						IPv4: new("192.169.90.0/24"),
+					},
+				},
```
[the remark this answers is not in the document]
fedepaol: this is outdated

internal/controller/routerconfiguration/mirror_controller.go
fedepaol: it's a matter of taste. The saving in terms of memory is negligible but the if reads better imo. In any case, changing

Claude.md
fedepaol: it is, but I always thought it'd fine as long as it's in a separate commit and self contained enough. For small fixes during PRs, I think it's reasonable but if you think it's distracting I'll start filing separate PRs.

maiqueb: no need to. it's OK.

internal/controller/routerconfiguration/mirror_controller.go
fedepaol: done

internal/controller/routerconfiguration/mirror_controller.go
fedepaol: done

internal/controller/routerconfiguration/mirror_controller.go
```diff
@@ -0,0 +1,235 @@
...
+		if err := client.IgnoreNotFound(r.Delete(ctx, item)); err != nil {
+			mirrorErrors = append(mirrorErrors, fmt.Errorf("failed to delete stale %s %s: %w", kind, item.GetName(), err))
+		}
+	}
+
+	return errors.Join(mirrorErrors...)
+}
+
+func (r *MirrorController) createOrUpdateMirrored(ctx context.Context, desired client.Object, logger *slog.Logger, kind string) (controllerutil.OperationResult, error) {
+	obj := newEmptyObject(desired)
```
maiqueb: can the bot be right @fedepaol ? can we ever get a nil from this ? Still, I think we should protect ourselves, and cmp against nil.

internal/controller/routerconfiguration/mirror_controller.go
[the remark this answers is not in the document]
maiqueb: I would vote updating `newEmptyObject` signature to return an error as well.

maiqueb (changes requested):
Borderline nit, but I still think we should protect ourselves. 

After this, we can merge.

internal/controller/routerconfiguration/mirror_controller.go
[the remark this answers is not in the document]
fedepaol: done

maiqueb (approved):
Thank you.

PR #472, opened 2026-06-08
Deploy in systemd mode using an interface name to calculate nodeIndex

### LINKED ISSUE #331: When running in systemd mode, retrieve the node index from a device instead of using the configuration file
A lot of our logic depends on the node index. This mostly includes all the ipam logic for any place where we need to assign an IP.
When running in k8s mode, relying on the bus provided by the k8s api to assign an unique node index provides a good ux as the user doesn't have to bother with it.

When we run in systemd mode though, we are expect the user to drop a different configuration on each node, providing a different node index.

The proposal here is to optionally retrieve the node index from the IP of a well known interface (ie assuming all the nodes are on the same subnet, we could retrieve the node index from the last two octets of its ipv4 address).

yahlifried:
/kind feature

this pr adds a new way to deploy in systemd mode using interface name instead of the existing nodeIndex param in teh static config file. it will support openshift deployment where files can not be written to the disk, and instead use a shared interface.

```release-note
Allow deriving the node index from a network interface address in systemd mode, enabling the same node-config.yaml to be deployed across all nodes.
```

fixes #331


internal/controller/nodeindex/annotate.go
```diff
@@ -0,0 +1,44 @@
...
+	"strconv"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+)
+
+func AnnotateNodeIndex(ctx context.Context, clientset kubernetes.Interface, nodeName string, nodeIndex int) error {
```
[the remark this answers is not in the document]
yahlifried: the const is defined in nodes_allocation.go under the same package

internal/staticconfiguration/nodeindex.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/openperouter/openperouter/api/static"
+)
+
+// ResolveNodeIndex returns the node index from the configuration.
+// If NodeIndexInterfaceName is set, the index is derived from the
+// host portion of the first IPv4 address on that interface.
+// NodeIndex and NodeIndexInterfaceName are mutually exclusive.
+func ResolveNodeIndex(config *static.NodeConfig) (int, error) {
+	if config.NodeIndex != 0 && config.NodeIndexInterfaceName != "" {
```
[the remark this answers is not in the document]
yahlifried: currently checked in the caller function
should I add this for future calls that might not check it? @fedepaol

internal/staticconfiguration/nodeindex.go
```diff
@@ -0,0 +1,67 @@
...
+	ip4 := ipNet.IP.To4()
+	mask := ipNet.Mask
+
+	hostPart := 0
+	for i := range 4 {
+		hostPart = (hostPart << 8) | int(ip4[i] & ^mask[i])
+	}
+
+	return hostPart
+}
```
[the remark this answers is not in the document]
yahlifried: Done

fedepaol:
please align the note field in the PR description with a small sentence describing what was changed here

internal/staticconfiguration/nodeindex.go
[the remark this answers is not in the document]
fedepaol: I would argue on why we are passing a pointer in first place, as we are not changing it

internal/staticconfiguration/nodeindex.go
```diff
@@ -0,0 +1,77 @@
...
+	if config.NodeIndexInterfaceName != "" {
+		return NodeIndexFromInterface(config.NodeIndexInterfaceName)
+	}
+
+	return config.NodeIndex, nil
+}
+
+// NodeIndexFromInterface derives a node index from the host portion
+// of the first IPv4 address found on the named network interface.
+func NodeIndexFromInterface(name string) (int, error) {
```
fedepaol: if this is local, should be private

cmd/hostbridge/node_index.go
```diff
@@ -4,32 +4,16 @@ package main
...
-
-// annotateCurrentNode adds the node index annotation to the current node based on
-// the content of the static configuration file.
-func annotateCurrentNode(ctx context.Context, nodeConfigPath, nodeName string) error {
-	nodeConfig, err := staticconfiguration.ReadNodeConfig(nodeConfigPath)
-	if err != nil {
-		return fmt.Errorf("failed to read node configuration from %s: %w", nodeConfigPath, err)
-	}
-
+func annotateCurrentNode(ctx context.Context, nodeConfig *static.NodeConfig, nodeName string) error {
```
fedepaol: same here, let's pass by value if we are not changing it

cmd/hostbridge/main.go
```diff
@@ -51,13 +52,25 @@ func main() {
...
-			"error", err, "config_path", config.HostConfigPath, "node", config.NodeName)
+	nodeConfig, err := staticconfiguration.ReadNodeConfig(config.HostConfigPath)
+	if err != nil {
+		slog.Error("Failed to read node configuration",
+			"error", err, "config_path", config.HostConfigPath)
 		os.Exit(1)
 	}
 
-	slog.Info("Successfully added node index annotation", "config_path", config.HostConfigPath, "node", config.NodeName)
+	if nodeConfig.NodeIndexInterfaceName != "" {
```
fedepaol: at this point I'd simplify the logic and let hostcontroller always annotate the node

cmd/hostbridge/main.go
```diff
@@ -51,13 +52,25 @@ func main() {
 	sigChan := make(chan os.Signal, 1)
 	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
 
-	if err := annotateCurrentNode(ctx, config.HostConfigPath, config.NodeName); err != nil {
-		slog.Error("Failed to add node index annotation",
-			"error", err, "config_path", config.HostConfigPath, "node", config.NodeName)
+	nodeConfig, err := staticconfiguration.ReadNodeConfig(config.HostConfigPath)
```
fedepaol: this must fail if nodeindex and nodeindexinterface name are set at the same time

internal/staticconfiguration/nodeindex.go
```diff
@@ -0,0 +1,77 @@
...
+	"net"
+
+	"github.com/openperouter/openperouter/api/static"
+)
+
+// ResolveNodeIndex returns the node index from the configuration.
+// If NodeIndexInterfaceName is set, the index is derived from the
+// host portion of the first IPv4 address on that interface.
+// NodeIndex and NodeIndexInterfaceName are mutually exclusive.
+func ResolveNodeIndex(config *static.NodeConfig) (int, error) {
```
fedepaol: I would check if we need to use NodeIndexFromInterface on the calling site instead of wrapping.

So it would be
nodeIndex = config.NodeIndex
if onfig.NodeIndexInterfaceName != ""
   nodeIndex, err = NodeIndexFromInterface(config.NodeIndexInterfaceName)

api/static/config.go
```diff
@@ -7,9 +7,10 @@ import (
 )
 
 type NodeConfig struct {
-	NodeIndex int    `json:"nodeIndex"`
-	NodeName  string `json:"nodeName"`
-	LogLevel  string `json:"logLevel"`
+	NodeIndex              int    `json:"nodeIndex"`
```
fedepaol: I would use a sub struct for the nodeIndex logic.
```go
type NodeIndex struct {
   Index int
   InterfaceName string
`

I am tempted to say it'd be nice to provide the cidr we want to use to select the address in case of multiple addresses. 'That would enable us also to use v6 addresses too.

File: cmd/hostbridge/main.go (hunk 7 above)
Comment (yahlifried) [reply cmd/hostbridge/main.go]: added nodeConfig.NodeIndex.Validate()

File: internal/staticconfiguration/nodeindex.go (hunk 4 above)
Comment (yahlifried) [reply internal/staticconfiguration/nodeindex.go]: not local anymore because of other changes

File: internal/staticconfiguration/nodeindex.go (hunk 2 above)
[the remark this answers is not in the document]
Comment (yahlifried) [reply internal/staticconfiguration/nodeindex.go]: Done

File: cmd/hostbridge/node_index.go (hunk 5 above)
Comment (yahlifried) [reply cmd/hostbridge/node_index.go]: Done

File: cmd/hostbridge/main.go (hunk 6 above)
Comment (yahlifried) [reply cmd/hostbridge/main.go]: Done

File: internal/staticconfiguration/nodeindex.go (hunk 8 above)
Comment (yahlifried) [reply internal/staticconfiguration/nodeindex.go]: Done

File: api/static/config.go (hunk 9 above)
Comment (yahlifried) [reply api/static/config.go]: Done

File: cmd/hostcontroller/main.go [hunk 10]
```diff
@@ -192,6 +193,18 @@ func runHostMode(
 		logger.Error("failed to load the node configuration file", "error", err)
 		os.Exit(1)
 	}
+	if err := staticconfiguration.ValidateNodeIndex(nodeConfig.NodeIndex); err != nil {
```
fedepaol: nit: let's embed this in ReadNodeConfig and let's have ReadNodeConfig return a valid config

internal/staticconfiguration/nodeindex.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to get addresses for interface %s: %w", name, err)
+	}
+
+	for _, addr := range addrs {
+		ipNet, ok := addr.(*net.IPNet)
+		if !ok {
+			continue
+		}
+		if ipNet.IP.To4() == nil {
```
fedepaol: sorry to raise this so late 😅 , but I think it'd make sense to consider the ipv6 scenario too, so instead of continuing we should understand if the cidr is a v6 one and act accordingly. 
I would still: 
- default to v4
- fallback to v6

if nothing is specified

cmd/hostcontroller/main.go
yahlifried: Done

internal/staticconfiguration/nodeindex.go
yahlifried: Done

fedepaol: one last thing, let's provide an optional cidr in case the interface has multiple ips. It will also help if the user wants to pick an ipv6 cidr over an ipv4 one

yahlifried: Done

internal/staticconfiguration/nodeindex.go
```diff
@@ -0,0 +1,103 @@
...
+			if cidrNet.Contains(ipNet.IP) {
+				return hostPartFromIPNet(ipNet), nil
+			}
+			continue
+		}
+		if ipNet.IP.To4() != nil {
+			return hostPartFromIPNet(ipNet), nil
+		}
+		if v6Fallback == nil {
+			v6Fallback = ipNet
```
fedepaol, line 75: let's ensure it's a global ipv6 address

internal/staticconfiguration/nodeindex.go
```diff
@@ -0,0 +1,103 @@
...
+
+func hostPartFromIPNet(ipNet *net.IPNet) int {
+	ip := ipNet.IP
+	mask := ipNet.Mask
+
+	if ip4 := ip.To4(); ip4 != nil {
+		if len(mask) == 16 {
+			mask = mask[12:]
+		}
+		if len(mask) < 4 {
```
fedepaol: we can use stdlib for this:


```
       // AndNot(ip, mask) = ip & (^mask), zeroing the network bits and keeping the host bits.
       // e.g. 192.168.1.67/28:
       //   IP:       11000000.10101000.00000001.01000011
       //   Mask:     11111111.11111111.11111111.11110000
       //   ^Mask:    00000000.00000000.00000000.00001111
       //   IP&^Mask: 00000000.00000000.00000000.00000011 = 3
       host := new(big.Int).AndNot(
               new(big.Int).SetBytes(ip),
               new(big.Int).SetBytes(mask),
       )
       return int(host.Int64())
```

internal/staticconfiguration/nodeindex.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to find interface %s: %w", name, err)
+	}
+
+	addrs, err := iface.Addrs()
+	if err != nil {
+		return 0, fmt.Errorf("failed to get addresses for interface %s: %w", name, err)
+	}
+
+	var cidrNet *net.IPNet
```
fedepaol: let's split this behavior in two, as it's different if we have the user provided cidr or not.
So I'd have a findIndexFromCIDR and a findIndexFromAddresses invoked in one way or the other

internal/staticconfiguration/reader.go
```diff
@@ -39,6 +37,16 @@ func ReadNodeConfig(path string) (*static.NodeConfig, error) {
 		return nil, fmt.Errorf("failed to parse YAML node config: %w", err)
 	}
 
+	if err := ValidateNodeIndex(config.NodeIndex); err != nil {
+		return nil, err
+	}
+	if config.NodeIndex.InterfaceName != "" {
```
fedepaol: let's flip and return early so we remove one indentation level

fedepaol (changes requested):

internal/staticconfiguration/nodeindex.go
yahlifried: Done

internal/staticconfiguration/nodeindex.go
yahlifried: Done

internal/staticconfiguration/nodeindex.go
yahlifried: Done

internal/staticconfiguration/reader.go
yahlifried: Done

fedepaol:
LGTM, cc @maiqueb if you want to have a look

@yahlifried please fix the commits. You need a feat: before the commit subject I guess

fedepaol:
@yahlifried mind fixing the conflicts?

cmd/hostcontroller/main.go
```diff
@@ -49,7 +49,8 @@ import (
 	"github.com/openperouter/openperouter/api/static"
 	periov1alpha1 "github.com/openperouter/openperouter/api/v1alpha1"
 	"github.com/openperouter/openperouter/internal/buildversion"
-	"github.com/openperouter/openperouter/internal/cni"
+  "github.com/openperouter/openperouter/internal/cni"
```
fedepaol: @yahlifried this is not aligned

yahlifried: Done

fedepaol:
Thanks!

PR #474, opened 2026-06-08
Rm IPFamily from passthr/l3vni frr_test.go, fix local_passthrough.tmpl

andreaskaris:

/kind cleanup

The IPFamily field for passthrough LocalNeighborV4/V6 and L3VNI
LocalNeighbor is never set by production code. For passthrough,
the underlay address family already has a one-to-one mapping to
the BGP address family (LocalNeighborV4 -> AF IPv4, LocalNeighborV6
-> AF IPv6), so the neighborenableipfamily template invocation is
a) redundant and b) its logic is actually never executed.

Remove the neighborenableipfamily template calls from
local_passthrough.tmpl and drop the IPFamily field from tests to
match production behavior. Adjust testdata accordingly.

The `neighborenableipfamily` template is still used for global sessions in `frr.tmpl` - only the passthrough usage is removed and tests are fixed for passthrough and l3vni.

**Release note**:

```release-note
Fix duplicate address-family blocks in FRR passthrough configuration by removing unused neighborenableipfamily template calls. Remove IPFamily from frr_test.go for l3vni and passthrough input as it is never set by production code.
```

### DIFF
--- a/internal/frr/testdata/TestPassthroughDual.golden
+++ b/internal/frr/testdata/TestPassthroughDual.golden
@@ -44,11 +44,6 @@ router bgp 64512
     neighbor 192.168.1.3 route-map allowall out
   exit-address-family
 
-  address-family ipv4 unicast
-    neighbor 192.168.1.3 activate
-    neighbor 192.168.1.3 allowas-in
-  exit-address-family
-
   neighbor 2001:db8:20::2 remote-as 64513
 
   address-family ipv6 unicast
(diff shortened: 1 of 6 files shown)


internal/frr/testdata/TestPassthroughDual.golden
```diff
@@ -44,11 +44,6 @@ router bgp 64512
     neighbor 192.168.1.3 route-map allowall out
   exit-address-family
 
-  address-family ipv4 unicast
-    neighbor 192.168.1.3 activate
```
andreaskaris, line 48: this was redundant

internal/frr/testdata/TestPassthroughDual.golden
```diff
@@ -44,11 +44,6 @@ router bgp 64512
...
-    neighbor 192.168.1.3 allowas-in
```
andreaskaris, line 49: this would never have been set in production for localneighbors, because the IPFamily isn't set on them

PR #484, opened 2026-06-11
Introduce inspect tool for debugging

ormergi:

>
> /kind bug
> /kind cleanup

/kind feature

> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

Debugging k8s controllers is tedious require knowing details how controller is deployed and  usually involve manually dumping multiple objects and logs.
We want to improve debug UX, at least for collecting relevant info for troubleshooting.

Introduce `inspect` tool for collecting OpenPERouter deployment state for easier debugging.

The tool dumps related objects such as pods, pod logs, CRDs, node router 
network and router (frr instance) information.

How to use: 
Inspect openperouter deployment:
```
$ make inspect

# override parameters 
$ KUBECONFIG_PATH=$KUBECONFIG  make inspect \
    NAMESPACE=openshift-openperouter
    KUBECTL=oc \
    INSPECT_DIR=./art \
    SINCE=1d

# run script directly
$ ./tools/inspect/inspect 

# override parameters
$ ./tools/inspect/inspect \
    --namespace=openshift-openperouter \
    --k8s-client=oc \
    --dest-dir=/tmp/openperouter-logs
    --since=1d
```
Inspect nodes running systemd mode:
```bash
# run script directly on node
$ ssh mynode.example.io -- bash <<< $(cat ./tools/inspect/inspect_host)
$ scp -r mynode.example.io:/openperouter-inspect-host ./openperoute-inspect-mynode.example.io

# debug kind cluster node
$ docker exec pe-kind-worker -i bash <<< $(cat ./tools/inspect/inspect_host)
$ docker cp pe-kind-worker:/openperouter-inspect-host ./pe-kind-worker-inspect-host
# via repo make target
$ make inspect-systemd-mode
```

Please see the added the README files, it contain more info about usage and output.

Since on systemd mode info cannot be collected via cluster API, a separate script is add to enable collecting info of node running on systemd mode. 
The script is meant to run directly on a target node, output can be copied to base station for inspection.

**Release note**:

```release-note
Introduce tool for inspecting OpenPERouter deployments for easier troubleshooting.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -703,7 +703,7 @@ grout-docker-build: docker-build
 INSPECT_DIR ?= /tmp/openperouter-inspect
 .PHONY: inspect
 inspect:
-	tools/inspect/inspect --k8s-client=$(KUBECTL) --namespace=$(NAMESPACE) --dest-dir=$(INSPECT_DIR)
+	tools/inspect/inspect --k8s-client=$(KUBECTL) --namespace=$(NAMESPACE) --dest-dir=$(INSPECT_DIR) --since=$(SINCE)
 
 HOST_INSPECT_DIR = /openperouter-inspect-host
 .PHONY: inspect-host
--- a/Makefile
+++ b/Makefile
@@ -705,11 +705,22 @@ INSPECT_DIR ?= /tmp/openperouter-inspect
 inspect:
 	tools/inspect/inspect --k8s-client=$(KUBECTL) --namespace=$(NAMESPACE) --dest-dir=$(INSPECT_DIR) --since=$(SINCE)
 
-HOST_INSPECT_DIR = /openperouter-inspect-host
-.PHONY: inspect-host
-inspect-host:
-	@ ! test -n "$(NODE)" && echo "target node is not set, make sure to pass NODE argument" && exit 1; \
-	NODE_INSPECT_DIR="/tmp/$(NODE)-inspect"; \
-	$(CONTAINER_ENGINE) exec -i $(NODE) bash <<< $$(cat tools/inspect/inspect_host); \
-	$(CONTAINER_ENGINE) cp $(NODE):$(HOST_INSPECT_DIR) $$NODE_INSPECT_DIR; \
-	echo "Inspect node $(NODE) completed. Artifacts are stored at $$NODE_INSPECT_DIR"
+# INSPECT_SYSTEMD_MODE_DIR is the root artifacts directory on the local machine, containing
+# one subdirectory per inspected node.
+INSPECT_SYSTEMD_MODE_DIR ?= /tmp/openperouter-systemd-mode-inspect
+# INSPECT_NODE_DIR is the path inside each node where inspect_host stores its artifacts
+INSPECT_NODE_DIR = /openperouter-inspect-host
+.PHONY: inspect-systemd-mode
+inspect-systemd-mode: kubectl
+	@ nodes="$(NODES)"; \
+	test -z "$$nodes" && nodes=$$($(KUBECTL) get nodes -o jsonpath='{.items[*].metadata.name}'); \
+	test -z "$$nodes" && echo "no node found" && exit 1 ;\
+	mkdir -p $(INSPECT_SYSTEMD_MODE_DIR); \
+	for node in $$nodes ; do \
+		echo "=== Inspecting node $$node ==="; \
+		node_dir="$(INSPECT_SYSTEMD_MODE_DIR)/$$node"; \
+		$(CONTAINER_ENGINE) exec -i $$node bash <<< $$(cat tools/inspect/inspect_host); \
+		$(CONTAINER_ENGINE) cp $$node:$(INSPECT_NODE_DIR) $$node_dir; \
+		echo "Inspect node $$node completed. Artifacts are stored at [$$node_dir]"; \
+	done; \
+	echo "Inspect completed for all nodes. Artifacts are stored at [$(INSPECT_SYSTEMD_MODE_DIR)]"
--- a/Makefile
+++ b/Makefile
@@ -678,7 +678,6 @@ deploy-olm: operator-sdk ## deploys OLM on the cluster
 
 build-and-push-bundle-images: bundle-build bundle-push catalog-build catalog-push
 
-
 .PHONY: grout-deploy
 grout-deploy: IMG_TAG=main-grout
 grout-deploy: export KUSTOMIZE_LAYER=grout
@@ -700,3 +699,8 @@ grout-deploy-helm: helm kind deploy-cluster load-on-kind deploy-helm
 grout-docker-build: IMG_TAG=main-grout
 grout-docker-build: DOCKERFILE=Dockerfile.grout
 grout-docker-build: docker-build
+
+INSPECT_DIR ?= /tmp/openperouter-inspect
+.PHONY: inspect
+inspect:
+	tools/inspect/inspect --k8s-client=$(KUBECTL) --namespace=$(NAMESPACE) --dest-dir=$(INSPECT_DIR)
--- a/Makefile
+++ b/Makefile
@@ -704,3 +704,12 @@ INSPECT_DIR ?= /tmp/openperouter-inspect
 .PHONY: inspect
 inspect:
 	tools/inspect/inspect --k8s-client=$(KUBECTL) --namespace=$(NAMESPACE) --dest-dir=$(INSPECT_DIR)
+
+HOST_INSPECT_DIR = /openperouter-inspect-host
+.PHONY: inspect-host
+inspect-host:
+	@ ! test -n "$(NODE)" && echo "target node is not set, make sure to pass NODE argument" && exit 1; \
+	NODE_INSPECT_DIR="/tmp/$(NODE)-inspect"; \
+	$(CONTAINER_ENGINE) exec -i $(NODE) bash <<< $$(cat tools/inspect/inspect_host); \
+	$(CONTAINER_ENGINE) cp $(NODE):$(HOST_INSPECT_DIR) $$NODE_INSPECT_DIR; \
+	echo "Inspect node $(NODE) completed. Artifacts are stored at $$NODE_INSPECT_DIR"
--- a/tools/inspect/README.md
+++ b/tools/inspect/README.md
@@ -18,6 +18,16 @@ $ ./inspect --dest-dir=mydir --k8s-client=oc
 
 # override openperouter namespace
 $ ./inspect --dest-dir=mydir --namespace=myns --k8s-client=oc
+
+# via global Make target
+$ make inspect
+
+# override parameters
+$ KUBECONFIG_PATH=$KUBECONFIG \
+    make inspect \
+      NAMESPACE=myns \
+      KUBECTL=oc \
+      INSPECT_DIR=./art
 ```
 **Note:** Options must be specified with `=`.
 
--- a/tools/inspect/README.md
+++ b/tools/inspect/README.md
@@ -19,6 +19,9 @@ $ ./inspect --dest-dir=mydir --k8s-client=oc
 # override openperouter namespace
 $ ./inspect --dest-dir=mydir --namespace=myns --k8s-client=oc
 
+# collect logs since relative time duration
+$ ./inspect --since=3m
+
 # via global Make target
 $ make inspect
 
@@ -27,7 +30,8 @@ $ KUBECONFIG_PATH=$KUBECONFIG \
     make inspect \
       NAMESPACE=myns \
       KUBECTL=oc \
-      INSPECT_DIR=./art
+      INSPECT_DIR=./art \
+      SINCE=3m
 ```
 **Note:** Options must be specified with `=`.
 
@@ -37,6 +41,7 @@ $ KUBECONFIG_PATH=$KUBECONFIG \
 | `--namespace`  | OpenPERouter namespace                                            | `openperouter-system`  |
 | `--dest-dir`   | Output directory path                                             | `openperouter-inspect` |
 | `--k8s-client` | Kubernetes client                                                 | `kubectl`              |
+| `--since`      | Collect pod logs newer then relative duration (e.g.: 5s, 10m, 2h) |                        |
 | `-h`, `--help` | Print usage instructions                                          |                        |
 
 ## Output
--- a/tools/inspect/README.md
+++ b/tools/inspect/README.md
@@ -150,8 +150,8 @@ $ scp -r <target node>/openperouter-inspect-host ./<target node>-perouter-inspec
 $ docker exec pe-kind-worker -i bash <<< $(cat inspect_host)
 $ docker cp pe-kind-worker:/openperouter-inspect-host ./pe-kind-worker-inspect-host
 
-# via global make, artifacts stored at /tmp/$NODE-inspect
-$ make inspect-host NODE=pe-kind-worker
+# via global make, each node artifacts stored at /tmp/openperouter-systemd-mode-inspect/<node name>
+$ make inspect-systemd-mode
 ```
 
 ## Output
@@ -167,13 +167,22 @@ NAME                    STATUS   ROLES           AGE     VERSION
 pe-kind-control-plane   Ready    control-plane   3h20m   v1.32.2
 pe-kind-worker          Ready    <none>          3h20m   v1.32.2
 
-$ make inspect-host NODE=pe-kind-worker
+$ make inspect-systemd-mode
 
-$ tree /tmp/pe-kind-worker-inspect/
-pe-kind-worker-inspect/
-├── config_files.log
-├── root_netns_info.log
-├── router_info_podman_quadlet.log
-└── configs
-    └── node-config.yaml
+$ tree /tmp/openperouter-systemd-mode-inspect/ 
+├── pe-kind-control-plane
+│   ├── node_infoconfig_files.log
+│   ├── root_netns_info.log
+│   ├── router_info_podman_quadlet.log
+│   └── configs
+│        └── node-config.yaml
+└── pe-kind-worker
+    ├── node_infoconfig_files.log
+    ├── root_netns_info.log
+    ├── router_info_podman_quadlet.log
+    └── configs
+        └── node-config.yaml
+     
+# inspect a specific node     
+$ make inspect-systemd-mode NODES=pe-kind-worker
 ```
--- a/tools/inspect/README.md
+++ b/tools/inspect/README.md
@@ -144,6 +144,9 @@ $ scp -r <target node>/openperouter-inspect-host ./<target node>-perouter-inspec
 # troubleshooting kind cluster node running OpenPERouter on host mode
 $ docker exec pe-kind-worker -i bash <<< $(cat inspect_host)
 $ docker cp pe-kind-worker:/openperouter-inspect-host ./pe-kind-worker-inspect-host
+
+# via global make, artifacts stored at /tmp/$NODE-inspect
+$ make inspect-host NODE=pe-kind-worker
 ```
 
 ## Output
--- a/tools/inspect/README.md
+++ b/tools/inspect/README.md
@@ -150,7 +150,8 @@ $ make inspect-systemd-mode
 ```
 
 ### Output
-- `router_info_podman_quadlet.log` - router infrastructure information collected via the router Podman Quadlet container
+- `router_info_podman.log` - router infrastructure information collected from podman (for Podman Quadlet containers)
+- `router_info_crictl.log` - router infrastructure information collected using crictl (e.g.: from CRI-O, containerd)
 - `root_netns_info.log` - Root network namespace information
 - `configs/` - Contains collected static config resources in YAML form
 - `config_files.log` - Static config resources collection log
@@ -168,13 +169,13 @@ $ tree /tmp/openperouter-systemd-mode-inspect/
 ├── pe-kind-control-plane
 │   ├── config_files.log
 │   ├── root_netns_info.log
-│   ├── router_info_podman_quadlet.log
+│   ├── router_info_podman.log
 │   └── configs
 │       └── node-config.yaml
 └── pe-kind-worker
     ├── config_files.log
     ├── root_netns_info.log
-    ├── router_info_podman_quadlet.log
+    ├── router_info_podman.log
     └── configs
         └── node-config.yaml
      
--- a/tools/inspect/README.md
+++ b/tools/inspect/README.md
@@ -1,0 +1,117 @@
+# Inspect OpenPERouter deployment
+
+The `inspect` tool makes debugging OpenPERouter deployments easier by collecting related objects and logs.
+
+## Prerequisites
+- Cluster API client set for the target cluster.
+- Read access to the OpenPERouter namespace; exec access to router pods for node logs.
+
+## How to use:
+```bash
+$ ./inspect
+
+# overriding output directory path
+$ ./inspect --dest-dir=/tmp/perouter-logs
+
+# use a different Kubernetes client
+$ ./inspect --dest-dir=mydir --k8s-client=oc
+
+# override openperouter namespace
+$ ./inspect --dest-dir=mydir --namespace=myns --k8s-client=oc
+```
+**Note:** Options must be specified with `=`.
+
+### Options
+| Option         | Description                                                       | Default                |
+|----------------|-------------------------------------------------------------------|------------------------|
+| `--namespace`  | OpenPERouter namespace                                            | `openperouter-system`  |
+| `--dest-dir`   | Output directory path                                             | `openperouter-inspect` |
+| `--k8s-client` | Kubernetes client                                                 | `kubectl`              |
+| `-h`, `--help` | Print usage instructions                                          |                        |
+
+## Output
+The output root directory contains the following:
+- `timestamp` - Execution timestamp
+- `inspect.log` - Execution log
+- `node_info/` - Per node network and routing infrastructure information
+- `<openperouter namesapce>/` - OpenPERouter namespace objects and workloads logs (defaults is `openperouter-system`)
+- `<namespace name>/` - Per namespaces containing config resources directory (Underlay, L3VNI, L2VNI, etc.)
+
+The OpenPERouter namespace directory structure:
+- `overview/all.log` - Existing resources in summary
+- `pod_logs/` - Pod logs
+- `namespace.yaml` - Namespace state
+- `events.yaml` - Events
+- `<resource-name>/` - Per resource directory (CRDs, workloads)
+
+### Example:
+```bash
+$ tree /tmp/openperouter-inspect/
+├── inspect.log
+├── timestamp
+├── blue
+│   ├── l2vnis
+│   │   └── blue-111.yaml
+│   └── l3vnis
+│       └── blue-101.yaml
+├── node_info
+│   ├── pe-kind-control-plane
+│   │   ├── root_netns_info.log
+│   │   ├── router_info.log
+│   │   └── router_netns_info.log
+│   └── pe-kind-worker
+│       ├── root_netns_info.log
+│       ├── router_info.log
+│       └── router_netns_info.log
+└── openperouter-system
+    ├── events.yaml
+    ├── namespace.yaml
+    ├── configmaps
+    │   ├── frr-startup.yaml
+    │   └── kube-root-ca.crt.yaml
+    ├── daemonsets
+    │   ├── controller.yaml
+    │   └── router.yaml
+    ├── deployments
+    │   └── nodemarker.yaml
+    ├── l2vnis
+    │   └── layer2.yaml
+    ├── l3vnis
+    │   └── red.yaml
+    ├── overview
+    │   └── all.log
+    ├── pod_logs
+    │   ├── controller-cdkqz_controller.log
+    │   ├── controller-j8tgb_controller.log
+    │   ├── nodemarker-7cf554c5b8-8sq72_nodemarker.log
+    │   ├── nodemarker-7cf554c5b8-8sq72_nodemarker_previous.log
+    │   ├── router-w5d2t_frr.log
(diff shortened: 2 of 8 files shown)


tools/inspect/router_info
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash 
+
+# node router network-namespace name
+readonly NETNS="perouter"
+
+readonly NETNS_EXEC="ip netns exec $NETNS"
+readonly COMMANDS=(\
```
maiqueb: I think we should also show the forwarding DB of the cluster - e.g. `bridge fdb`

tools/inspect/inspect
```diff
@@ -0,0 +1,186 @@
...
+if ! "${CTL_BIN}" get ns "${NAMESPACE}" &> /dev/null; then
+	echo "FATAL: '${NAMESPACE}' namespace not found. Make sure KUBECONFIG is set correctly." && exit 1
+fi
+
+# artifacts sub-directories
+readonly NS_DIR="${ARTIFACTS_DIR}/${NAMESPACE}"
+readonly OVERVIEW_DIR="${NS_DIR}/overview"
+readonly RESOURCES_DIR="${NS_DIR}/resources"
+readonly POD_LOGS_DIR="${NS_DIR}/pod_logs"
+readonly NODE_LOGS_DIR="${NS_DIR}/node_logs"
```
maiqueb: I think this is not accurate ... these are not logs. 

It gather a per node view of the router's infra. Maybe `routing_infra` ? @fedepaol will surely have a better suggestion

tools/inspect/inspect
```diff
@@ -0,0 +1,186 @@
...
+
+if ! command -v "${CTL_BIN}" &> /dev/null; then
+	echo "FATAL: '${CTL_BIN}' command not found" && exit 1
+fi
+
+if ! "${CTL_BIN}" get ns "${NAMESPACE}" &> /dev/null; then
+	echo "FATAL: '${NAMESPACE}' namespace not found. Make sure KUBECONFIG is set correctly." && exit 1
+fi
+
+# artifacts sub-directories
```
maiqueb: does it make sense to also capture the NODE infra view ? i.e. the node's net ns infra (neighbors, FDB, routes, etc).

maiqueb (commented):
Some questions

fedepaol:
first high level comment, need to look at this, but we should have a knob for when running on the host

tools/inspect/inspect
ormergi: I will add more info for the root netns.

tools/inspect/inspect
ormergi: Will change to `node_info`, if no preferred alternative

ormergi:
@fedepaol thanks for looking, regarding host-mode, what would you expect to collect?

tools/inspect/router_info
ormergi: done

tools/inspect/inspect
ormergi: done

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/c4e30933a11e3cbe0d5aba1fc55802b506ab7e04..f774e2f6aed30739deabb76c0f52014d5e55ec70): 
- Extract netns info logic form `router_info` to new script: `net_info`
Now it runs the `net_info` on router pod's `reloader` container for collecting root netns info.
And on `frr` container for collecting the router infra netns info (`perouter`). $NETNS is defined to explicitly exec the commands at the desired netns. 
`router_info` runs on `frr` container, collecting the usual FRR information via `vtysh`.
- Collect `bridge fdb`
- The above info is  now stored at `node_info` dir (outside the namespace dir).
- Do best effort to collect all related info and tolerate intermediate failures (e.g.: some resource not found).
- Dump openperouter namespace state
- Dump `openoerouter` resource (managed by the project operator)
- Store execution log in artifacts dir (inspect.log)
- Make target enable overriding artifacts dir path via $INSPECT_DIR.
- Fixed formatting, typos, shellcheck issues, removed redundant vars, updated README.

maiqueb:
> @fedepaol thanks for looking, regarding host-mode, what would you expect to collect?

regarding the data plane, I'd say the exact same thing. 

for the API objects, I would say you should get the open pe router types from the config. 
I know there's a [PR](https://github.com/openperouter/openperouter/pull/469) in flight to mirror these to the API, but I think we should adjust to it when it lands.

ormergi:
@maiqueb Do you mean running the tool inside a host (e.g: via ssh) and gather opeperouter related stuff?

maiqueb:
> @maiqueb Do you mean running the tool inside a host (e.g: via ssh) and gather opeperouter related stuff?

IDK how you access systemd mode's router netns. But that's what you must do. Checking the e2e tests would be the first step. 

@fedepaol can point you in the right direction.

tools/inspect/inspect
```diff
@@ -0,0 +1,202 @@
...
+			if $ctl logs "$p" -c "$c" --previous &> /dev/null; then
+				logfile="${dir}/${p}_${c}_previous.log"	
+				$ctl logs "$p" -c "$c" --previous > "$logfile"
+				echo "Dumped previous logs for [$p/$c] to [$logfile]"
+			fi
+		done
+	done
+}
+
+function dump_node_logs() {
```
maiqueb: dump node info, right ?

maiqueb (approved):
I still think the variable name "logfile" is being abused here and there (we store the network inventory there on both the router netns and the node root netns). 

But, this is a nit. @fedepaol can propose a better name (or you can).
Tks for this.

tools/inspect/inspect
ormergi: Thanks, fixed

ormergi:
Added new script for troubleshooting individual node for when openperouter runs on host mode.
`tools/inspect/inspect_host`.

FWIW initially I tried to extend the existing `inspect` script but it cannot work because when host-mode is on, we cannot collect related info via k8s API. Info must be collected directed from the target node (e.g.: via ssh).
Thus the new script is meant to run directly on target node.

tools/inspect/inspect
```diff
@@ -0,0 +1,202 @@
...
+			echo "# $r"
+			$ctl get "$r" -o wide
+			echo "" 
+		done
+	} > "$logfile"
+
+	echo "Dumped namespace resources overview to [$logfile]"
+}
+
+function dump_resources() {
```
RamLavi: nit: this function handles only namespaced resources. consider renaming the function accordingly. suggestion: `dump_namespaced_resources`

tools/inspect/inspect
```diff
@@ -0,0 +1,202 @@
...
+	done
+}
+
+function dump_pod_logs() {
+	local -r ctl="$1"
+	local -r dir="$2"
+
+	mkdir -p "$dir"
+
+	local -r pods="$($ctl get pods -o jsonpath='{.items[*].metadata.name}')"
```
RamLavi, line 127: don't you want namespaced? or do you intentially only want default namespace

tools/inspect/inspect_host
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash 
+
+# SPDX-License-Identifier:Apache-2.0
+
```
RamLavi, line 4: why no `set -euo pipefail` on the host script?

RamLavi (commented):
Hey, first pass

tools/inspect/inspect
maiqueb: it is namespaced; CTL is initialized to `readonly CTL="${CTL_BIN} -n ${NAMESPACE}"` , and is what's passed into all of these helpers. 

I wonder why name it `ctl` though.

RamLavi: ah, thanks didn't notice that, thanks. I would keep CTL with out the `-n ${NAMESPACE}` but I suppose that's a style nit.

ormergi: The caller pass the client string with the namespace embaded, i.e.: `kubectl -n openperouter`, so it will find in openperouter namesapce pods only.

ormergi: It seems I commented on cached version, didn't saw previous comments.

tools/inspect/inspect_host
ormergi: Unlike the other script this one is straight forward and has less prior steps that could fail
I removed -e to have it  tolerate failures and collecting whatever possible,
Looking at this again, adding those could help improve stability.

tools/inspect/inspect
ormergi: `ctl` as in kubectl, its stand for 'control' for short, it used in other common tools such as sysctl, systemctl.

tools/inspect/inspect
```diff
@@ -0,0 +1,202 @@
...
+readonly RESOURCES=(\
+	# CRDs
+	"underlays" \
+	"l2vnis" \
+	"l3vnis" \
+	"l3passthroughs" \
+	"rawfrrconfigs" \
+	"routernodeconfigurationstatuses" \
+	"openperouters" \
+	# deployment resources
```
qinqon: btw we have teh "all" meta resource to gather all the core resouces, maybe we can use that so we don't have to harcoded them here ? 

```bash
$ kubectl get all -n ovn-kubernetes
Warning: kubevirt.io/v1 VirtualMachineInstancePresets is now deprecated and will be removed in v2.
NAME                                        READY   STATUS    RESTARTS        AGE
pod/ovnkube-control-plane-7445b4857-wpw77   1/1     Running   1 (3d18h ago)   3d18h
pod/ovnkube-identity-245vt                  1/1     Running   1 (3h24m ago)   3d18h
pod/ovnkube-node-2brcw                      6/6     Running   0               3d18h
pod/ovnkube-node-8r6b6                      6/6     Running   0               3d18h
pod/ovnkube-node-gcz8b                      6/6     Running   0               3d18h
pod/ovnkube-node-lslvp                      6/6     Running   0               3d18h
pod/ovs-node-7wkq6                          1/1     Running   0               3d18h
pod/ovs-node-8npm7                          1/1     Running   0               3d18h
pod/ovs-node-9s2hw                          1/1     Running   0               3d18h
pod/ovs-node-qtm4j                          1/1     Running   0               3d18h

NAME                              DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR                                                   AGE
daemonset.apps/ovnkube-identity   1         1         1       1            1           kubernetes.io/os=linux,node-role.kubernetes.io/control-plane=   3d18h
daemonset.apps/ovnkube-node       4         4         4       4            4           kubernetes.io/os=linux                                          3d18h
daemonset.apps/ovs-node           4         4         4       4            4           kubernetes.io/os=linux                                          3d18h

NAME                                    READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/ovnkube-control-plane   1/1     1            1           3d18h

NAME                                              DESIRED   CURRENT   READY   AGE
replicaset.apps/ovnkube-control-plane-7445b4857   1         1         1       3d18h

```

It can be used with -o yaml

tools/inspect/inspect
maiqueb: fair enough.

tools/inspect/inspect
```diff
@@ -0,0 +1,202 @@
...
+readonly NETNS="perouter"
+# resources to collect
+readonly RESOURCES=(\
+	# CRDs
+	"underlays" \
+	"l2vnis" \
+	"l3vnis" \
+	"l3passthroughs" \
+	"rawfrrconfigs" \
+	"routernodeconfigurationstatuses" \
```
qinqon: These can be created by users at different namespaces, I think the inspector need to dump these from all the namespaces, so better let's create a different group from there since these are not kubernetes core ones.

tools/inspect/inspect
```diff
@@ -0,0 +1,202 @@
...
+    cat <<EOF
+Inspect OpenPERouter deployment and export related resources and logs.
+
+Usage: inspect [options]=<value>
+
+Example
+  inspect --dest-dir=/tmp/openperouter-logs --k8s-client=./bin/kubectl --namespace=openperouter-system
+
+Options:
+  --namespace   The namespace to inspect (default: "openperouter-system")
```
qinqon: This should be the `--operator-namespace` the CRDs like L2VNI can be on random namespaces.

qinqon (changes requested):
openperouter CRDs like L2VNI can live on other namespace, we should prepare this for that.

tools/inspect/inspect
ormergi: The intention is to have overview file that enable quick lookup at the object state related to openperouter.

`all` provides basic info such as pods,ds,deply and replicaset.
The output would include those plus additional  openperouter related objects.

It was easier to use the global RESOURCES then use all and another set of object names.

BTW the object YAMLs are dumped anyways by different functions.

tools/inspect/inspect
ormergi: Renamed to descriptive name as suggested

tools/inspect/inspect_host
ormergi: Done

tools/inspect/inspect
ormergi: Done
The script now collect the above types from all namepaces.
Minus `routernodeconfigurationstatuses`, because for now it generated in openperouter namespace only.

tools/inspect/inspect
ormergi: Now the script collect info from all namespaces
The arg remained as is pointing to opneperouter system namespace, I updated the flag description.

ormergi:
[Changes: ](https://github.com/openperouter/openperouter/compare/bdd34a2827c25bf2e3bc0d1bf9889e98d0af96d5..c6918ca4664b967304e7686f4c7c875e8b5b6d95): resolved https://github.com/openperouter/openperouter/pull/484#pullrequestreview-4541915438 and https://github.com/openperouter/openperouter/pull/484#pullrequestreview-4543386300. @RamLavi @qinqon PTAL:
- `inspect` now collect config API resources from all namesapces and not just openperouter namespace.
- `inspect_host` now set some bash flags for better stability
- Add make target for inspecting host-mode nodes, updated README.

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/c6918ca4664b967304e7686f4c7c875e8b5b6d95..b7b9e82c29f27ca77d417d0d92bfa5182b9cc843): fix formatting, missing license, rm dead code, fix bugs. 
Output dir structure slightly change, each namespace has its own dir with per resource dir containing yaml and related info.

fedepaol:
I'll defer the review / approval to @maiqueb .
Only a couple of comments:

- this should work for systemd mode too, even when the api server is not available yet (so we must scrape the config)
- we should try to use this as the base for our test reports instead of the k8s reporter we have now. This probably means we need to add a way to grab only the last 10 mins of logs as we do in the test library

This will ensure we can rely on it and will force us to extend it whenever some information is missing.

maiqueb:
> I'll defer the review / approval to @maiqueb . Only a couple of comments:
> 
>     * this should work for systemd mode too, even when the api server is not available yet (so we must scrape the config)
> 
>     * we should try to use this as the base for our test reports instead of the k8s reporter we have now. This probably means we need to add a way to grab only the last 10 mins of logs as we do in the test library
> 
> 
> This will ensure we can rely on it and will force us to extend it whenever some information is missing.

Would it be OK if we did that as follow-ups ? We'd open up two issues, and work on them as our capacity allows.

fedepaol:
> > I'll defer the review / approval to @maiqueb . Only a couple of comments:
> > ```
> > * this should work for systemd mode too, even when the api server is not available yet (so we must scrape the config)
> > 
> > * we should try to use this as the base for our test reports instead of the k8s reporter we have now. This probably means we need to add a way to grab only the last 10 mins of logs as we do in the test library
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will ensure we can rely on it and will force us to extend it whenever some information is missing.
> 
> Would it be OK if we did that as follow-ups ? We'd open up two issues, and work on them as our capacity allows.

for systemd mode I'd rather have it from the get go as we should treat it as part of the whole project.
The other one is in the interest of the tool itself

qinqon (approved):
LGTM

maiqueb:
> > > I'll defer the review / approval to @maiqueb . Only a couple of comments:
> > > ```
> > > * this should work for systemd mode too, even when the api server is not available yet (so we must scrape the config)
> > > 
> > > * we should try to use this as the base for our test reports instead of the k8s reporter we have now. This probably means we need to add a way to grab only the last 10 mins of logs as we do in the test library
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This will ensure we can rely on it and will force us to extend it whenever some information is missing.
> > 
> > 
> > Would it be OK if we did that as follow-ups ? We'd open up two issues, and work on them as our capacity allows.
> 
> for systemd mode I'd rather have it from the get go as we should treat it as part of the whole project. The other one is in the interest of the tool itself

After speaking with @ormergi he showed me this already features the implementation for systemd mode: 
https://github.com/openperouter/openperouter/pull/484/changes#diff-62245ffbcfe9b7c3e5797e417b429e88f8522f4428ab459a0e4af1f38d0044f9R116

Could you tell us if this makes sense from your perspective ?

maiqueb:
> > > > I'll defer the review / approval to @maiqueb . Only a couple of comments:
> > > > ```
> > > > * this should work for systemd mode too, even when the api server is not available yet (so we must scrape the config)
> > > > 
> > > > * we should try to use this as the base for our test reports instead of the k8s reporter we have now. This probably means we need to add a way to grab only the last 10 mins of logs as we do in the test library
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > This will ensure we can rely on it and will force us to extend it whenever some information is missing.
> > > 
> > > 
> > > Would it be OK if we did that as follow-ups ? We'd open up two issues, and work on them as our capacity allows.
> > 
> > 
> > for systemd mode I'd rather have it from the get go as we should treat it as part of the whole project. The other one is in the interest of the tool itself
> 
> After speaking with @ormergi he showed me this already features the implementation for systemd mode: https://github.com/openperouter/openperouter/pull/484/changes#diff-62245ffbcfe9b7c3e5797e417b429e88f8522f4428ab459a0e4af1f38d0044f9R116
> 
> Could you tell us if this makes sense from your perspective ?

@fedepaol ^

ormergi:
@fedepaol regarding systemd mode, I added support for it while ago, please see https://github.com/ormergi/openperouter/blob/3ebae865a020bc2ba81ee249bdee91d46c8b8436/tools/inspect/README.md#inspect-openperouter-nodes-when-running-on-systemd-mode

ormergi:
[changes](https://github.com/openperouter/openperouter/compare/d407418640ca74f70b9b8fa3a203cd21dab04c7d..eb366e7b5fed14954cbc163bfb1ad566c6d2f2aa): Rebase & fix commitlint issues

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/0a0ef3c7748ad3e56a3f16edef7312dcf85eb2a6..a64a35a7842e358c3f836b8d471ffd8bd7feed2b): collect init container logs as well

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/0a0ef3c7748ad3e56a3f16edef7312dcf85eb2a6..a64a35a7842e358c3f836b8d471ffd8bd7feed2b): enable collecting logs since  relative time duration, partially resolve https://github.com/openperouter/openperouter/pull/484#issuecomment-4841410296

ormergi:
@maiqueb PTAL

tools/inspect/net_info
```diff
@@ -0,0 +1,28 @@
...
+if [ -n "$NETNS" ]; then
+	args="ip netns exec $NETNS"
+fi
+
+readonly NETNS_INFO=(\
+"ip link" \
+"ip address" \
+"ip route" \
+"ip neigh" \
+"ip vrf" \
```
fedepaol, line 10: we should iterate over all the vrfs and dump routes and addresses for each vrf

tools/inspect/router_info
```diff
@@ -0,0 +1,20 @@
...
+
+# SPDX-License-Identifier:Apache-2.0
+
+readonly FRR_INFO=(\
+"timeout 10s vtysh -c 'show running-config'" \
+"timeout 10s vtysh -c 'show bgp ipv4'" \
+"timeout 10s vtysh -c 'show bgp ipv6'" \
+"timeout 10s vtysh -c 'show bgp neighbor'" \
+"timeout 10s vtysh -c 'show bfd peer'" \
+"timeout 10s vtysh -c 'show bgp l2vpn evpn'" \
```
fedepaol: show received and advertised routes (ipv4, ipv6 and evpn for each neighbor)

https://docs.frrouting.org/en/latest/bgp.html#clicmd-show-bgp-afi-safi-neighbor-PEER-routes-advertised-routes-received-routes-A.B.C.D-M-X-X-X-X-M-detail-json

tools/inspect/net_info
ormergi: Done

tools/inspect/router_info
ormergi: Done

ormergi:
[Changes](https://github.com/openperouter/openperouter/compare/604dbd45d82091df9f98fa303c5446f9942e1e05..a8413efade00061e881cb1471a2213dd8f7005ba): resolve https://github.com/openperouter/openperouter/pull/484#pullrequestreview-4631184112, @fedepaol PTAL :pray: 
- Collect per VRF routes and addresses 
- Collect per neighbor received and advertised routes 
- Wire makefile $NAMESPACE to the new `inspect` target
- Fix README formatting, add example for overriding namespace parameter and setting KUBECONFIG

tools/inspect/README.md
```diff
@@ -0,0 +1,179 @@
+# Inspect OpenPERouter deployment
```
fedepaol, line 1: We should add this to the website imo, not here. Going to comment to reduce teh content a bit

tools/inspect/README.md
```diff
@@ -0,0 +1,179 @@
...
+$ KUBECONFIG_PATH=$KUBECONFIG \
+    make inspect \
+      NAMESPACE=myns \
+      KUBECTL=oc \
+      INSPECT_DIR=./art \
+      SINCE=3m
+```
+**Note:** Options must be specified with `=`.
+
+### Options
```
fedepaol: I'd rather have this table and not all the list of examples above

tools/inspect/README.md
```diff
@@ -0,0 +1,179 @@
...
+- `<namespace name>/` - Per namespaces containing config resources directory (Underlay, L3VNI, L2VNI, etc.)
+
+The OpenPERouter namespace directory structure:
+- `overview/all.log` - Existing resources in summary
+- `pod_logs/` - Pod logs
+- `namespace.yaml` - Namespace state
+- `events.yaml` - Events
+- `<resource-name>/` - Per resource directory (CRDs, workloads)
+
+### Example:
```
fedepaol, line 57: I'd remove the example alltogehter (or we might consider leaving this here and a reduced variant under the website)

tools/inspect/README.md
```diff
@@ -0,0 +1,179 @@
...
+```bash
+$ tree /tmp/openperouter-inspect/
+├── inspect.log
+├── timestamp
+├── blue
```
fedepaol: what's blue here?

Makefile
```diff
@@ -664,3 +664,16 @@ deploy-olm: operator-sdk ## deploys OLM on the cluster
...
 
+INSPECT_DIR ?= /tmp/openperouter-inspect
+.PHONY: inspect
+inspect:
+	tools/inspect/inspect --k8s-client=$(KUBECTL) --namespace=$(NAMESPACE) --dest-dir=$(INSPECT_DIR) --since=$(SINCE)
+
+HOST_INSPECT_DIR = /openperouter-inspect-host
+.PHONY: inspect-host
+inspect-host:
+	@ ! test -n "$(NODE)" && echo "target node is not set, make sure to pass NODE argument" && exit 1; \
```
fedepaol: we should iterate over the nodes

tools/inspect/net_info
```diff
@@ -0,0 +1,63 @@
...
+  local -r netns_name=$1
+  local -r args="$2"
+
+  vrfs=("$($args $VRF_SHOW | awk '{print $1}')")
+  [[ ${#vrfs} -eq 0 ]] && return
+
+  echo "### netns [$netns_name] VRFs:"
+  $args $VRF_SHOW
+
+  for vrf in "${vrfs[@]}"; do
```
fedepaol, line 30: sorry I didn't realize before, but I think we should do this only on the namespace side. For the host netns this is going to be distracting

tools/inspect/net_info
```diff
@@ -0,0 +1,63 @@
...
+for cmd in "${NETNS_INFO[@]}"; do
+	c="$args $cmd"
+	echo "$ $c"
+	$c
+	echo ""
+done
+echo "###"
+echo ""
+
+if [[ -n $NETNS ]]; then
```
fedepaol: this if is already driving args, consider squashing one of them. Also, args is misleading.
I would propagate NETNS in print_vrf_info , having it drive setting cmd_prefix or something like that

fedepaol:
some minor nits and it's good to go

maiqueb:
@ormergi please provide docs for this feature as well.

Makefile
ormergi: The intention of this target is to inspect single node?
Do you mean there should be another target for inspecting all nodes?

tools/inspect/README.md
ormergi: So this README should stay? 
Asking because https://github.com/openperouter/openperouter/pull/484/changes/BASE..a8413efade00061e881cb1471a2213dd8f7005ba#r3528099751

tools/inspect/README.md
ormergi: A namespace called "blue", containing the below l2/l3vnis.

tools/inspect/README.md
ormergi: The tool is not release or something like that meaning we will need to refer users to the project repo to use it, are we good with that?

fedepaol: fine with me, but we should make sure we refer at the proper tag in the website, and it's considered in the make cutrelease automation (just check that when we call that, the entry in the website gets the proper tag).

Makefile
fedepaol: I would only have one target with all nodes symmetric to the rule we have for pods

tools/inspect/README.md
fedepaol: but we won't have that in openperouter right? Or, what's the case where we might have that?
I am probably missing something. Is that for the leaves?

ormergi: Right, this is for demonstrating how configs are collected when they are created in different namespaces then OpenPEROuter system namespace.
In this case we have namesapce called "blue" and it has l2/l3 VNI objects. Inspect will create directory with for "blue" namespace objects

Makefile
ormergi: OK, I"ll fix that.
BTW, the other target will also collect node info such as network and routing infra info (frr, via vtysh).  
This target is for when openperouter runs on systemd mode. I should've reflected that in the target name.

tools/inspect/README.md
ormergi: I left the examples and reduced them for less noise, and have the website doc refer to the tool README in the repository.

tools/inspect/README.md
ormergi: I updated the example to make it easier to understand.

ormergi:
[push1](https://github.com/openperouter/openperouter/compare/a8413efade00061e881cb1471a2213dd8f7005ba..8f0dac9d3df918f4034a6178587cf1b978e808df): rebase

[push N](https://github.com/openperouter/openperouter/compare/56494663b6c659521495762a0756e0065147cfe3..ee79632fb2d0e7edd95126a28677eb9be4782b50): resolve review comments https://github.com/openperouter/openperouter/pull/484#pullrequestreview-4631184112 https://github.com/openperouter/openperouter/pull/484#issuecomment-4905454431: 
- Add website doc for inspect tools, refile README to act as source of truth containing additional details and examples
- Change make target (previously inspect_host) to collect info from all nodes
- Do not dump root netns each VRF info
- Support inspecting systemd mode running router container using crio (info collected using crictl)

@fedepaol @maiqueb thanks for the review, PTAL

tools/inspect/README.md
fedepaol: openperouter only listen for resources on its namespace. I'd get rid of this alltogether

fedepaol:
one nit, then lgtm

tools/inspect/net_info
ormergi: Done

tools/inspect/net_info
ormergi: Done

Makefile
ormergi: Done

tools/inspect/README.md
```diff
@@ -0,0 +1,189 @@
...
+│   ├── serviceaccounts
+│   │   ├── controller.yaml
+│   │   ├── default.yaml
+│   │   └── perouter.yaml
+│   ├── services
+│   │   └── openpe-webhook-service.yaml
+│   └── underlays
+│      └── underlay.yaml
+└── other-namespace
+    ├── l2vnis
```
fedepaol, line 82: openpe listens only to resources on its namespace, so this is not relevant

ormergi: Done

tools/inspect/README.md
ormergi: Done

PR #505, opened 2026-06-17
fix(validation): remove ValidateRouteTargets from validateL2VNI

### LINKED ISSUE #475: Remove unnecessary `ValidateRouteTargets` call from `ValidateL2VNIs`
L2VNIs do not have route targets, but `ValidateL2VNIs` calls `ValidateRouteTargets`. This was inherited from the shared `validateVNIs` function that validated both L3 and L2 VNIs together.

The call is harmless (it validates empty route targets, which always passes) but unnecessary and misleading.

RamLavi:
L2VNIs do not have route targets (exportRTs and importRTs are always nil), so the ValidateRouteTargets call was a harmless no-op. Remove it to avoid confusion.

Closes: https://github.com/openperouter/openperouter/issues/475

/kind cleanup

Remove the unnecessary `ValidateRouteTargets` call from `validateL2VNI`. L2VNIs do not have route targets (`exportRTs` and `importRTs` are always nil in `vniFromL2VNI`), so the call was a harmless no-op but misleading.

None.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/conversion/validate_vni.go
+++ b/internal/conversion/validate_vni.go
@@ -156,16 +156,13 @@ func FilterUniqueVNIs(l3Vnis []v1alpha1.L3VNI, l2Vnis []v1alpha1.L2VNI) ([]v1alp
 	return validL3, validL2, errors.Join(allErrors...)
 }
 
-// validateL2VNI validates a single L2VNI's fields (VRF name, route targets, HostMaster, L2GatewayIPs).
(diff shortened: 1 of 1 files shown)


RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/dfade37dc072f0598a83e648b0ca0db6b503eaed..01a8383aee9e54c5f4efd937823aedd147b09e96): fix commit lint

maiqueb (approved):
tks

PR #527, opened 2026-06-24
Add unit tests for validateRouteTargets and fix typo

andreaskaris:

/kind cleanup

Add unit tests for validateRouteTargets and fix typo

N/A

**Release note**:

```release-note
Add unit tests for validateRouteTargets and fix typo
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved route-target validation by distinguishing true ASN prefixes from malformed IPv4-like values.
  - When an IPv4-like pattern is detected in the ASN portion, validation now returns an error that requires `A.B.C.D:MN` with a valid IPv4 address.
  - Refined error messages for invalid route-target formats, including missing separators, empty values, multiple colons, and numeric overflow/limit violations.
- **Tests**
  - Expanded table-driven unit tests for route-target validation, covering valid `ASN:MN` and `IPv4Address:MN` inputs plus many invalid cases with exact error assertions.

### DIFF
--- a/internal/conversion/validate_vni.go
+++ b/internal/conversion/validate_vni.go
@@ -24,10 +24,14 @@ import (
 	"github.com/openperouter/openperouter/internal/ipfamily"
 )
 
-var interfaceNameRegexp *regexp.Regexp
+var (
+	interfaceNameRegexp *regexp.Regexp
+	ipv4LikeRegexp      *regexp.Regexp
+)
 
 func init() {
 	interfaceNameRegexp = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9._-]*$`)
+	ipv4LikeRegexp = regexp.MustCompile(`^\d+\.\d+\.\d+\.\d+$`)
 }
 
 // FilterValidL3VNIs validates L3VNIs per-field and returns the valid resources
@@ -630,9 +634,14 @@ func validateRouteTarget(rt string) error {
 		return nil
 	}
 
+	// Catch values that look like an IPv4 address but failed validation (e.g. 999.1.2.3).
+	if ipv4LikeRegexp.MatchString(rtParam[0]) {
+		return fmt.Errorf("RT format must have A.B.C.D:MN where A.B.C.D is a valid IPv4 address: %s", rt)
+	}
+
 	asn, err := strconv.ParseUint(rtParam[0], 10, 32)
 	if err != nil {
-		return fmt.Errorf("RT format must have ASN:MN %s", rt)
+		return fmt.Errorf("RT format must have ASN:MN: %s", rt)
 	}
 
 	memberNumber, err := parseMemberNumber(rtParam[1])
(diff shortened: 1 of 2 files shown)


internal/conversion/validate_vni_test.go
```diff
@@ -716,3 +716,104 @@ func TestHasSubnetOverlap(t *testing.T) {
...
+			wantErrString: "RT format with 2-byte ASN must have ASN:MN where MN <= 4294967295: 65535:4294967296",
+		},
+		{
+			name:        "IP route target with max member number",
+			routeTarget: "10.0.0.1:65535",
+		},
+		{
+			name:        "4-byte ASN with max member number",
+			routeTarget: "100000:65535",
+		},
```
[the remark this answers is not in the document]
andreaskaris: good suggestion, done; I also threw in a check for strings.Contains("."), in that case we can display an error message geared towards the IPv4 address

internal/conversion/validate_vni.go
```diff
@@ -630,9 +630,13 @@ func validateRouteTarget(rt string) error {
 		return nil
 	}
 
+	if strings.Contains(rtParam[0], ".") {
```
fedepaol: soryr but I don't get this. It means the part before the : can't contain a . while we are asking for a format with a .

andreaskaris: at this stage, we know that rtParam[0] is not an IP address (see line 622). But the user used a "." in the input, so they likely wanted to enter an IP address. For example, 999.999.999.999.

Test:
```
		{
			name:          "invalid IP route target invalid IP address",
			routeTarget:   "999.999.999.999:100",
			wantErrString: "RT format must have A.B.C.D:MN where A.B.C.D is a valid IPv4 address: 999.999.999.999:100",
		},
```

Otherwise, with 999.999.999.999, we'd hit the ParseUint and we'd get:
```
return fmt.Errorf("RT format must have ASN:MN: %s", rt)
```

andreaskaris: maybe better to test for a full IPv4 regex here?

fedepaol: I don't understand this check alltogether then :) . If we are excluding ips already, we should just fail in the parseuint below

andreaskaris: we do, but the error message is currently: `RT format must have ASN:MN: %s`

Let's assume I put in `"999.999.999.999:100"`, it'd be better to get `RT format must have A.B.C.D:MN where A.B.C.D is a valid IPv4 address: 999.999.999.999:100` instead of `RT format must have ASN:MN: %s`

That's what that check is for.

See https://github.com/openperouter/openperouter/pull/527#discussion_r3467065123

fedepaol: seems arbitrary, you are just speculating that if it contains a . the user intention was to put an ip in there.
I'd just change the error message to include both formats and that's it.

andreaskaris: check if the new version (using a regex for x.x.x.x) is better. the function has error messages for either ASN:MN or IP:MN.

if you still think it's arbitrary, I'll just remove that line and modify the last unit test, I don't care and yes it's kind of arbitrary

fedepaol:
Thanks!

Issue #428, opened 2026-05-19
flaky test: Hybrid mode: static files and API configuration works with API-only config, adds static file config, then removes it   (systemdmode suite)


maiqueb:
The test in the title flakes on the `systemdmode` suite.

## Problem
The FileWatcher (fsnotify/inotify) in internal/filewatcher/filewatcher.go does not detect file changes when the watched directory is accessed through different bind mount paths in nested container environments.

## Evidence

From CI run [25934886368](https://github.com/openperouter/openperouter/actions/runs/26047408700/job/76575942295):
```
  # File watcher starts, never fires:
  May 15 18:48:23 controller: "file watcher started" directory=/etc/openperouter/configs
  # (no "static file created/modified" log entries ever appear)

  # Controller reconciles at 18:50:14, then goes silent for 5 minutes:
  May 15 18:50:14 controller: "end reconcile" request=openperouter-system/underlay
  # (no reconcile until 18:55:26 — after the test already timed out)
```

PR #542, opened 2026-07-02
fix(status): prevent reconcile storm from LastTransitionTime mismatch

RamLavi:
/kind bug

`reconcileNodeStatus` patches the status on every reconcile because `buildStatus()` creates conditions with a fresh `LastTransitionTime`. `DeepEqual` always returns false against the server-side value, triggering a patch that fires the `RouterNodeConfigurationStatus` watch and re-enqueues -- creating a reconcile storm (85 patches in 30s observed on a live cluster).

Fix: apply desired conditions via `SetStatusCondition` on a `DeepCopy` of the existing status. `SetStatusCondition` preserves `LastTransitionTime` when the condition `Status` is unchanged, so `DeepEqual` correctly returns true and no unnecessary patch is emitted.

Also makes `HostConfigurator` injectable on `PERouterReconciler` for testability, adds unit tests for the status update path (Ready, Degraded, no-storm, transitions), and adds e2e status condition checks to the node status test.

JIRA: [CNV-91855](https://redhat.atlassian.net/browse/CNV-91855)

**Release note**:

```release-note
Fix reconcile storm caused by status being re-patched on every reconcile due to LastTransitionTime mismatch.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved router node status reconciliation to preserve existing condition history and avoid unnecessary status updates.
  * Node status now more reliably transitions between **Ready** and **Degraded** based on reconcile outcomes (success, failure, then recovery).

* **Tests**
  * Enhanced end-to-end node status checks with explicit “all nodes ready” verification before and after status recreation.
  * Added controller-level coverage for idempotency, correct condition transitions, and recovery from failed reconciles.

### DIFF
--- a/e2etests/pkg/openperouter/status.go
+++ b/e2etests/pkg/openperouter/status.go
@@ -4,8 +4,14 @@ package openperouter
 
 import (
 	"context"
+	"fmt"
+	"time"
 
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 	"github.com/openperouter/openperouter/api/v1alpha1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -19,4 +25,26 @@ func GetNodeStatus(cli client.Client, nodeName string) (*v1alpha1.RouterNodeConf
 		return nil, err
 	}
 	return status, nil
+}
+
+// AssertNodesStatusReady verifies that all RouterNodeConfigurationStatus CRs
+// have Ready=True and Degraded=False conditions set.
+func AssertNodesStatusReady(cli client.Client) {
+	GinkgoHelper()
+	Eventually(func(g Gomega) {
+		statusList := &v1alpha1.RouterNodeConfigurationStatusList{}
+		g.Expect(cli.List(context.Background(), statusList, &client.ListOptions{Namespace: Namespace})).To(Succeed())
+		g.Expect(statusList.Items).NotTo(BeEmpty(), "should have at least one node status")
+
+		for _, nodeStatus := range statusList.Items {
+			g.Expect(nodeStatus.Status).NotTo(BeNil(),
+				fmt.Sprintf("node-status %q should have status set", nodeStatus.Name))
+			g.Expect(apimeta.IsStatusConditionPresentAndEqual(nodeStatus.Status.Conditions, v1alpha1.ConditionTypeReady, metav1.ConditionTrue)).
+				To(BeTrue(), fmt.Sprintf("node-status %q Ready should be True", nodeStatus.Name))
+			g.Expect(apimeta.IsStatusConditionPresentAndEqual(nodeStatus.Status.Conditions, v1alpha1.ConditionTypeDegraded, metav1.ConditionFalse)).
+				To(BeTrue(), fmt.Sprintf("node-status %q Degraded should be False", nodeStatus.Name))
+			g.Expect(nodeStatus.Status.FailedResources).To(BeEmpty(),
+				fmt.Sprintf("node-status %q should have no failed resources", nodeStatus.Name))
+		}
+	}).WithTimeout(30 * time.Second).WithPolling(2 * time.Second).Should(Succeed())
 }
--- a/internal/controller/routerconfiguration/node_status.go
+++ b/internal/controller/routerconfiguration/node_status.go
@@ -39,7 +39,7 @@ func (r *PERouterReconciler) reconcileNodeStatus(ctx context.Context, reconcileE
 		return fmt.Errorf("failed to create or update node status: %w", err)
 	}
 
-	newStatus := buildStatus(reconcileErr)
+	newStatus := buildStatus(reconcileErr, nodeStatus.Status)
 
 	if equality.Semantic.DeepEqual(nodeStatus.Status, &newStatus) {
 		return nil
--- a/internal/controller/routerconfiguration/node_status_test.go
+++ b/internal/controller/routerconfiguration/node_status_test.go
@@ -1,0 +1,225 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package routerconfiguration
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/openperouter/openperouter/api/v1alpha1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	"github.com/openperouter/openperouter/internal/conversion"
+)
+
+type failingDatapathConfigurator struct{}
+
+func (f *failingDatapathConfigurator) Configure(_ context.Context, _ interfacesConfiguration) error {
+	return fmt.Errorf("failed to setup underlay: link not found")
+}
+
+func (f *failingDatapathConfigurator) Validate(_ conversion.APIConfigData) error {
+	return nil
+}
+
+func TestReconcileUpdatesNodeStatus(t *testing.T) {
+	t.Run("sets Ready status on successful reconcile", func(t *testing.T) {
+		r := newTestPERouterReconciler(t, &noopDatapathConfigurator{}, testNode())
+
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err != nil {
+			t.Fatalf("Reconcile() returned error: %v", err)
+		}
+
+		assertStatusReady(t, r.Client)
+	})
+
+	t.Run("sets Degraded status on reconcile error", func(t *testing.T) {
+		r := newTestPERouterReconciler(t, &failingDatapathConfigurator{}, testNode())
+
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err == nil {
+			t.Fatal("expected Reconcile() to return error")
+		}
+
+		assertStatusDegraded(t, r.Client)
+	})
+
+	t.Run("no status patch on repeated successful reconcile", func(t *testing.T) {
+		r := newTestPERouterReconciler(t, &noopDatapathConfigurator{}, testNode())
+
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err != nil {
+			t.Fatalf("first Reconcile() returned error: %v", err)
+		}
+		first := getNodeStatus(t, r.Client)
+		firstRV := first.ResourceVersion
+		firstTransition := apimeta.FindStatusCondition(first.Status.Conditions, v1alpha1.ConditionTypeReady).LastTransitionTime
+
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err != nil {
+			t.Fatalf("second Reconcile() returned error: %v", err)
+		}
+		second := getNodeStatus(t, r.Client)
+
+		if second.ResourceVersion != firstRV {
+			t.Errorf("ResourceVersion changed from %s to %s; expected no status patch (reconcile storm)", firstRV, second.ResourceVersion)
+		}
+		secondTransition := apimeta.FindStatusCondition(second.Status.Conditions, v1alpha1.ConditionTypeReady).LastTransitionTime
+		if !secondTransition.Equal(&firstTransition) {
+			t.Errorf("LastTransitionTime changed from %v to %v; expected preserved", firstTransition, secondTransition)
+		}
+	})
+
+	t.Run("no status patch on repeated failed reconcile", func(t *testing.T) {
+		r := newTestPERouterReconciler(t, &failingDatapathConfigurator{}, testNode())
+
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err == nil {
+			t.Fatal("expected Reconcile() to return error")
+		}
+		first := getNodeStatus(t, r.Client)
+		firstRV := first.ResourceVersion
+		firstTransition := apimeta.FindStatusCondition(first.Status.Conditions, v1alpha1.ConditionTypeReady).LastTransitionTime
+
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err == nil {
+			t.Fatal("expected Reconcile() to return error")
+		}
+		second := getNodeStatus(t, r.Client)
+
+		if second.ResourceVersion != firstRV {
+			t.Errorf("ResourceVersion changed from %s to %s; expected no status patch (reconcile storm)", firstRV, second.ResourceVersion)
+		}
+		secondTransition := apimeta.FindStatusCondition(second.Status.Conditions, v1alpha1.ConditionTypeReady).LastTransitionTime
+		if !secondTransition.Equal(&firstTransition) {
+			t.Errorf("LastTransitionTime changed from %v to %v; expected preserved", firstTransition, secondTransition)
+		}
+	})
+
+	t.Run("transitions from Degraded to Ready", func(t *testing.T) {
+		r := newTestPERouterReconciler(t, &failingDatapathConfigurator{}, testNode())
+
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err == nil {
+			t.Fatal("expected Reconcile() to return error")
+		}
+		assertStatusDegraded(t, r.Client)
+
+		r.DatapathConfigurator = &noopDatapathConfigurator{}
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err != nil {
+			t.Fatalf("Reconcile() returned error: %v", err)
+		}
+		assertStatusReady(t, r.Client)
+	})
+}
+
+type mockRouterProvider struct{}
+
+func (m *mockRouterProvider) New(_ context.Context) (Router, error) {
+	return &mockRouter{}, nil
+}
+
+func (m *mockRouterProvider) NodeIndex(_ context.Context) (int, error) {
+	return 0, nil
+}
+
+type mockRouter struct{}
+
+func (m *mockRouter) TargetNS(_ context.Context) (string, error)        { return "", nil }
+func (m *mockRouter) CanReconcile() (bool, error)                       { return true, nil }
+func (m *mockRouter) HandleNonRecoverableError(_ context.Context) error { return nil }
+
+func newTestPERouterReconciler(t *testing.T, datapathConfigurator DatapathConfigurator, objects ...client.Object) *PERouterReconciler {
+	t.Helper()
+	s := testScheme(t)
+	cli := fake.NewClientBuilder().
+		WithScheme(s).
+		WithObjects(objects...).
+		WithStatusSubresource(&v1alpha1.RouterNodeConfigurationStatus{}).
+		Build()
+	notMirrored, err := labels.NewRequirement(StaticSourceLabel, selection.DoesNotExist, nil)
+	if err != nil {
+		t.Fatalf("failed to build label requirement: %v", err)
+	}
+	frrConfigPath := filepath.Join(t.TempDir(), "frr.conf")
+	if err := os.WriteFile(frrConfigPath, nil, 0600); err != nil {
+		t.Fatalf("failed to create frr config: %v", err)
+	}
+	return &PERouterReconciler{
+		Client:                   cli,
+		Scheme:                   s,
+		MyNode:                   testNodeName,
+		MyNamespace:              testNamespace,
+		Logger:                   testLogger(),
+		RouterProvider:           &mockRouterProvider{},
+		DatapathConfigurator:     datapathConfigurator,
+		FRRReloadSocket:          newFRRSocketServer(t),
+		FRRConfigPath:            frrConfigPath,
+		notStaticConfigsListOpts: &client.ListOptions{LabelSelector: labels.NewSelector().Add(*notMirrored)},
+	}
+}
+
+func newFRRSocketServer(t *testing.T) string {
+	t.Helper()
+	socketPath := filepath.Join(t.TempDir(), "frr.sock")
+	listener, err := net.Listen("unix", socketPath)
+	if err != nil {
+		t.Fatalf("failed to create unix socket: %v", err)
+	}
+	server := &http.Server{
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		}),
+	}
+	go func() { _ = server.Serve(listener) }()
+	t.Cleanup(func() { _ = server.Close() })
+	return socketPath
+}
+
+func assertStatusReady(t *testing.T, cli client.Client) {
+	t.Helper()
+	status := getNodeStatus(t, cli)
+	if status.Status == nil {
+		t.Fatal("Status should not be nil")
+	}
+	if len(status.Status.FailedResources) != 0 {
+		t.Errorf("expected no failed resources, got: %+v", status.Status.FailedResources)
+	}
+	if !apimeta.IsStatusConditionPresentAndEqual(status.Status.Conditions, v1alpha1.ConditionTypeReady, metav1.ConditionTrue) {
+		t.Errorf("expected Ready=True, got conditions: %+v", status.Status.Conditions)
+	}
+	if !apimeta.IsStatusConditionPresentAndEqual(status.Status.Conditions, v1alpha1.ConditionTypeDegraded, metav1.ConditionFalse) {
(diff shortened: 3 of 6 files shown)


RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/acdeb85224c5c365b157a6c4e42d76c9ec750c3c..bef0ce1ee82a014a11c88e854d5e76d713909c25): fix linter issues

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/bef0ce1ee82a014a11c88e854d5e76d713909c25..adb9dda2fcbe559096a7ee663de0aefe56a8bd55): fix function ordering

internal/controller/routerconfiguration/node_status_test.go
```diff
@@ -0,0 +1,212 @@
...
+		t.Errorf("Ready status = %s, want %s", ready.Status, metav1.ConditionFalse)
+	}
+	degraded := apimeta.FindStatusCondition(status.Status.Conditions, v1alpha1.ConditionTypeDegraded)
+	if degraded == nil {
+		t.Fatal("Degraded condition not set")
+	}
+	if degraded.Status != metav1.ConditionTrue {
+		t.Errorf("Degraded status = %s, want %s", degraded.Status, metav1.ConditionTrue)
+	}
+}
```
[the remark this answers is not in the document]
RamLavi: DONE, please recheck

internal/controller/routerconfiguration/node_status_test.go
```diff
@@ -0,0 +1,212 @@
...
+		MyNode:                   testNodeName,
+		MyNamespace:              testNamespace,
+		Logger:                   testLogger(),
+		RouterProvider:           &mockRouterProvider{},
+		HostConfigurator:         hostConfigurator,
+		FRRReloadSocket:          newFRRSocketServer(t),
+		FRRConfigPath:            frrConfigPath,
+		notStaticConfigsListOpts: &client.ListOptions{LabelSelector: labels.NewSelector().Add(*notMirrored)},
+	}
+}
```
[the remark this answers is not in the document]
RamLavi: DONE, please recheck

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/adb9dda2fcbe559096a7ee663de0aefe56a8bd55..8332a82ffbdd6ac7828acb31e4b3ba9872300e1a): change buildStatus to receive the old condition thus natively handling the LastTS issue.

internal/controller/routerconfiguration/node_status.go
```diff
@@ -39,7 +39,11 @@ func (r *PERouterReconciler) reconcileNodeStatus(ctx context.Context, reconcileE
 		return fmt.Errorf("failed to create or update node status: %w", err)
 	}
 
-	newStatus := buildStatus(reconcileErr)
+	var existingConditions []metav1.Condition
+	if nodeStatus.Status != nil {
+		existingConditions = nodeStatus.Status.Conditions
+	}
+	newStatus := buildStatus(reconcileErr, existingConditions)
```
qinqon: Let's just pass nodeStatus to build status and do the nil check inside.

qinqon (changes requested):

internal/controller/routerconfiguration/node_status.go
RamLavi: DONE

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/8332a82ffbdd6ac7828acb31e4b3ba9872300e1a..23dfdfe4b1a72ecdb5773db5896d77d0992dbaeb): address @qinqon review

internal/controller/routerconfiguration/status.go
```diff
@@ -10,8 +10,12 @@ import (
 	"github.com/openperouter/openperouter/api/v1alpha1"
 )
 
-func buildStatus(err error) v1alpha1.RouterNodeConfigurationStatusStatus {
+func buildStatus(err error, existingStatus *v1alpha1.RouterNodeConfigurationStatusStatus) v1alpha1.RouterNodeConfigurationStatusStatus {
 	var s v1alpha1.RouterNodeConfigurationStatusStatus
+	if existingStatus != nil && len(existingStatus.Conditions) > 0 {
+		s.Conditions = make([]metav1.Condition, len(existingStatus.Conditions))
+		copy(s.Conditions, existingStatus.Conditions)
+	}
```
qinqon, line 18: I would simplify this as 
```suggestion
func buildStatus(err error, s *v1alpha1.RouterNodeConfigurationStatusStatus) v1alpha1.RouterNodeConfigurationStatusStatus {
if s == nil {
   s = new(v1alpha1.RouterNodeConfigurationStatusStatus)
}
```

e2etests/pkg/openperouter/status.go
```diff
@@ -20,3 +25,28 @@ func GetNodeStatus(cli client.Client, nodeName string) (*v1alpha1.RouterNodeConf
 	}
 	return status, nil
 }
+
+// AssertNodesStatusReady verifies that all RouterNodeConfigurationStatus CRs
+// have Ready=True and Degraded=False conditions set.
+func AssertNodesStatusReady(cli client.Client) {
+	Eventually(func(g Gomega) {
```
qinqon, line 34: Add the "GinkgoHelper" call here so ginkgo is able to propertly backtrace the error.

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -299,6 +300,9 @@ func (r *PERouterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return fmt.Errorf("failed to build label requirement: %w", err)
 	}
 	r.notStaticConfigsListOpts = &client.ListOptions{LabelSelector: labels.NewSelector().Add(*notMirrored)}
+	if r.HostConfigurator == nil {
+		r.HostConfigurator = configureInterfaces
+	}
```
qinqon: How is this related ?

qinqon (changes requested):

e2etests/pkg/openperouter/status.go
RamLavi: DONE

internal/controller/routerconfiguration/underlay_vni_controller.go
RamLavi: Needed for testability. 
The unit tests inject a mock HostConfigurator (noop/failing) to test status updates through Reconcile without requiring real network namespaces. Production defaults to configureInterfaces via this nil check in SetupWithManager.

internal/controller/routerconfiguration/status.go
RamLavi: but we're not sending with nil, so we will not get into new() function, and it will create an alias bug (see [example](https://go.dev/play/p/8q8eL5QX2MS)). That could only work if I send a copy of the status (see [fixed](https://go.dev/play/p/oacsXO2Zvnr)) - is that what you mean?
```go
// node_status.go
newStatus := buildStatus(reconcileErr, nodeStatus.Status.DeepCopy())

// status.go
func buildStatus(err error, s *v1alpha1.RouterNodeConfigurationStatusStatus) v1alpha1.RouterNodeConfigurationStatusStatus {
	if s == nil {
		s = new(v1alpha1.RouterNodeConfigurationStatusStatus)
	}

	if err == nil {
		setReady(s, v1alpha1.ConditionReasonConfigSuccessful, "All configuration applied successfully")
		return *s
	}

	s.FailedResources = openpeerrors.CollectFailures(err)
	reason, msg := degradedReason(err, s.FailedResources)
	setDegraded(s, reason, msg)
	return *s
}
```

qinqon: Just to clarify what we mean here is that if we change the original status then the DeepEquals thing always return true, so we will not update it.

So let's not do what I propose.

e2etests/pkg/openperouter/status.go
```diff
@@ -20,3 +26,29 @@ func GetNodeStatus(cli client.Client, nodeName string) (*v1alpha1.RouterNodeConf
...
 }
+
+// AssertNodesStatusReady verifies that all RouterNodeConfigurationStatus CRs
+// have Ready=True and Degraded=False conditions set.
+func AssertNodesStatusReady(cli client.Client) {
+	GinkgoHelper()
+	Eventually(func(g Gomega) {
+		statusList := &v1alpha1.RouterNodeConfigurationStatusList{}
+		g.Expect(cli.List(context.Background(), statusList, &client.ListOptions{Namespace: Namespace})).To(Succeed())
+		g.Expect(statusList.Items).NotTo(BeEmpty(), "should have at least one node status")
```
qinqon, line 37: Don't we check the FailedResources too ?

qinqon (changes requested):

internal/controller/routerconfiguration/status.go
RamLavi: agreed, closing the thread then.

e2etests/pkg/openperouter/status.go
RamLavi: yeah we should check that's it's empty as well.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/01b484ab14c4195bac1abb987fc8ba65062f507a..1deb2bc70fec1f05bfbfca77f1ae67f5caed703b): Adding assertion for empty FailedResources on e2e helper function

qinqon (approved):
LGTM

RamLavi:
@maiqueb PTAL

qinqon (approved):
LGTM

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -299,6 +300,9 @@ func (r *PERouterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return fmt.Errorf("failed to build label requirement: %w", err)
 	}
 	r.notStaticConfigsListOpts = &client.ListOptions{LabelSelector: labels.NewSelector().Add(*notMirrored)}
+	if r.HostConfigurator == nil {
```
fedepaol: let's init this from outside. Also, I don't think having configureInterfaces as a variable we override is not that terrible

maiqueb:
@RamLavi please rebase.

internal/controller/routerconfiguration/status.go
```diff
@@ -10,8 +10,12 @@ import (
 	"github.com/openperouter/openperouter/api/v1alpha1"
 )
 
-func buildStatus(err error) v1alpha1.RouterNodeConfigurationStatusStatus {
+func buildStatus(err error, existingStatus *v1alpha1.RouterNodeConfigurationStatusStatus) v1alpha1.RouterNodeConfigurationStatusStatus {
```
maiqueb, line 13: would it make sense to pass the existing conditions instead ? 

Like, that's the only thing you're using on the status ...

internal/controller/routerconfiguration/underlay_vni_controller.go
maiqueb: +1 on injecting the variable from outside always.

internal/controller/routerconfiguration/node_status_test.go
```diff
@@ -0,0 +1,212 @@
...
+			t.Fatalf("Reconcile() returned error: %v", err)
+		}
+
+		assertStatusReady(t, r.Client)
+	})
+
+	t.Run("sets Degraded status on reconcile error", func(t *testing.T) {
+		r := newTestPERouterReconciler(t, failingHostConfigurator, testNode())
+
+		_, _ = r.Reconcile(context.Background(), ctrl.Request{})
```
maiqueb: why don't you check the error here ? fwiw, you're checking it on the other tests (which also run a mock ... don't they ? )

internal/controller/routerconfiguration/node_status_test.go
```diff
@@ -0,0 +1,212 @@
...
+		first := getNodeStatus(t, r.Client)
+		firstRV := first.ResourceVersion
+
+		if _, err := r.Reconcile(context.Background(), ctrl.Request{}); err != nil {
+			t.Fatalf("second Reconcile() returned error: %v", err)
+		}
+		second := getNodeStatus(t, r.Client)
+
+		if second.ResourceVersion != firstRV {
+			t.Errorf("ResourceVersion changed from %s to %s; expected no status patch (reconcile storm)", firstRV, second.ResourceVersion)
```
maiqueb, line 73: would it make sense to also check the timestamp ?

internal/controller/routerconfiguration/node_status_test.go
```diff
@@ -0,0 +1,212 @@
...
+	if ready.Status != metav1.ConditionTrue {
+		t.Errorf("Ready status = %s, want %s", ready.Status, metav1.ConditionTrue)
+	}
+	degraded := apimeta.FindStatusCondition(status.Status.Conditions, v1alpha1.ConditionTypeDegraded)
+	if degraded == nil {
+		t.Fatal("Degraded condition not set")
+	}
+	if degraded.Status != metav1.ConditionFalse {
+		t.Errorf("Degraded status = %s, want %s", degraded.Status, metav1.ConditionFalse)
+	}
```
maiqueb: nit: maybe you want to use `apimeta.IsStatusConditionPresentAndEqual`

internal/controller/routerconfiguration/node_status_test.go
```diff
@@ -0,0 +1,212 @@
...
+	if ready.Status != metav1.ConditionFalse {
+		t.Errorf("Ready status = %s, want %s", ready.Status, metav1.ConditionFalse)
+	}
+	degraded := apimeta.FindStatusCondition(status.Status.Conditions, v1alpha1.ConditionTypeDegraded)
+	if degraded == nil {
+		t.Fatal("Degraded condition not set")
+	}
+	if degraded.Status != metav1.ConditionTrue {
+		t.Errorf("Degraded status = %s, want %s", degraded.Status, metav1.ConditionTrue)
+	}
```
maiqueb: ditto

e2etests/pkg/openperouter/status.go
```diff
@@ -20,3 +26,31 @@ func GetNodeStatus(cli client.Client, nodeName string) (*v1alpha1.RouterNodeConf
...
+
+		for _, nodeStatus := range statusList.Items {
+			g.Expect(nodeStatus.Status).NotTo(BeNil(),
+				fmt.Sprintf("node-status %q should have status set", nodeStatus.Name))
+			readyCond := apimeta.FindStatusCondition(nodeStatus.Status.Conditions, v1alpha1.ConditionTypeReady)
+			g.Expect(readyCond).NotTo(BeNil(),
+				fmt.Sprintf("node-status %q should have a Ready condition", nodeStatus.Name))
+			g.Expect(readyCond.Status).To(Equal(metav1.ConditionTrue),
+				fmt.Sprintf("node-status %q Ready should be True", nodeStatus.Name))
```
maiqueb: nit: I think you could `apimeta.IsStatusConditionPresentAndEqual`. 

Or do you think it's valuable to separate the "condition not found" from the "unexpected condition" errors ?

maiqueb (changes requested):
minor comments. 

to me, the deal breaker is how you inject the configureInterface fn into the open pe router reconciler.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/1deb2bc70fec1f05bfbfca77f1ae67f5caed703b..61c4b75ab29c2a844b5a3aa7634da464e087bf10): Rebase with conflicts:
  1. Keeping upstream's DatapathConfigurator field (not our HostConfigurator)
  2. Keeping upstream's Reconcile() call signature with both r.DatapathConfigurator and configureFRR
  3. Removing our SetupWithManager nil-check default (upstream sets DatapathConfigurator from the caller)
  4. Updating node_status_test.go to use failingDatapathConfigurator struct (matching upstream's
  noopDatapathConfigurator pattern) instead of the old failingHostConfigurator function

internal/controller/routerconfiguration/underlay_vni_controller.go
RamLavi: yes, this was fixed as part of the rebase conflict resolution.

internal/controller/routerconfiguration/status.go
RamLavi: note that @qinqon [asked](https://github.com/openperouter/openperouter/pull/542#discussion_r3511824131) the opposite, to pass the full status and handle the nil check inside buildStatus.
honestly I could argue both ways. your call

internal/controller/routerconfiguration/node_status_test.go
RamLavi: yeah totally. fixed.

internal/controller/routerconfiguration/node_status_test.go
RamLavi: good idea

e2etests/pkg/openperouter/status.go
RamLavi: not important imo
fixed

internal/controller/routerconfiguration/node_status_test.go
RamLavi: fixed

internal/controller/routerconfiguration/node_status_test.go
RamLavi: fixed

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/61c4b75ab29c2a844b5a3aa7634da464e087bf10..bb4cd39737ce2910d2e94afea3b66ec08091e9a7): address @fedepaol and @maiqueb reviews

internal/controller/routerconfiguration/status.go
maiqueb: I don't find this a deal-breaker tbh, but I'd like to understand from @qinqon what's the advantage.

maiqueb:
you need to address the "check generated" lane. Other than that, LGTM.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/bb4cd39737ce2910d2e94afea3b66ec08091e9a7..541974d9d2fa6333cb2458c42efa5baa8d8a920d): fix go fmt issue

maiqueb (approved):
Thank you.

RamLavi:
@fedepaol can you take a look?

internal/controller/routerconfiguration/status.go
```diff
@@ -10,8 +10,12 @@ import (
...
 	var s v1alpha1.RouterNodeConfigurationStatusStatus
+	if existingStatus != nil && len(existingStatus.Conditions) > 0 {
+		s.Conditions = make([]metav1.Condition, len(existingStatus.Conditions))
+		copy(s.Conditions, existingStatus.Conditions)
```
fedepaol, line 17: today we always set the same 2 conditions. This might cause bugs if in the future we introduce new optional conditions (if it even makes sense).
Let's say that there's an optional "FooCondition" together with ready or degraded, we might end up with an entry that is copied from the previous status.

It's fine to ignore if you think it's not a problem.

maiqueb: I understand the concern, but I would say that the scenario you refer to would be easily caught in unit tests, and thus we'd have to fix it in the prod code.

IMHO, that mitigates the concern. Having said that, I'm OK with @RamLavi tackling it now (just carry over the 2 conditions you know you want).

RamLavi: I agree it's worth keeping in mind, but since we always set both conditions today and any new condition would come with its own tests, I'd rather not add complexity until we need it. 
If we ever introduce optional conditions we'll catch it then.

maiqueb:
@RamLavi please rebase, then I can merge.

LGTM.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/541974d9d2fa6333cb2458c42efa5baa8d8a920d..e3ef19fcaa69f9d0407f76aa2ff0f7ad13512bb9): Rebase (no conflicts)
FYI @maiqueb

maiqueb:
@RamLavi it fails on unit tests + check generated ... you need to sort that out.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/e3ef19fcaa69f9d0407f76aa2ff0f7ad13512bb9..f1dbfbec005938c2c3bc1bdffe13da3c6a6faac5): 
  Commit 1 (fix/status) - node_status_test.go:
  - Fixed mockRouter.CanReconcile signature: removed context.Context param to match updated Router interface
  - Added FailedResources empty assertion in assertStatusReady (CodeRabbit review)

maiqueb:
> 🧹 Nitpick comments (1)
> 
> > internal/controller/routerconfiguration/node_status_test.go (1)> `105-117`: _🎯 Functional Correctness_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Assert transition-time updates during recovery.**
> > > This passes if recovery incorrectly retains the prior timestamps. Capture both conditions after the failed reconcile and assert their `LastTransitionTime` values change after the successful reconcile.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@internal/controller/routerconfiguration/node_status_test.go` around lines 105
> > > - 117, Update the “transitions from Degraded to Ready” test to capture both
> > > condition LastTransitionTime values after the failed Reconcile and assert that
> > > each value changes after the successful Reconcile, while preserving the existing
> > > status assertions.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> 
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Nitpick comments:
> In `@internal/controller/routerconfiguration/node_status_test.go`:
> - Around line 105-117: Update the “transitions from Degraded to Ready” test to
> capture both condition LastTransitionTime values after the failed Reconcile and
> assert that each value changes after the successful Reconcile, while preserving
> the existing status assertions.
> ```
> 
> ℹ️ Review info

 @RamLavi please take this one into account as well. IMHO, it does make sense to assert the timestamp is updated on state transitions.

RamLavi:
> > 🧹 Nitpick comments (1)
> > > internal/controller/routerconfiguration/node_status_test.go (1)> `105-117`: _🎯 Functional Correctness_ | _🔵 Trivial_ | _⚡ Quick win_
> > > > **Assert transition-time updates during recovery.**
> > > > This passes if recovery incorrectly retains the prior timestamps. Capture both conditions after the failed reconcile and assert their `LastTransitionTime` values change after the successful reconcile.
> > > > 🤖 Prompt for AI Agents
> > > > ```
> > > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > > rest with a brief reason, keep changes minimal, and validate.
> > > > 
> > > > In `@internal/controller/routerconfiguration/node_status_test.go` around lines 105
> > > > - 117, Update the “transitions from Degraded to Ready” test to capture both
> > > > condition LastTransitionTime values after the failed Reconcile and assert that
> > > > each value changes after the successful Reconcile, while preserving the existing
> > > > status assertions.
> > > > ```
> > 
> > 
> > 🤖 Prompt for all review comments with AI agents
> > ```
> > Verify each finding against current code. Fix only still-valid issues, skip the
> > rest with a brief reason, keep changes minimal, and validate.
> > 
> > Nitpick comments:
> > In `@internal/controller/routerconfiguration/node_status_test.go`:
> > - Around line 105-117: Update the “transitions from Degraded to Ready” test to
> > capture both condition LastTransitionTime values after the failed Reconcile and
> > assert that each value changes after the successful Reconcile, while preserving
> > the existing status assertions.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ℹ️ Review info
> 
> @RamLavi please take this one into account as well. IMHO, it does make sense to assert the timestamp is updated on state transitions.

Good point in principle, but this isn't reliably testable with the controller-runtime fake client: AFAIU `metav1.Time` [serializes](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/time.go#L141) via `time.RFC3339` which has second-level resolution, so sub-second differences are lost ([playground example](https://go.dev/play/p/SXNE_XWrebJ)). Both reconciles in the test complete well under a second, making `LastTransitionTime` identical even when the status genuinely transitions. The fake client doesn't expose a fake clock to control this.
We can alternatively make the time paramerized so it can be faked but I think it's an overkill.

The transition correctness is already covered by asserting the condition values flip (Ready: False->True, Degraded: True->False). Adding a `time.Sleep(time.Second)` to force distinct timestamps would make the test slower and more fragile for marginal value.

RamLavi:
All 4 e2e failures seem unrelated to this PR:

**helm [lane](https://github.com/openperouter/openperouter/actions/runs/29736040388/job/88334248086?pr=542):**
- `Single Session Baseline / verifies L2 and L3 connectivity` - curl connection timeout (exit 28) during L3 connectivity check. 
- `SRV6 routes / single stack ipv4 without gatewayIPs` - same curl timeout pattern (exit 28) on `192.170.20.2:8090`.

**systemdmode [lane](https://github.com/openperouter/openperouter/actions/runs/29736040388/job/88334248106?pr=542):**
- `Controller auto-recovers when operator deletes named netns` - timed out after 180s waiting for netns recreation. 
- `Hybrid mode: static files and API configuration` - `neighbor frr-k8s-daemon-lb6gf to red is not established` after 300s. inotify/FileWatcher doesn't fire in nested containers. matches known flake #428

This PR only touches status reconciliation logic and e2e status assertions. None of the flakes involve status conditions.

maiqueb (approved):
tks

PR #543, opened 2026-07-02
feature(CNIDevice): API and Controller

qinqon:
**Is this a BUG FIX or a FEATURE ?**:

/kind feature

**What this PR does / why we need it**:

multus mode replacement: API + controller per [`enhancements/controller-provisioned-underlay-interfaces.md`](https://github.com/openperouter/openperouter/blob/main/enhancements/controller-provisioned-underlay-interfaces.md): a new `CNI` mode in the underlay `interfaces` union so the controller provisions underlay interfaces in the router netns by invoking CNI plugins, replacing the removed Multus-based underlay integration.

**Special notes for your reviewer**:

DHCP daemon support will be done on follow up PR.

The CNI CHECK call will be implemented at follow up PR.

**Release note**:

```release-note
The underlay `interfaces` union supports a new `CNI` type: the controller provisions underlay interfaces in the router namespace by invoking a CNI plugin defined inline in the Underlay spec (`cniDevice.rawConfig`), with IPAM delegated to the plugin. The macvlan and static CNI plugins are bundled in the controller image. The non-functional `multusNetworkAnnotation` chart value and operator API field were removed.
```

**AI Guidelines Acknowledgment**:

- [x] I have reviewed all changes in this PR, including any AI-generated content, and I take full responsibility for its accuracy and correctness.

## Summary by CodeRabbit

* **New Features**
  * Added CNI-provisioned interfaces for Underlay resources, including raw CNI configuration, interface naming, and runtime capabilities.
  * Added lifecycle handling for creating, updating, restoring, and removing CNI interfaces.
  * Added an EVPN CNI underlay example and deployment workflow.

* **Bug Fixes**
  * Prevented duplicate CNI provisioning and preserved interfaces across controller or router restarts.
  * Added validation preventing mixed interface types and enforcing immutable CNI configuration.

* **Documentation**
  * Updated API, CRD, configuration, and Helm documentation with CNI underlay guidance.

* **Chores**
  * Replaced the single CNI binary directory setting with configurable plugin directories.
  * Removed Multus-based underlay annotation configuration.

PR #544, opened 2026-07-02
Controller cni integration plugins install

maiqueb:

>
> /kind bug
> /kind cleanup

/kind feature

> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

## CNI plugin integration for underlay interface provisioning

The controller image now bundles CNI plugin binaries (macvlan, ipvlan, static, dhcp) from containernetworking/plugins v1.6.2. 

A new internal/cni.Invoker wraps libcni and is initialized at controller startup via --cni-bin-dir and --cni-cache-dir flags. This lays the groundwork for the controller to provision underlay interfaces directly in the router netns, replacing the removed Multus-based integration.

All deployment methods (helm, operator, kustomize, systemd quadlet) are wired with the new flags and a persistent cache volume. CI verifies the binaries are present in the image before publishing the artifact, and e2e tests validate them at runtime across both K8s and systemd modes.

**Release note**:

```release-note
The controller now bundles CNI plugin binaries (macvlan, ipvlan, static, dhcp) and exposes a libcni-based invoker, preparing for direct underlay interface provisioning in the router netns without Multus.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,17 @@
 ARG FRR_IMAGE=quay.io/frrouting/frr:10.6.0
+ARG CNI_PLUGINS_VERSION=v1.6.2
+
+# Build CNI plugin binaries
+FROM golang:1.26.4 AS cni-plugins-builder
+
+ARG CNI_PLUGINS_VERSION
+ARG TARGETOS
+ARG TARGETARCH
+
+WORKDIR /cni-plugins
+RUN git clone --branch ${CNI_PLUGINS_VERSION} --depth 1 https://github.com/containernetworking/plugins.git .
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} ./build_linux.sh \
+  -ldflags "-extldflags -static"
 
 # Build the manager binary
 FROM golang:1.26.4 AS builder
@@ -47,6 +60,10 @@ COPY --from=builder /go/openperouter/hostbridge .
 COPY --from=builder /go/openperouter/nodemarker .
 COPY --from=builder /go/openperouter/operatorbinary ./operator
 COPY operator/bindata bindata
+COPY --from=cni-plugins-builder /cni-plugins/bin/macvlan /opt/openperouter/cni/bin/
+COPY --from=cni-plugins-builder /cni-plugins/bin/ipvlan /opt/openperouter/cni/bin/
+COPY --from=cni-plugins-builder /cni-plugins/bin/static /opt/openperouter/cni/bin/
+COPY --from=cni-plugins-builder /cni-plugins/bin/dhcp /opt/openperouter/cni/bin/
 # Copy FRR startup configuration to the default location
 COPY systemdmode/frrconfig/daemons /etc/frr/daemons
 COPY systemdmode/frrconfig/vtysh.conf /etc/frr/vtysh.conf
--- a/charts/openperouter/templates/controller.yaml
+++ b/charts/openperouter/templates/controller.yaml
@@ -62,6 +62,8 @@ spec:
         {{- with .Values.openperouter.ovsSocketPath }}
         - --ovssocket={{ . }}
         {{- end }}
+        - --cni-bin-dir={{ .Values.openperouter.controller.cniBinDir }}
+        - --cni-cache-dir={{ .Values.openperouter.controller.cniCacheDir }}
         command:
         - /controller
         env:
@@ -132,6 +134,8 @@ spec:
           mountPropagation: HostToContainer
         - mountPath: /var/run/frr
           name: frr-sockets
+        - mountPath: {{ .Values.openperouter.controller.cniCacheDir | default "/var/lib/openperouter/cni/cache" }}
+          name: cni-cache
       hostNetwork: true
       hostPID: true
       tolerations:
@@ -175,4 +179,8 @@ spec:
           path: /var/run/openperouter/frr
           type: DirectoryOrCreate
         name: frr-sockets
+      - hostPath:
+          path: {{ .Values.openperouter.controller.cniCacheDir | default "/var/lib/openperouter/cni/cache" }}
+          type: DirectoryOrCreate
+        name: cni-cache
 {{- end }}
--- a/cmd/hostcontroller/main.go
+++ b/cmd/hostcontroller/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/labels"
@@ -48,6 +49,7 @@ import (
 	"github.com/openperouter/openperouter/api/static"
 	periov1alpha1 "github.com/openperouter/openperouter/api/v1alpha1"
 	"github.com/openperouter/openperouter/internal/buildversion"
+	"github.com/openperouter/openperouter/internal/cni"
 	"github.com/openperouter/openperouter/internal/controller/routerconfiguration"
 	"github.com/openperouter/openperouter/internal/filewatcher"
 	"github.com/openperouter/openperouter/internal/hostnetwork"
@@ -97,6 +99,8 @@ type parameters struct {
 	nodeName       string
 	namespace      string
 	logLevel       string
+	cniBinDir      string
+	cniCacheDir    string
 }
 
 func main() {
@@ -132,6 +136,10 @@ func main() {
 		systemdctl.HostDBusSocket, "the path of systemd control socket")
 	flag.IntVar(&hostModeParams.routerHealthCheckPort, "router-health-check-port",
 		9080, "the port for router health check endpoint")
+	flag.StringVar(&args.cniBinDir, "cni-bin-dir", "/opt/openperouter/cni/bin/",
+		"colon-separated list of directories to search for CNI plugin binaries")
+	flag.StringVar(&args.cniCacheDir, "cni-cache-dir", "/var/lib/openperouter/cni/cache",
+		"directory to store CNI result cache")
 
 	flag.Parse()
 
@@ -146,6 +154,30 @@ func main() {
 	ctrl.SetLogger(logr.FromSlogHandler(logger.Handler()))
 	setupLog.Info("version", "version", buildversion.Version())
 	setupLog.Info("arguments", "args", fmt.Sprintf("%+v", args))
+
+	if args.cniBinDir == "" {
+		setupLog.Info("cni-bin-dir cannot be empty")
+		os.Exit(1)
+	}
+	if args.cniCacheDir == "" {
+		setupLog.Info("cni-cache-dir cannot be empty")
+		os.Exit(1)
+	}
+
+	var cniPluginDirs []string
+	for dir := range strings.SplitSeq(args.cniBinDir, ":") {
+		if trimmed := strings.TrimSpace(dir); trimmed != "" {
+			cniPluginDirs = append(cniPluginDirs, trimmed)
+		}
+	}
+	if len(cniPluginDirs) == 0 {
+		setupLog.Info("no valid CNI plugin directories specified", "cni-bin-dir", args.cniBinDir)
+		os.Exit(1)
+	}
+
+	cniInvoker := cni.NewInvoker(cniPluginDirs, args.cniCacheDir)
+	setupLog.Info("CNI plugin invoker initialized", "pluginDirs", cniInvoker.PluginDirs(), "cacheDir", args.cniCacheDir)
+	_ = cniInvoker
 
 	// Setup signal handler once for the entire process
 	ctx := ctrl.SetupSignalHandler()
--- a/e2etests/suite/suite_test.go
+++ b/e2etests/suite/suite_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -68,6 +69,12 @@ var _ = ginkgo.BeforeSuite(func() {
 	reporter, err := k8s.InitReporter(kubeconfig, tests.ReportPath, openperouter.Namespace, frrk8s.Namespace)
 	Expect(err).NotTo(HaveOccurred(), "failed to initialize k8s reporter (kubeconfig=%s)", kubeconfig)
 	tests.K8sReporter = reporter
+
+	cs := k8sclient.New()
+	ginkgo.By("validating CNI binaries and cache directory in controller")
+	Eventually(func(g Gomega) {
+		tests.ValidateCNIBinaries(g, cs)
+	}).WithTimeout(30 * time.Second).WithPolling(2 * time.Second).Should(Succeed())
 })
 
 var _ = ginkgo.AfterSuite(func() {
--- a/e2etests/tests/hostconfiguration.go
+++ b/e2etests/tests/hostconfiguration.go
@@ -28,6 +28,18 @@ import (
 
 var ValidatorPath string
 
+var cniBinaries = []string{"macvlan", "ipvlan", "static", "dhcp"}
+
+const (
+	cniBinDir   = "/opt/openperouter/cni/bin"
+	cniCacheDir = "/var/lib/openperouter/cni/cache"
+)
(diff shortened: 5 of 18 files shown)


maiqueb:
/cc @qinqon

Please review.

e2etests/tests/hostconfiguration.go
```diff
@@ -1493,3 +1505,49 @@ func routerPodsWithValidator(cs clientset.Interface) ([]*corev1.Pod, error) {
...
+	cniTargets := k8sModeCNITarget
+	if HostMode {
+		cniTargets = hostModeCNITarget
+	}
+	for _, t := range cniTargets(g, cs) {
+		for _, bin := range cniBinaries {
+			path := cniBinDir + "/" + bin
+			out, err := t.exec.Exec("test", "-x", path)
+			g.Expect(err).NotTo(HaveOccurred(), "CNI binary %s missing or not executable on %s: %s", path, t.label, out)
+		}
```
maiqueb, line 1519: maybe this is redundant, since we already check the img. 

... still, super cheap to test also in runtime.

Dockerfile
```diff
@@ -1,4 +1,17 @@
...
+FROM golang:1.26.4 AS cni-plugins-builder
+
+ARG CNI_PLUGINS_VERSION
+ARG TARGETOS
+ARG TARGETARCH
+
+WORKDIR /cni-plugins
+RUN git clone --branch ${CNI_PLUGINS_VERSION} --depth 1 https://github.com/containernetworking/plugins.git .
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} ./build_linux.sh \
+  -ldflags "-extldflags -static"
```
qinqon, line 14: We can directly install the binaries from the the tarball at https://github.com/containernetworking/plugins/releases.

charts/openperouter/templates/controller.yaml
```diff
@@ -62,6 +62,8 @@ spec:
         {{- with .Values.openperouter.ovsSocketPath }}
         - --ovssocket={{ . }}
         {{- end }}
+        - --cni-bin-dir={{ .Values.openperouter.controller.cniBinDir | default "/opt/openperouter/cni/bin/" }}
```
qinqon: bin-dir is not necesary, we have all the binaries we need inside the container.

hack/verify-cni-binaries.sh
```diff
@@ -0,0 +1,33 @@
...
+        echo "OK"
+    fi
+done
+
+if [ "${failed}" -eq 1 ]; then
+    echo "ERROR: one or more CNI binaries are missing"
+    exit 1
+fi
+
+echo "All CNI binaries verified successfully"
```
qinqon: First time I see we check the contents of the container, not sure it's worth it, what we have to check if that we can use CNIs with e2e if they fail we will discover what is missing.

e2etests/suite/suite_test.go
```diff
@@ -68,6 +69,12 @@ var _ = ginkgo.BeforeSuite(func() {
 	reporter, err := k8s.InitReporter(kubeconfig, tests.ReportPath, openperouter.Namespace, frrk8s.Namespace)
 	Expect(err).NotTo(HaveOccurred(), "failed to initialize k8s reporter (kubeconfig=%s)", kubeconfig)
 	tests.K8sReporter = reporter
+
+	cs := k8sclient.New()
+	ginkgo.By("validating CNI binaries and cache directory in controller")
+	Eventually(func(g Gomega) {
+		tests.ValidateCNIBinaries(g, cs)
+	}).WithTimeout(30 * time.Second).WithPolling(2 * time.Second).Should(Succeed())
 })
```
qinqon, line 78: We will propertly test al CNIs we provide at e2e, is not worth it to test containers contents.

qinqon (changes requested):
Let's remove the container contents tests is very low level thing, at the follow PR that implementes api and controller we will add e2e tests for all the CNIs, we will discover problems at container there.

charts/openperouter/templates/controller.yaml
maiqueb: I know. This is in case the user wants to override it for whatever reason.

Dockerfile
maiqueb: imho po-tay-to po-ta-to. 

It takes seconds to build this.

hack/verify-cni-binaries.sh
maiqueb: I just want to have tests that prove THIS PR works. 

Saying we'll test it in a follow-up PR is not good enough for me - hence the trouble of adding these tests.

e2etests/suite/suite_test.go
maiqueb: exactly. But I don't agree with "e2e tests will be provided in the future". 

I'm testing that what I'm shipping in **this** PR works as I expect / need. 

The follow-up PR can (and should ...) revert this commit, because it'll be pointless once proper e2e tests exist.

charts/openperouter/templates/controller.yaml
qinqon: Then we will impliment that, let's not do it now so we know what we are running.

hack/verify-cni-binaries.sh
qinqon: fair enough, but feels like we are doing the checks here and at e2e ?

e2etests/suite/suite_test.go
qinqon: Ack, I will take this into account

charts/openperouter/templates/controller.yaml
maiqueb: we know what **we** are running. 

I just don't want to close the door on *other* distributions that choose to provide alternative CNIs. i.e. they can mount the host paths into the controller pod and override the cni bin path.

hack/verify-cni-binaries.sh
maiqueb: yeah, that's a question I have ... maybe the deploy time checks should care *only* about the CNI cache ?...

What do you think ?

charts/openperouter/templates/controller.yaml
qinqon: On second though, let's do this at the end we want to put the CNI inside so we support day 0 deployments where CNI are not installed, but for other use cases we will use whatever is installed at node and it should override whatever we provide at container or even provide with new mechanism not provided by container (they may need extra privileted though).

Dockerfile
qinqon: at least parameterize the golang compiler version.

charts/openperouter/templates/controller.yaml
maiqueb: exactly. It's just a matter of filling the template / provide an alternative manifest on downstream distributions.

Dockerfile
maiqueb: well, I'm literally doing exactly what's being done for building our code - check below.

e2etests/suite/suite_test.go
```diff
@@ -68,6 +69,13 @@ var _ = ginkgo.BeforeSuite(func() {
 	reporter, err := k8s.InitReporter(kubeconfig, tests.ReportPath, openperouter.Namespace, frrk8s.Namespace)
 	Expect(err).NotTo(HaveOccurred(), "failed to initialize k8s reporter (kubeconfig=%s)", kubeconfig)
 	tests.K8sReporter = reporter
+
+	cs := k8sclient.New()
+	ginkgo.By("validating CNI binaries and cache directory in controller")
+	Eventually(func() error {
+		tests.ValidateCNIBinaries(cs)
+		return nil
+	}).WithTimeout(30 * time.Second).WithPolling(2 * time.Second).Should(Succeed())
```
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist dude you've allucinated. ValidateCNIBinaries does NOT return an error.

cmd/hostcontroller/main.go
```diff
@@ -147,6 +155,11 @@ func main() {
 	setupLog.Info("version", "version", buildversion.Version())
 	setupLog.Info("arguments", "args", fmt.Sprintf("%+v", args))
 
+	cniPluginDirs := strings.Split(args.cniBinDir, ":")
+	cniInvoker := cni.NewInvoker(cniPluginDirs, args.cniCacheDir)
+	setupLog.Info("CNI plugin invoker initialized", "pluginDirs", cniInvoker.PluginDirs(), "cacheDir", args.cniCacheDir)
+	_ = cniInvoker
```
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist hmmm no ... 

They both have defaults - this is code that defines them:
```
flag.StringVar(&args.cniBinDir, "cni-bin-dir", "/opt/openperouter/cni/bin/",
		"colon-separated list of directories to search for CNI plugin binaries")
	flag.StringVar(&args.cniCacheDir, "cni-cache-dir", "/var/lib/openperouter/cni/cache",
		"directory to store CNI result cache")
```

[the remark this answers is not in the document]
maiqueb: @gemini-code-assist these flags are defined by the operator / whoever writes the manifest. 

It's not a regular user that will be updating these. I think we're safe ...
Please explain the attack vector here.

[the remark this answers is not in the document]
maiqueb: @gemini-code-assist what would you validate against ? empty string ? so the operator - or the person that re-writes the manifest takes the trouble of actually passing an empty string as parameter. Is that it ? 

OK, I've followed your advice. 

How about now ?

hack/verify-cni-binaries.sh
qinqon: I would do all this tests at e2e so we can run at other opneprouter deployments, if we do it at github CI we will not able to run those easily on other deployments.

cmd/hostcontroller/main.go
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist check now.

Dockerfile
qinqon: Ok, fair enough, also it may happend that CNI plugins golang version has to be different so maybe it make sense.

qinqon (changes requested):

hack/verify-cni-binaries.sh
maiqueb: I agree it makes sense to have it on the e2e to check the deployment.

Still, I'd like to have a cheap way to check if the img is properly built. This would be it.

cmd/hostcontroller/main.go
```diff
@@ -147,6 +155,30 @@ func main() {
...
+		}
+	}
+	if len(cniPluginDirs) == 0 {
+		setupLog.Info("no valid CNI plugin directories specified", "cni-bin-dir", args.cniBinDir)
+		os.Exit(1)
+	}
+
+	cniInvoker := cni.NewInvoker(cniPluginDirs, args.cniCacheDir)
+	setupLog.Info("CNI plugin invoker initialized", "pluginDirs", cniInvoker.PluginDirs(), "cacheDir", args.cniCacheDir)
+	_ = cniInvoker
```
fedepaol, line 180: I'd rather have the whole picture and not small bits of this. I can't say anything about invoker being right (or even naming it invoker) if I don't see how it's being used.
So, @maiqueb can you submit this when it's done?

qinqon: > I'd rather have the whole picture and not small bits of this. I can't say anything about invoker being right (or even naming it invoker) if I don't see how it's being used. So, @maiqueb can you submit this when it's done?

@fedepaol this is the PR that will work with this:
- https://github.com/openperouter/openperouter/pull/543

Although the idea is to make it independant by not using CNIs from the container.

charts/openperouter/templates/controller.yaml
fedepaol: I don't think the default bit is required, as we have one default in values.yaml already

Makefile
```diff
@@ -123,8 +123,10 @@ BRANCH = $(shell git rev-parse --abbrev-ref HEAD)
 docker-build: ## Build docker image with the manager.
 	@if [ "$(CONTAINER_ENGINE)" = "podman" ]; then \
 		sudo $(CONTAINER_ENGINE) build  -t ${IMG} .; \
+		hack/verify-cni-binaries.sh "sudo podman" $(IMG); \
```
fedepaol: I agree we should verify the content as this is self contained, but the validation should happen in e2e tests / unit tests imo.

cmd/hostcontroller/main.go
```diff
@@ -147,6 +155,30 @@ func main() {
...
+		if trimmed := strings.TrimSpace(dir); trimmed != "" {
+			cniPluginDirs = append(cniPluginDirs, trimmed)
+		}
+	}
+	if len(cniPluginDirs) == 0 {
+		setupLog.Info("no valid CNI plugin directories specified", "cni-bin-dir", args.cniBinDir)
+		os.Exit(1)
+	}
+
+	cniInvoker := cni.NewInvoker(cniPluginDirs, args.cniCacheDir)
```
fedepaol, line 178: How do we know here (in this pr) that this it the right interface of the invoker and that, by extension, the parameters we are exposing via helm are the right ones?

internal/cni/invoker.go
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package cni
+
+import (
+	"github.com/containernetworking/cni/libcni"
+)
+
+type Invoker struct {
```
fedepaol, line 9: I would say that either we go beyond and add the whole api this type is expected to expose, or we do a step back and remove it all together.

cmd/hostcontroller/main.go
maiqueb: I haven't defined ANY interface of the invoker. I'm just feeding it the 2 parameters I know libcni requires to work: cni bin path, and cni cache path. 

I know this is what CNI library requires, because I've [checked](https://github.com/containernetworking/cni/blob/7c270076995b6a35f4774ce94dafcf266d1c6925/libcni/api.go#L147). 

We can also use a version that doesn't manage the cache, but I think having the operator template the CNI cache path is an interesting feature.

Makefile
maiqueb: but if I just want to see if the img features the proper binaries ?... 

IMHO this is:
- super cheap
- super simple
- fills the user need (I want to check if the img features the bits I require)

cmd/hostcontroller/main.go
maiqueb: I would like us to try to deliver small pieces of the implementation, and pay the integration tax more often.

If this proves to be wrong, we can always go back ...

charts/openperouter/templates/controller.yaml
maiqueb: which default would you rather keep ? this one or the values.yaml ?

internal/cni/invoker.go
maiqueb: I just want to consume the 2 attributes I know libcni requires, and want to do something with them. 

If you prefer, I'll just print them, and get this over with.

fedepaol: ok then, resolving the comment.

charts/openperouter/templates/controller.yaml
fedepaol: values.yaml, as per all our other defaults

Makefile
fedepaol: then we should check that go compiled our binaries, or that everything that is supposed to be in the container is actually there. Which we don't, or we don't in this phase at least. I don't understand how this is different from the rest.

cmd/hostcontroller/main.go
fedepaol: still I don't get how those are going to be used given we feed libcni with them already, but I trust you here and I'll approve this based on that and not on what I am seeing from this PR.

Makefile
maiqueb: Dropped since every one raised the question of if this is useful or not.

charts/openperouter/templates/controller.yaml
maiqueb: Done.

maiqueb (commented):
@fedepaol ptal

fedepaol:
LGTM

PR #545, opened 2026-07-03
docs: separate images for FRR and Grout

zeeke:
Feature

>
> /kind bug
> /kind cleanup
> /kind feature
/kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

To simplify deployment and CI after 
- https://github.com/openperouter/openperouter/pull/338

**Release note**:

```release-note
None
```

### DIFF
--- a/enhancements/image-split.md
+++ b/enhancements/image-split.md
@@ -1,0 +1,153 @@
+# Separate Container Images for FRR and Grout
+
+## Summary
+
+OpenPERouter ships a single monolithic image (`quay.io/openperouter/router:main`)
+containing Go binaries, FRR, and —in the grout variant— DPDK/grout binaries.
+Every container uses this image regardless of what it needs.
+
+This enhancement splits it into three purpose-built images:
+
+1. **`openperouter/perouter`** — Go binaries only.
+2. **FRR image** — user-provided, defaults to upstream `quay.io/frrouting/frr`.
+3. **Grout image** — user-provided, defaults to upstream quay.io/grout/grout.
+
+## Motivation
+
+### Goals
+
+- **User-swappable FRR**: plug in any FRR version via a single Helm value.
+- **Smaller images**: each image carries only what it needs, hence a reduced 
+  security attack surface.
+- **Eliminate `main-grout` tag**: openperouter/perouter:main image would ship
+  the grcli grout client or equivalent.
+- **Independent release cadence**: FRR/grout versions can be bumped without
+  rebuilding the perouter image.
+- **Simplify CI lanes**: Avoid building multiple images in Github CI.
+
+### Non-Goals
+
+- Changing pod topology (same containers, different images).
+- Changing the FRR config mechanism (frr-reload.py, vtysh).
+- Multi-arch grout image (DPDK is x86_64-only).
+
+## Proposal
+
+### Container-to-Image Mapping
+
+| Container | New Image | Notes |
+|-----------|-----------|-------|
+| controller | `perouter` | +grcli |
+| hostbridge | `perouter` | |
+| nodemarker | `perouter` | |
+| operator | `perouter` | |
+| reloader | FRR image | +reloader binary via init container |
+| frr | FRR image | |
+| grout | Grout image | only when datapath=grout |
+| cp-frr-files | FRR image | |
+
+### The Reloader Problem
+
+The reloader invokes `frr-reload.py` and `vtysh`, which must match the
+running FRR version. 
+Solutions:
+
+a. the reloader container runs the **FRR image**,
+and the Go `reloader` binary is injected via an init container:
+
+```yaml
+initContainers:
+  - name: cp-reloader
+    image: <perouter image>
+    command: ["cp", "/reloader", "/shared/reloader"]
+    volumeMounts:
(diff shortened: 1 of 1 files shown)


enhancements/image-split.md
```diff
@@ -0,0 +1,153 @@
...
+  grout:
+    image:
+      repository: quay.io/grout/grout
+      tag: "0.16.0"
+```
+
+
+## Drawbacks
+
+- **More init containers**: one extra in the router pod (`cp-reloader`), one
```
fedepaol, line 136: please note on podman quadlets we don't have init containers, so the end result is a frankenstein of synchronization scripts to mimic what init containers do

enhancements/image-split.md
```diff
@@ -0,0 +1,153 @@
...
+## Motivation
+
+### Goals
+
+- **User-swappable FRR**: plug in any FRR version via a single Helm value.
+- **Smaller images**: each image carries only what it needs, hence a reduced 
+  security attack surface.
+- **Eliminate `main-grout` tag**: openperouter/perouter:main image would ship
+  the grcli grout client or equivalent.
+- **Independent release cadence**: FRR/grout versions can be bumped without
```
fedepaol, line 24: I think there's value in having us dictate the version of frr for a given release. Even if it was separate containers, our manifests would choose what version to use (which is the one we validate CI against), so a bump is still required.

enhancements/image-split.md
```diff
@@ -0,0 +1,153 @@
...
+
+1. **`openperouter/perouter`** — Go binaries only.
+2. **FRR image** — user-provided, defaults to upstream `quay.io/frrouting/frr`.
+3. **Grout image** — user-provided, defaults to upstream quay.io/grout/grout.
+
+## Motivation
+
+### Goals
+
+- **User-swappable FRR**: plug in any FRR version via a single Helm value.
```
fedepaol, line 19: this is a compelling point, especially for cves

enhancements/image-split.md
```diff
@@ -0,0 +1,153 @@
...
+- **Smaller images**: each image carries only what it needs, hence a reduced 
+  security attack surface.
+- **Eliminate `main-grout` tag**: openperouter/perouter:main image would ship
```
fedepaol, line 22: not sure how this is related and what prevents us doing only this part today

enhancements/image-split.md
```diff
@@ -0,0 +1,153 @@
...
+| Container | New Image | Notes |
+|-----------|-----------|-------|
+| controller | `perouter` | +grcli |
+| hostbridge | `perouter` | |
+| nodemarker | `perouter` | |
+| operator | `perouter` | |
+| reloader | FRR image | +reloader binary via init container |
+| frr | FRR image | |
+| grout | Grout image | only when datapath=grout |
+| cp-frr-files | FRR image | |
```
fedepaol, line 47: this is basically what we had in the beginning and we switched over later to reduce the complexity

enhancements/image-split.md
```diff
@@ -0,0 +1,153 @@
...
+    volumeMounts:
+      - name: reloader-bin
+        mountPath: /shared
+```
+
+b. The reloader uses the vtysh and frr-reload.py shipped in 
+the openperouter Dockerfile (this requires verifying if vtysh is compatible with 
+different versions of FRR).
+
+### The Controller + grcli Problem
```
fedepaol, line 79: I might not have understood properly, but is this the only thing preventing us to release one single image today?

fedepaol:
General comment is, we used to have a thinner image with the copy trick, but the complexity of covering that with host based deployments was such that we rolled back and had a single image based out of frr.

Now, iiuc, the only blocker that prevents us to have one single interface is the grcli binary. That aside, it'd be desirable to have frr independent from our version and I feel like that's orthogonal.

enhancements/image-split.md
zeeke: It's doable but with a little more effort:
a. if we keep a one huge image (with controller, FRR, grout): then
  a1. that image can't be an alpine based, as grout doesn't build on that. supposing to use a centos one
    a1a. Dockerfile installs Grout and FRR via RPMs
b. if we use separated images:
  b1. we can still use the upstream `quay.io/frrouting/frr` image
  b2. grout image is taken from quay.io/grout/grout
  b3. controller image can be
    b3a. an alpline based image if we implement a grcli Go library
    b3b. a centos based image if we want to use grcli binary

enhancements/image-split.md
zeeke: got it. 

I agree that this enhancement brings in a relevant complexity. If no other good aspects emerge, I'd close this PR

enhancements/image-split.md
zeeke: yes.

also, with a monolith image, we ship controller+FRR+grout+grcli . If we solve the grcli issue, we can ship `controller+FRR+grcli` and grout is consumed from quay.io/grout/grout

zeeke:
Most of the goals of this enhancement can be easily achieved if and when we have a Golang grout client.
No need to go with other complicated workarounds whose benefits are not as clear as the generated problems

PR #547, opened 2026-07-06
fix(controller): handle underlay NIC change without restart

### LINKED ISSUE #531: Don't kill the frr namespace / pod anymore
Killing the pod was a shortcut to avoid reversing the operations we were performing, while forcing a clean slate and letting the reconciler doing its work.
This was mostly due to the underlay setup (to the point that we are able to clean up l2vnis and l3vnis).

With https://github.com/openperouter/openperouter/pull/442 landing, we can remove the logic entirely and avoid killing / removing the network namespace.

maiqueb:

>
> /kind bug

/kind feature

> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

When the underlay interface changes, instead of returning an `UnderlayExistsError` that triggers
pod and netns deletion, restore the old interfaces to the default namespace and move the new ones in. 

VNIs and bridge-refresh goroutines are cleaned up before the swap.

Removes the UnderlayExistsError type and nonRecoverableHostError check from both PERouterReconciler and StaticConfigReconciler.

Closes: #531

**Release note**:

```release-note
The controller now handles underlay NIC changes by swapping interfaces in-place instead of deleting and recreating the router pod and network namespace.
```


internal/controller/routerconfiguration/host_config.go
```diff
@@ -162,9 +162,32 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
...
+	if len(existing) == 0 {
+		return false
+	}
+	for _, name := range existing {
+		if !slices.Contains(requested, name) {
+			return true
+		}
+	}
+	return false
 }
```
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist this is already being done. 

Check it, and get back to me.

internal/hostnetwork/underlay.go
```diff
@@ -57,17 +57,15 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
-	// the network namespace.
 	existingIfaces, err := findInterfacesInGroup(targetNetNS, underlayGroupID)
 	if err != nil {
 		return fmt.Errorf("failed to check existing underlay interfaces: %w", err)
 	}
-	for _, name := range existingIfaces {
-		if !slices.Contains(params.UnderlayInterfaces, name) {
-			return UnderlayExistsError(fmt.Sprintf(
-				"existing underlay found: %s, new interfaces are %v", name, params.UnderlayInterfaces))
+	if underlayInterfacesChanged(existingIfaces, params.UnderlayInterfaces) {
```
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist done already. is it OK ?

internal/controller/routerconfiguration/host_config.go
```diff
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"slices"
```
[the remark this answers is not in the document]
maiqueb: Done. @gemini-code-assist

internal/controller/routerconfiguration/host_config.go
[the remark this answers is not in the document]
maiqueb: @gemini-code-assist check the PR for the presence of the function you're telling me to delete. I don't see it.

internal/controller/routerconfiguration/host_config.go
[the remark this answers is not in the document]
maiqueb: the import was already removed.

internal/hostnetwork/underlay.go
```diff
@@ -107,10 +105,32 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
+		if !slices.Contains(requested, name) {
+			return true
+		}
+	}
+	return false
+}
+
+// UnderlayInterfaces returns the names of all underlay interfaces
+// currently present in the given network namespace.
+func UnderlayInterfaces(namespace string) ([]string, error) {
```
fedepaol, line 125: I'd rename this to what really does, which is returning all the interfaces in the given ns and not only the underlay ones

internal/hostnetwork/underlay.go
```diff
@@ -107,10 +105,32 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
 	return nil
 }
 
-type UnderlayExistsError string
+func UnderlayInterfacesChanged(existing, requested []string) bool {
+	if len(existing) == 0 {
+		return false
+	}
+	for _, name := range existing {
+		if !slices.Contains(requested, name) {
```
fedepaol, line 116: restoreUnderlay it's all or none, while it might make sense to save what's already in the network namespace.

So I'd change this to return the interfaces that were removed and act only on those.

internal/controller/routerconfiguration/host_config.go
```diff
@@ -162,9 +161,20 @@ func configureInterfaces(ctx context.Context, config interfacesConfiguration) er
...
-	return errors.As(e, &underlayExistsError)
+func setupUnderlayWithCleanup(ctx context.Context, targetNamespace string, underlayParams hostnetwork.UnderlayParams) error {
+	existing, err := hostnetwork.UnderlayInterfaces(targetNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to list existing underlay interfaces: %w", err)
+	}
+	if hostnetwork.UnderlayInterfacesChanged(existing, underlayParams.UnderlayInterfaces) {
+		slog.InfoContext(ctx, "underlay interfaces changed, cleaning up VNIs before interface swap",
+			"existing", existing, "requested", underlayParams.UnderlayInterfaces)
+		if err := hostnetwork.RemoveAllVNIs(targetNamespace); err != nil {
```
fedepaol: why we wipe all the vnis here?

internal/hostnetwork/underlay.go
```diff
@@ -107,10 +109,32 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
-type UnderlayExistsError string
+func UnderlayInterfacesWereRemoved(existing, requested []string) bool {
+	if len(existing) == 0 {
+		return false
+	}
+	for _, name := range existing {
+		if !slices.Contains(requested, name) {
+			return true
+		}
+	}
```
qinqon, line 119: I would reter use a Set for this kind of things, although the amount of interfaces is not going to be massive.

Also you can remove the for loop using slices.ContainsFunc 
```golang
slices.ContainsFunc(existing, func(name string) bool {
   return !slices.Contains(requested, name)
})
```

internal/hostnetwork/underlay.go
```diff
@@ -107,10 +109,32 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
 	return nil
 }
 
-type UnderlayExistsError string
+func UnderlayInterfacesWereRemoved(existing, requested []string) bool {
+	if len(existing) == 0 {
+		return false
+	}
```
qinqon: we can add another optimization if len(requested) < len(exisging) -> true

internal/controller/routerconfiguration/host_config.go
```diff
@@ -198,11 +197,22 @@ func restoreUnderlay(ctx context.Context, targetNamespace string) {
...
+	existing, err := hostnetwork.UnderlayInterfaces(targetNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to list existing underlay interfaces: %w", err)
+	}
+	if hostnetwork.UnderlayInterfacesWereRemoved(existing, underlayParams.UnderlayInterfaces) {
+		slog.InfoContext(ctx, "underlay interfaces changed, cleaning up VNIs before interface swap",
+			"existing", existing, "requested", underlayParams.UnderlayInterfaces)
+		if err := hostnetwork.RemoveAllVNIs(targetNamespace); err != nil {
+			slog.Warn("failed to remove vnis during underlay change", "err", err)
+		}
```
qinqon: Shouldn't we remove the VNIs affected by the removed interface instead of removing all the VNIs ?

qinqon (changes requested):

internal/controller/routerconfiguration/host_config.go
fedepaol: :+1:

maiqueb: right, we should. I forgot to filter them.

internal/hostnetwork/underlay.go
maiqueb: I'd rather keep the for loop over that.

internal/hostnetwork/underlay.go
maiqueb: sure, I can.

internal/controller/routerconfiguration/host_config.go
maiqueb: but now I do wonder ... how do we map the VNI to the exact interface ? like, all VNIs go via the same underlay interface, don't they ?

internal/controller/routerconfiguration/host_config.go
maiqueb: AFAIU, when we're here, it means the physical interface that all VXLANs are tunneled over is about to disappear (it gets restored to the default netns).

Thus, all the vxlan dataplane traffic will be broken - since there's no valid tunnel endpoint. I'm just cleaning resources, which are no longer usable. Plus I don't want to "reconnect" them when a new underlay interface is configured - I'd rather start from scratch.

internal/hostnetwork/underlay.go
maiqueb: I can do that, but I'd like some clarity on the affected VNIs (on this function's calling site).

internal/hostnetwork/underlay.go
maiqueb: I am not sure we need to use a set, but I will if you insist. 

Regarding the for, I'd rather use that simpler API, than this newer `slices` pkg.

internal/hostnetwork/underlay.go
maiqueb: all the interfaces with the groupID ...

Do we have a similar function elsewhere ?... odd if not

internal/controller/routerconfiguration/host_config.go
qinqon: I have see something call "Interfaces" at the ISIS I think don't know if it's related.

internal/controller/routerconfiguration/host_config.go
fedepaol: not really. If you have multiple interfaces connected to multiple switches, your packets may be ecmp-ed until one of the link goes down (or gets removed). In that case, only some sessions (those going via that route) will be affected, while the others will persist.

IIRC, vxlan uses a function of the inner session for the src port, to ensure all the packets belonging to a given session (ie tcp) go via the same route.

internal/controller/routerconfiguration/host_config.go
fedepaol: we get here when _at least_ one underlay is removed. If we have remaining ones, we shouldn't wipe.

internal/hostnetwork/underlay.go
fedepaol: ah sorry I didn't see the final line. Ignore me

internal/controller/routerconfiguration/host_config.go
maiqueb: and can I map that ? how can I understand which interfaces affect which VNIs ?...

internal/controller/routerconfiguration/host_config.go
maiqueb: apologies @fedepaol but you be more precise ? That I should not wipe the VNIs if we have at least one existing interface left in the router netns ?

internal/controller/routerconfiguration/host_config.go
fedepaol: one optimization can be to remove vnis if you don't have vteps related to that vni on a given interface... but I would go on the easy path and remove vnis only if we are removing all the interfaces, unless there's a functional reason on why we might want to remove them.

maiqueb: let's take the easy path then ... 

only clear all VNIs if we've cleared all interfaces.

internal/controller/routerconfiguration/host_config.go
maiqueb: done - I'm only clearing the VNIs when all interfaces are removed.

internal/controller/routerconfiguration/host_config.go
maiqueb: done

internal/hostnetwork/underlay.go
maiqueb: I did what @fedepaol asked instead - return the removed interfaces list, leaving the caller to make the decision.

internal/controller/routerconfiguration/host_config.go
fedepaol: yep that was my suggestion, unless we have a valid reason for not keeping the vnis.

fedepaol:
@zeeke please review this PR

e2etests/tests/hostconfiguration.go
```diff
@@ -431,14 +431,57 @@ var _ = ginkgo.Describe("Router Host configuration", func() {
...
+				for _, old := range routerPods {
+					if old.Spec.NodeName == p.Spec.NodeName {
+						oldUID = string(old.UID)
+					}
 				}
-				return podsRolled(cs, routerPods, newRouterPods)
-			}, time.Minute, time.Second).ShouldNot(HaveOccurred())
+				Expect(string(p.UID)).To(Equal(oldUID),
+					"router pod on %s should NOT have been restarted", p.Spec.NodeName)
+			}
```
zeeke: nit: what about simplifying like this:


``` suggestion
			currentPods, err := openperouter.RouterPods(cs)
			Expect(podNameToUIDs(currentPods)).To(Equal(podNameToUIDs(routerPods)))
```

```go

func podNameToUIDs(pods []*corev1.Pod) map[string]types.UID {
	res := make(map[string]types.UID, len(pods))
	for _, p := range pods {
		res[p.Spec.NodeName] = p.UID
	}
	return res
}

```

internal/controller/routerconfiguration/host_config.go
```diff
@@ -200,11 +199,22 @@ func restoreUnderlay(ctx context.Context, targetNamespace string) {
...
-// should be restarted instead of being reconfigured.
-func nonRecoverableHostError(e error) bool {
-	underlayExistsError := hostnetwork.UnderlayExistsError("")
-	return errors.As(e, &underlayExistsError)
+func setupUnderlayWithCleanup(ctx context.Context, targetNamespace string, underlayParams hostnetwork.UnderlayParams) error {
+	existing, err := hostnetwork.UnderlayInterfaces(targetNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to list existing underlay interfaces: %w", err)
+	}
+	if removedInterfaces := hostnetwork.RemovedUnderlayInterfaces(existing, underlayParams.UnderlayInterfaces); len(removedInterfaces) > 0 && len(removedInterfaces) == len(existing) {
```
zeeke: please, simplify the reading

```suggestion
    removedInterfaces := hostnetwork.RemovedUnderlayInterfaces(existing, underlayParams.UnderlayInterfaces); 
	if len(removedInterfaces) > 0 && len(removedInterfaces) == len(existing) {
```

zeeke (commented):
grout part looks good to me. 
The strategy "when a single underlay nic changes, remove all the underlay and reconfigure them" sounds a little brutal, but the implementation simplicity pays back.

only a few minor comments on my side

e2etests/tests/hostconfiguration.go
maiqueb: nice suggestion ! reads a lot better, and as a side-effect I got rid of a bunch of code. Tks !

internal/controller/routerconfiguration/host_config.go
maiqueb: Sure. Done.

e2etests/tests/hostconfiguration.go
```diff
@@ -431,14 +432,48 @@ var _ = ginkgo.Describe("Router Host configuration", func() {
...
+
+			ginkgo.By("restoring the underlay nic to a valid one")
+			resources.Underlays[0].Spec.Interfaces[0].NetworkDevice.InterfaceName = "toswitch1"
+			err = Updater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By("verifying nodes recover from degraded status")
+			for _, p := range routerPods {
+				Eventually(func(g Gomega) {
+					status, err := openperouter.GetNodeStatus(Updater.Client(), p.Spec.NodeName)
```
fedepaol: does it make sense to have a "NodeHasStatusCondition" function in the openperouter function and make this a one liner?

internal/controller/routerconfiguration/host_config.go
```diff
@@ -200,11 +199,24 @@ func restoreUnderlay(ctx context.Context, targetNamespace string) {
...
-func nonRecoverableHostError(e error) bool {
-	underlayExistsError := hostnetwork.UnderlayExistsError("")
-	return errors.As(e, &underlayExistsError)
+func setupUnderlayWithCleanup(ctx context.Context, targetNamespace string, underlayParams hostnetwork.UnderlayParams) error {
+	existing, err := hostnetwork.UnderlayInterfaces(targetNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to list existing underlay interfaces: %w", err)
+	}
+
+	removedInterfaces := hostnetwork.RemovedUnderlayInterfaces(existing, underlayParams.UnderlayInterfaces)
```
fedepaol: nit, non blocking. I'd move the cleaning logic to the calling site instead of having setupWithCleaning

internal/hostnetwork/underlay.go
```diff
@@ -57,17 +57,20 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
 		return fmt.Errorf("failed to check existing underlay interfaces: %w", err)
 	}
-	for _, name := range existingIfaces {
-		if !slices.Contains(params.UnderlayInterfaces, name) {
-			return UnderlayExistsError(fmt.Sprintf(
-				"existing underlay found: %s, new interfaces are %v", name, params.UnderlayInterfaces))
+	if removedInterfaces := RemovedUnderlayInterfaces(existingIfaces, params.UnderlayInterfaces); len(removedInterfaces) > 0 {
+		slog.InfoContext(ctx, "underlay interfaces changed, restoring old interfaces before setup",
+			"removed", removedInterfaces, "requested", params.UnderlayInterfaces)
+		if err := RestoreUnderlay(ctx, params.TargetNS); err != nil {
```
fedepaol: this will move all the interfaces to the original netns, causing disruption over those that are still in the list

e2etests/tests/hostconfiguration.go
maiqueb: Yes. I've added an helper for that, and re-used in across resiliency.go.

internal/hostnetwork/underlay.go
maiqueb: Done, now I only move back the interfaces which the user removed from the underlay.

It's a bit more work on the underlay deletion call site, but not that big a deal.

Let me know what you think.

maiqueb (commented):
@fedepaol please take another look

maiqueb:
@fedepaol PTAL. 

Question ... is this a feature ?... or a cleanup ?...

fedepaol:
> @fedepaol PTAL.
> 
> Question ... is this a feature ?... or a cleanup ?...

I would say framed the proper way it's a feature, because we don't kill the pod anymore while functionally nothing changes, we offer a bit more resiliency

internal/controller/routerconfiguration/host_config.go
```diff
@@ -195,16 +194,29 @@ func restoreUnderlay(ctx context.Context, targetNamespace string) {
...
 		slog.Warn("failed to remove underlay interfaces after underlay removal", "err", err)
 	}
 }
 
-// nonRecoverableHostError tells whether the router pod
-// should be restarted instead of being reconfigured.
-func nonRecoverableHostError(e error) bool {
-	underlayExistsError := hostnetwork.UnderlayExistsError("")
-	return errors.As(e, &underlayExistsError)
+func setupUnderlayWithCleanup(ctx context.Context, targetNamespace string, underlayParams hostnetwork.UnderlayParams) error {
```
fedepaol: sorry if I get back on this but this is confusing. This is called "with cleanup" but doesn't clean the underlay.

internal/controller/routerconfiguration/host_config.go
```diff
@@ -31,12 +31,12 @@ type KernelDatapathConfigurator struct {
...
 func (k *KernelDatapathConfigurator) Configure(ctx context.Context, config interfacesConfiguration) error { // nolint:gocognit
-	hasAlreadyUnderlay, err := hostnetwork.HasUnderlayInterface(config.targetNamespace)
+	currentUnderlayIfaces, err := hostnetwork.UnderlayInterfacesInNetNs(config.targetNamespace)
 	if err != nil {
 		return fmt.Errorf("failed to check if target namespace %s has underlay: %w", config.targetNamespace, err)
 	}
-	if hasAlreadyUnderlay && len(config.Underlays) == 0 {
-		restoreUnderlay(ctx, config.targetNamespace)
+	if len(currentUnderlayIfaces) > 0 && len(config.Underlays) == 0 {
+		restoreUnderlay(ctx, config.targetNamespace, currentUnderlayIfaces)
```
fedepaol: I think we should restore interfaces that are not in the list anymore, even if len > 0

fedepaol: also, I'd move the cleaning part of setupUnderlayWitCleanup here (or move this whole block below, probably even better because it's where we mess up with the host)

internal/controller/routerconfiguration/grout_config.go
```diff
@@ -30,13 +30,13 @@ func NewGroutConfigurator(groutSocketPath string) *GroutDatapathConfigurator {
...
-	hasAlreadyUnderlay, err := grout.HasUnderlayInterface(ctx, groutClient)
+	underlayIfaces, err := grout.UnderlayInterfaces(ctx, groutClient)
 	if err != nil {
 		return fmt.Errorf("failed to check if target namespace %s has underlay: %w", config.targetNamespace, err)
 	}
-	if hasAlreadyUnderlay && len(config.Underlays) == 0 {
+	if len(underlayIfaces) > 0 && len(config.Underlays) == 0 {
 		slog.InfoContext(ctx, "underlay removed, cleaning up grout underlay")
-		if err := grout.RestoreUnderlay(ctx, groutClient, config.targetNamespace); err != nil {
+		if err := grout.RestoreUnderlay(ctx, groutClient, config.targetNamespace, underlayIfaces); err != nil {
```
fedepaol, line 39: same, we should probably remove the delta.
Also, should we be removing all the vnis too?

internal/controller/routerconfiguration/host_config.go
maiqueb: done, removed this helper, and inlined its code on the caller.

internal/controller/routerconfiguration/grout_config.go
maiqueb: this is a shortcut for the scenario where the underlay is removed, and needs to be reset.

As we've discussed out of band, we decided to keep it.

internal/controller/routerconfiguration/host_config.go
maiqueb: this is a shortcut for the scenario where the underlay is removed, and needs to be reset.

As we've discussed out of band, we decided to keep it.

internal/controller/routerconfiguration/host_config.go
maiqueb: done

maiqueb: done

internal/controller/routerconfiguration/host_config.go
```diff
@@ -62,15 +62,9 @@ func (k *KernelDatapathConfigurator) Configure(ctx context.Context, config inter
...
-		return fmt.Errorf("failed to list existing underlay interfaces: %w", err)
-	}
-
-	removedInterfaces := hostnetwork.UnderlayInterfacesToRemove(existing, hostConfig.Underlay.UnderlayInterfaces)
-	if len(removedInterfaces) > 0 && len(removedInterfaces) == len(existing) {
-		slog.InfoContext(ctx, "all underlay interfaces removed, cleaning up VNIs before interface swap",
-			"removed", removedInterfaces, "requested", hostConfig.Underlay.UnderlayInterfaces)
+	if removeAllVNIs, err := areAllUnderlayInterfacesToBeRemoved(ctx, config, hostConfig); err != nil {
+		return err
+	} else if removeAllVNIs {
```
fedepaol: no else is required here

internal/controller/routerconfiguration/host_config.go
```diff
@@ -62,7 +62,17 @@ func (k *KernelDatapathConfigurator) Configure(ctx context.Context, config inter
 		return err
 	}
 
+	if removeAllVNIs, err := areAllUnderlayInterfacesToBeRemoved(ctx, config, hostConfig); err != nil {
+		return err
+	} else if removeAllVNIs {
+		if err := hostnetwork.RemoveAllVNIs(config.targetNamespace); err != nil {
```
fedepaol, line 73: drop a comment on why we are doing this

internal/hostnetwork/underlay.go
```diff
@@ -217,12 +243,12 @@ func restoreUnderlayWithHandles(ctx context.Context, fromNetNSPath string,
 	return restoreFn(ctx, fromNetNSHandle, defaultNetNSHandle, defaultNetNS)
 }
 
-// HasUnderlayInterface returns true if the given network
-// namespace already has a configured underlay interface.
-func HasUnderlayInterface(namespace string) (bool, error) {
+// UnderlayInterfacesInNetNs returns the list of interfaces currently configured
+// with the UnderlayGroupID in the target net ns
+func UnderlayInterfacesInNetNs(namespace string) (map[string]struct{}, error) {
```
fedepaol: having this and the version returning a slice with the same implementation is confusing, please consider using only one (and if we need the map because inner functions need the map, let's have a conversion function from slices to maps).

internal/controller/routerconfiguration/host_config.go
maiqueb: hm, yes it is ... without it, we lose the scope of the `removeAllVNIs` and `err` variables. 

What I can do instead is to invoke `areAllUnderlayInterfacesToBeRemoved` outside the if, thus having those variables in the scope of the `Configure` fn.

internal/hostnetwork/underlay.go
maiqueb: Done.

internal/controller/routerconfiguration/host_config.go
maiqueb: done

internal/controller/routerconfiguration/host_config.go
maiqueb: I chose to do that instead. Resolving.

fedepaol:
Thanks!

PR #549, opened 2026-07-06
docs: add DPDK-accelerated underlay enhancement

zeeke:
Proposes a new GroutPort mode for the UnderlayInterface union that binds an SR-IOV VF directly to grout via DPDK, replacing the current TAP+remote= mechanism on the underlay fast path. Defines VF selection (by PCI address, netdev name, or PF+VF index), inline IPAM for DPDK-bound interfaces, and the controller provisioning flow.

>
> /kind bug
> /kind cleanup
> /kind feature
/kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

Discuss the accelerated datapath for Grout, on the underlay side.

Acceleration on the host side (L2VNI and L3VNI) will be proposed later.

**Release note**:

```release-note
None
```

## Summary by CodeRabbit

- **New Features**
  - Added opt-in DPDK-accelerated underlay support through `grout` configuration.
  - Supports configurable MTU, receive queue count, and queue size.
  - Preserves and restores network device settings during setup and teardown.
  - Existing TAP-based underlay behavior remains the default.

- **Documentation**
  - Added configuration examples, validation rules, datapath limitations, error-handling guidance, and testing coverage.

### DIFF
--- a/enhancements/grout-dpdk-underlay.md
+++ b/enhancements/grout-dpdk-underlay.md
@@ -1,0 +1,339 @@
+# DPDK-Accelerated Underlay Ports for Grout
+
+## Summary
+
+Replace the TAP-with-`remote=` underlay port mechanism in grout with
+direct DPDK port attachment. After the
+[grout-dataplane](grout-dataplane.md) enhancement (M1),
+`configureUnderlayPort()` creates a `net_tap` PMD with `remote=<nic>`,
+which installs TC ingress rules to redirect packets between the underlay
+NIC and grout — adding kernel overhead on the underlay fast path. By
+binding a network device directly to a DPDK poll-mode driver
+(e.g. `vfio-pci`), grout can send and receive underlay traffic entirely
+in user-space, eliminating the kernel data path for VXLAN encap/decap
+and BGP-learned forwarding.
+
+The controller takes ownership of driver binding and IP address
+migration: it reads the IP configuration from the kernel netdev before
+binding the DPDK driver, applies it to the grout port, and restores it
+on teardown.
+
+This enhancement extends `NetworkDevice` with an optional `Grout` field.
+When `Grout` is set, the controller binds the device as a DPDK port
+instead of creating a TAP+`remote=` bridge. When `Grout` is absent, the
+existing TAP-based path is used (no behavioral change).
+
+## Motivation
+
+### Goals
+
+- **Remove the kernel from the underlay fast path.** The current
+  TAP+`remote=` approach routes every underlay packet through TC ingress
+  rules in the kernel. Direct DPDK port attachment eliminates this
+  overhead.
+- **Opt-in per interface.** Operators choose DPDK acceleration on a
+  per-interface basis by adding the `grout` field to a `NetworkDevice`
+  entry. Interfaces without it continue to use the TAP path.
+- **Automatic IP address migration.** The controller reads the IP
+  configuration from the kernel netdev before binding the DPDK driver
+  and applies it to the grout port via `grcli`. On teardown the
+  original driver and IP addresses are restored.
+- **Controller-managed driver binding.** The controller determines the
+  NIC type and binds the appropriate driver (`vfio-pci` for Intel,
+  namespace move for Mellanox/bifurcated). The user does not need to
+  pre-bind devices to a DPDK driver.
+- **Preserve the kernel-based TAP path as default.** A `NetworkDevice`
+  without `grout` behaves exactly as today — no migration required for
+  existing Underlay resources.
+
+### Non-Goals
+
+- **Managing VF creation on the host.** When SR-IOV is used, the user
+  creates VFs (`echo N > /sys/.../sriov_numvfs`) before the controller
+  consumes them.
+- **Workload-facing VF pairs for L2VNI HW acceleration.** Covered by
+  milestone M3a in [grout-dataplane](grout-dataplane.md).
+- **Replacing CNIDevice mode.** CNIDevice serves kernel-based NIC
+  sharing (macvlan/ipvlan). The `grout` field on `NetworkDevice` is for
+  DPDK-bound devices.
+
+## User Stories
+
+#### Story 1: High-Throughput Underlay
+As a network operator with DPDK-capable NICs, I want grout to attach
+a device directly via DPDK so that underlay traffic avoids the kernel
+entirely and achieves line-rate forwarding.
+
+#### Story 2: Opt-In Acceleration
+As an operator running grout, I want to add `grout: {}` to an existing
+`NetworkDevice` interface to switch it from TAP mode to DPDK mode
+without changing anything else in my Underlay spec.
+
+#### Story 3: Tuning Port Parameters
+As an operator, I want to set MTU, RX queue count, and queue size on
+a DPDK-bound underlay port to match my NIC capabilities and traffic
+profile.
+
+## Proposal
+
+### Overview
+
+The existing `NetworkDevice` type gains an optional `grout` field. When
+present, the controller binds the device as a DPDK port instead of
+creating a TAP+`remote=` bridge:
+
+| `grout` field | Behavior | IPAM | Datapath |
+|---------------|----------|------|----------|
+| Absent | Moves host device; grout creates TAP+`remote=` | Native (CIDR-derived) | TAP PMD (kernel TC redirect) |
+| Present | Binds device to DPDK driver; creates grout port | Scraped from kernel netdev | DPDK PMD (user-space) |
+
+The `grout` field is only valid when `--datapath=grout`. The controller
+rejects it in kernel datapath mode. A webhook validation must be
+implemented for this.
+
+### Device Selection
+
+The device is identified by the existing `interfaceName` field on
+`NetworkDevice`, which is the kernel netlink device name (e.g.
+`enp3s0f0v0`). The controller resolves the PCI address from sysfs at
+reconcile time via `/sys/class/net/<name>/device`.
+
+### Port Creation Flow
+
+1. **Resolve device** — Read the PCI address from
+   `/sys/class/net/<interfaceName>/device` symlink.
+2. **Save original state** — Read the IP addresses from the netlink
+   device and the current kernel driver name (from
+   `/sys/bus/pci/devices/<addr>/driver`). Write both to a per-device
+   state file at `/var/run/openperouter/grout/<pci-address>.json`.
+3. **Bind DPDK driver** — Determine the NIC type:
+   - *Non-bifurcated* (e.g. Intel): unbind from the current kernel driver
+     and bind to `vfio-pci`. This destroys the kernel netdev.
+   - *Bifurcated* (e.g. Mellanox `mlx5`): move the netlink device into
+     the perouter namespace. The kernel driver stays; DPDK shares the
+     device via the bifurcated model.
+4. **Create grout port** —
+   `grcli interface add port u_<name> devargs <pci> [mtu MTU] [rxqs N_RXQ] [qsize Q_SIZE]`
+   Options are appended only when set in `grout`.
+5. **Assign addresses** —
+   `grcli address add <cidr> iface u_<name>` for each saved IP address.
+6. **Kernel route for FRR** — add connected route on `main` so BGP
+   sessions transit grout (same as today).
+
+### API Examples
+
+##### NetworkDevice with DPDK acceleration (defaults)
+
+The device `enp3s0f0v0` must already exist as a kernel netdev with
+an IP address configured (e.g. via `ip addr add`).
+
+```yaml
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-dpdk
+spec:
+  asn: 64514
+  interfaces:
+    - type: NetworkDevice
+      networkDevice:
+        interfaceName: enp3s0f0v0
+        grout: {}
+  tunnelEndpoint:
+    cidrs:
+      - "100.65.0.0/24"
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+##### NetworkDevice with DPDK acceleration and port options
+
+```yaml
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-dpdk-tuned
+spec:
+  asn: 64514
+  interfaces:
+    - type: NetworkDevice
+      networkDevice:
+        interfaceName: enp3s0f0v0
+        grout:
+          mtu: 9000
+          rxQueues: 4
+          qSize: 1024
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+##### NetworkDevice without grout (TAP path, unchanged behavior)
+
+```yaml
+apiVersion: openpe.openperouter.github.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay-tap
+spec:
+  asn: 64514
+  interfaces:
+    - type: NetworkDevice
+      networkDevice:
+        interfaceName: enp3s0f0
+  neighbors:
+    - address: 192.168.1.1
+      asn: 65000
+```
+
+## Design Details
+
+### API Types
+
+```go
+// NetworkDevice moves an existing host network device into the router netns.
+type NetworkDevice struct {
+	// interfaceName is the name of the host network device to move into
+	// the router netns.
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9._-]*$`
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=15
+	// +required
+	InterfaceName string `json:"interfaceName,omitempty"`
+
+	// grout, when set, binds the device as a DPDK port instead of
+	// creating a TAP+remote= bridge. Only valid when --datapath=grout.
+	// +optional
+	Grout *GroutPortOptions `json:"grout,omitempty"`
+}
+
+type GroutPortOptions struct {
+	// +kubebuilder:validation:Minimum=68
+	// +kubebuilder:validation:Maximum=9702
+	MTU *int `json:"mtu,omitempty"`
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=64
+	RXQueues *int `json:"rxQueues,omitempty"`
+	// +kubebuilder:validation:Minimum=64
+	// +kubebuilder:validation:Maximum=32768
+	QSize *int `json:"qSize,omitempty"`
+}
+```
+
+The `UnderlayInterface` union is unchanged — no new discriminator
+variant is needed:
+
+```go
+// +union
+type UnderlayInterface struct {
+	// +kubebuilder:validation:Enum=NetworkDevice;CNIDevice
+	// +unionDiscriminator
+	Type          UnderlayInterfaceType `json:"type,omitempty"`
+	NetworkDevice *NetworkDevice        `json:"networkDevice,omitempty"`
+	CNIDevice     *CNIDevice            `json:"cniDevice,omitempty"`
+}
+```
+
+### Saved Device State
+
+Before binding the DPDK driver the controller writes a per-device state
+file to `/var/run/openperouter/grout/<pci-address>.json`:
+
+```json
+{
+  "pciAddress": "0000:03:02.0",
+  "netlinkName": "enp3s0f0v0",
+  "originalDriver": "ice",
+  "addresses": ["192.168.1.10/24", "fd00::10/64"]
+}
+```
+
+This file is read at teardown to restore the original driver and IP
+addresses. It is deleted after successful restoration.
+
+### Datapath Validation
+
+`KernelDatapathConfigValidator` is extended to reject `NetworkDevice`
+entries that have `grout` set.
+
+### Device Resolution
+
+The device is identified by `interfaceName` (the kernel netlink name).
+The controller resolves the PCI address at reconcile time:
+
+- Validate the device exists at `/sys/class/net/<interfaceName>`.
+- Read PCI address from the `/sys/class/net/<interfaceName>/device`
+  symlink (resolves to `/sys/bus/pci/devices/<addr>`).
+
+### Teardown
+
+On Underlay deletion or netns rebuild:
+
+1. `grcli interface del u_<name>` — removes the DPDK port.
+2. Read the saved state file from
+   `/var/run/openperouter/grout/<pci-address>.json`.
+3. **Restore driver** — Rebind the device to the original kernel driver
+   recorded in the state file (`echo <pci> > /sys/bus/pci/drivers/<orig>/bind`).
+   For bifurcated drivers, move the netlink device back to the host
+   namespace.
+4. **Restore IP addresses** — Re-apply the saved IP addresses to the
+   re-created kernel netdev.
+5. Delete the state file.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| Device not available (no kernel netdev) | Clear error at reconcile with device name |
+| No IP address configured on the device | The grout port is configured with zero IP addresses |
+| `vfio-pci` module not loaded | Check before binding; surface actionable error in status |
+| State file lost (e.g. `/var/run` cleared) | Log warning; operator must manually rebind driver and restore IP |
+| Driver rebind fails at teardown | Retry with backoff; leave state file for manual recovery |
+| Multiple Underlays claim the same device | `grcli interface add` fails with "device busy"; surfaced in status |
+
+### Test Plan
+
+- **E2E tests / Kind**: Against grout in test-mode (no hugepages),
+  verify port creation with `net_tap` devargs (no VF hardware in CI),
+  address assignment, and teardown.
+- **E2E tests / QEMU**: Deploy a cluster based on KVM / QEMU with emulated
+  SR-IOV NICs. Running the entire e2etest suite is hard, as the same clab
+  topology can't be implemented with VMs. A small set of test cases will be
+  implemented for this lane, using a simple FRR BGP peer in a container.
+- **Validation tests**: `grout` field rejected when grout disabled;
+  verify field validation ranges for MTU, RXQueues, QSize.
+
+## Alternatives
+
+### Alternative 1: Use CNIDevice with SR-IOV CNI
+
+Use existing `CNIDevice` mode with an `sriov` CNI plugin config.
+
+**Why not chosen:** The SR-IOV CNI moves a kernel netdev into a
+container namespace — it does not hand off to grout's DPDK port creation.
+IPAM via CNI is meaningless for DPDK-bound interfaces (no kernel netdev
+to assign the IP to).
+
+### Alternative 2: New GroutDevice union variant
+
+Add a `GroutDevice` discriminator to `UnderlayInterface` with its own
+selectors (PCI address, PF+VF index, netlink name).
+
+**Why not chosen:** The DPDK port is still a network device identified
+by its netlink name — the same selection mechanism `NetworkDevice` already
+uses. Adding a separate union variant duplicates device selection logic
+and forces operators to change `type:` when toggling between TAP and
+DPDK mode. An optional field on `NetworkDevice` is simpler: operators
+add or remove `grout: {}` without restructuring the spec.
+
+### Alternative 3: Inline IPAM in the API spec
+
+Carry IP addresses in the `Grout` struct (e.g. an `addresses` list)
+instead of reading them from the kernel.
+
+**Why not chosen:** The device must have a kernel netdev with an IP
+address *before* the controller binds it — reading the IP from the
+existing configuration is simpler, avoids duplication between the host
+config and the Underlay spec, and guarantees that the address is
+restored correctly on teardown.
--- a/enhancements/grout-dpdk-underlay.md
+++ b/enhancements/grout-dpdk-underlay.md
@@ -1,0 +1,382 @@
+# DPDK-Accelerated Underlay Ports for Grout
+
+## Summary
+
+Replace the TAP-with-`remote=` underlay port mechanism in grout with
+direct DPDK port attachment. After the
+[grout-dataplane](grout-dataplane.md) enhancement (M1),
+`configureUnderlayPort()` creates a `net_tap` PMD with `remote=<nic>`,
+which installs TC ingress rules to redirect packets between the underlay
+NIC and grout — adding kernel overhead on the underlay fast path. By
+binding a network device directly to a DPDK poll-mode driver
+(e.g. `vfio-pci`), grout can send and receive underlay traffic entirely
+in user-space, eliminating the kernel data path for VXLAN encap/decap
+and BGP-learned forwarding.
+
+The controller takes ownership of driver binding and IP address
+migration: it reads the IP configuration from the kernel netdev before
+binding the DPDK driver, applies it to the grout port, and restores it
+on teardown.
+
+This enhancement extends `NetworkDevice` with an optional `AcceleratedConfig` field.
+When `AcceleratedConfig` is set, the controller binds the device as a DPDK port
+instead of creating a TAP+`remote=` bridge. When `AcceleratedConfig` is absent, the
+existing TAP-based path is used (no behavioral change).
+
+## Motivation
+
+### Goals
+
+- **Remove the kernel from the underlay fast path.** The current
+  TAP+`remote=` approach routes every underlay packet through TC ingress
+  rules in the kernel. Direct DPDK port attachment eliminates this
+  overhead.
+- **Opt-in per interface.** Operators choose DPDK acceleration on a
+  per-interface basis by adding the `acceleratedConfig` field to a `NetworkDevice`
+  entry. Interfaces without it continue to use the TAP path.
+- **Automatic IP address migration.** The controller reads the IP
+  configuration from the kernel netdev before binding the DPDK driver
+  and applies it to the grout port via `grcli`. On teardown the
+  original driver and IP addresses are restored.
+- **Controller-managed driver binding.** The controller determines the
+  NIC type and binds the appropriate driver (`vfio-pci` for Intel,
+  namespace move for Mellanox/bifurcated). The user does not need to
+  pre-bind devices to a DPDK driver.
+- **Preserve the kernel-based TAP path as default.** A `NetworkDevice`
+  without `acceleratedConfig` behaves exactly as today — no migration required for
+  existing Underlay resources.
+
+### Non-Goals
+
+- **Managing VF creation on the host.** When SR-IOV is used, the user
+  creates VFs (`echo N > /sys/.../sriov_numvfs`) before the controller
+  consumes them.
+- **Workload-facing VF pairs for L2VNI HW acceleration.** Covered by
+  milestone M3a in [grout-dataplane](grout-dataplane.md).
+- **Replacing CNIDevice mode.** CNIDevice serves kernel-based NIC
+  sharing (macvlan/ipvlan). The `acceleratedConfig` field on `NetworkDevice` is for
+  DPDK-bound devices.
+
+## User Stories
+
+### Story 1: High-Throughput Underlay
+As a network operator with DPDK-capable NICs, I want grout to attach
+a device directly via DPDK so that underlay traffic avoids the kernel
+entirely and achieves line-rate forwarding.
+
+### Story 2: Opt-In Acceleration
+As an operator running grout, I want to enable DPDK acceleration
+on a selected underlay device.
+
+### Story 3: Tuning Port Parameters
+As an operator, I want to set RX queue count, and queue size on
+a DPDK-bound underlay port to match my NIC capabilities and traffic
+profile.
+
+## Proposal
+
+### Overview
+
+The existing `NetworkDevice` type gains an optional `acceleratedConfig` field. When
+present, the controller binds the device as a DPDK port instead of
+creating a TAP+`remote=` bridge:
+
+| `acceleratedConfig` field | Behavior | IPAM | Datapath |
+|---------------|----------|------|----------|
+| Absent | Moves host device; grout creates TAP+`remote=` | Native (CIDR-derived) | TAP PMD (kernel TC redirect) |
+| Present | Binds device to DPDK driver; creates grout port | Scraped from kernel netdev | DPDK PMD (user-space) |
+
+The `acceleratedConfig` field is only valid when `--datapath=grout`. The controller
+rejects it in kernel datapath mode. A webhook validation must be
+implemented for this.
+
+### Device Selection
+
+The device is identified by the existing `interfaceName` field on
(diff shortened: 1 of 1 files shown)


enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,278 @@
...
+
+- **Remove the kernel from the underlay fast path.** The current
+  TAP+`remote=` approach routes every underlay packet through TC ingress
+  rules in the kernel. Direct VF attachment eliminates this overhead.
+- **Expose VF selection in the API.** Users need a way to tell the
+  controller which VF to bind — by PCI address, or PF
+  name + VF index — without external CNI tooling.
+- **Support IPAM for DPDK-bound interfaces.** A VF bound to a DPDK
+  driver has no kernel netdev, so CNI IPAM plugins cannot assign an IP
+  to it. `GroutPort` carries inline IPAM applied via `grcli`.
```
rjarry: Slight precision needed: bifurcated DPDK drivers **do** have a kernel netdev which is used (required in fact) only for control plane operations (MTU change, link state, mac address filter, etc.). But the IP address must **not** be configured on it.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,278 @@
...
+
+1. `grcli interface del u_<name>` — removes the DPDK port.
+2. If the VF binds a bifurcated driver (e.g. `mlx5`) the VF returns
+   to the host namespace.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| VF not available (not created, already bound) | Clear error at reconcile with PCI address and resolution source |
```
rjarry: What component is expected to allocate the VFs? Should it be delegated to `nmstate`?

If so, who will take care of re-binding the device to `vfio-pci` (most DPDK drivers) or move the linux netdev to the router namespace (bifurcated drivers, e.g. mlx5)?

zeeke: >What component is expected to allocate the VFs? Should it be delegated to nmstate?

yes, either nmstate or any other configuration would be on the user. At least at this stage of the development.

>If so, who will take care of re-binding the device to vfio-pci (most DPDK drivers) or move the linux netdev to the router namespace (bifurcated drivers, e.g. mlx5)?

For `vfio-pci` driver, the controller will do nothing on the VF during setup/teardown. Just passing the PCIAddress to a `grcli interaface add ...`

For bifurcated drivers, the controller will also move the netlink in and out of the perouter namespace during setup/teardown

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,278 @@
...
+
+After the controller-provisioned-underlay-interfaces enhancement, the
+`UnderlayInterface` union has two modes: `NetworkDevice` and `CNIDevice`.
+This enhancement adds a third: `GroutPort`.
+
+| Mode | Behavior | IPAM | Datapath |
+|------|----------|------|----------|
+| `NetworkDevice` | Moves host device; grout creates TAP+`remote=` | Native (CIDR-derived) | TAP PMD (kernel TC redirect) |
+| `CNIDevice` | Invokes CNI plugin | Delegated to CNI | Kernel or TAP PMD |
+| `GroutPort` | Binds a VF to grout as a DPDK port | Inline in spec | DPDK PMD (user-space) |
```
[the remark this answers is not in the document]
maiqueb: I think gemini hallucinated. 

Let's align everything w/ the `Device` suffix - please call this new attribute `GroutDevice`.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,281 @@
...
+  sharing (macvlan/ipvlan). GroutPort is for DPDK-bound VFs.
+
+## User Stories
+
+#### Story 1: High-Throughput Underlay
+As a network operator with SR-IOV-capable NICs, I want grout to attach
+a VF directly via DPDK so that underlay traffic avoids the kernel
+entirely and achieves line-rate forwarding.
+
+#### Story 2: VF Selection by PCI Address
```
fedepaol: why a user would want this if they have  story 3? Also, the pci address is likely to be different on each node I think

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,281 @@
...
+
+After the controller-provisioned-underlay-interfaces enhancement, the
+`UnderlayInterface` union has two modes: `NetworkDevice` and `CNIDevice`.
+This enhancement adds a third: `GroutPort`.
+
+| Mode | Behavior | IPAM | Datapath |
+|------|----------|------|----------|
+| `NetworkDevice` | Moves host device; grout creates TAP+`remote=` | Native (CIDR-derived) | TAP PMD (kernel TC redirect) |
+| `CNI` | Invokes CNI plugin | Delegated to CNI | Kernel |
+| `GroutPort` | Binds a VF to grout as a DPDK port | Inline in spec | DPDK PMD (user-space) |
```
fedepaol: I feel like groutport is a bit generic. We should mention it's a sriov vf, and optionally that it relates to grout. Something like acceleratedVF? FastDatapathVF? dpdkVF?

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,281 @@
...
+   `/sys/class/net/<pfName>/device/virtfn<vfIndex>` symlink. The VF must
+   already be bound to a DPDK-compatible driver (e.g. `vfio-pci`) or use
+   a bifurcated driver (e.g. `mlx5`); driver binding is the
+   responsibility of the user or SR-IOV device plugin (see Non-Goals).
+2. Moves the VF netlink to the perouter namespace, when dealing with
+   bifurcated PMD drivers (e.g. `mlx5`)
+3. **Create grout port** —
+   `grcli interface add port u_<name> devargs <pci> [mtu MTU] [rxqs N_RXQ] [qsize Q_SIZE]`
+   Options are appended only when set in `portOptions`.
+4. **Assign addresses** —
```
fedepaol: how are addresses assigned? The proposal doesn't say that

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,281 @@
...
+
+### VF Resolution
+
+Operates on sysfs:
+
+- **PCIAddress**: validate format, check `/sys/bus/pci/devices/<addr>`.
+- **PFVFIndex**: read symlink `/sys/class/net/<pf>/device/virtfn<idx>`,
+  extract PCI address from target path.
+
+*Optional*: The resolved PCI address is stored in Underlay status for observability.
```
fedepaol: we don't have underlay status

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,281 @@
...
+// +kubebuilder:validation:XValidation:rule="!has(self.vfIndex) || has(self.pfName)",message="pfName is required when vfIndex is set"
+type GroutPortConfig struct {
+	// +kubebuilder:validation:Pattern=`^[0-9a-fA-F]{4}:[0-9a-fA-F]{2}:[0-9a-fA-F]{2}\.[0-7]$`
+	PCIAddress *string            `json:"pciAddress,omitempty"`
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z][a-zA-Z0-9._-]*$`
+	// +kubebuilder:validation:MaxLength=15
+	PFName      *string           `json:"pfName,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	VFIndex     *int              `json:"vfIndex,omitempty"`
+	IPAM        GroutPortIPAM     `json:"ipam"`
```
fedepaol: do we envision allowing dhcp too?

maiqueb: integrating grout w/ any CNI based approach will not be straightforward. 

Grout takes away the IPs to the user space, which will cause CNI CHECK to fail.

This integration deserves - IMHO - a separate enhancement (or a new chapter focusing on the grout integration).

maiqueb (approved):

enhancements/grout-dpdk-underlay.md
zeeke: In real applications, this lets you bind PFs without needing to use SR-IOV or VFs.

for testing purposes, this will become useful when we'll deal with qEMU environment.

enhancements/grout-dpdk-underlay.md
zeeke: It can also be a Physical Function. 
Besides that, I'm open to suggestions

fedepaol: but the enhancement describes vfs only, mind expanding then?

enhancements/grout-dpdk-underlay.md
fedepaol: > In real applications, this lets you bind PFs without needing to use SR-IOV or VFs.

not sure I understood what you mean by "needing to user sr-iov"

> 
> for testing purposes, this will become useful when we'll deal with qEMU environment.

enhancements/grout-dpdk-underlay.md
zeeke: That's a good point, and I'm still in the process of defining it. 
My current ideas are:
a. put an `GroutDevice.Addresses map[string]string[]` map array with all the addresses to add to the port for each node (the map is `nodeName` -> `list of addresses`)

b. add a GroutDevice.CIDR field and use the nodeIndex to assign IP Address.

c. The user puts the addresses on the network device and the perouter reads them. This is my favorite, but it also implies
   c1. The perouter takes care of binding/unbinding the driver of the device (i.e. an Intel device is configured by the user as a netlink device, then the controller saves the addresses to a file and binds the vfio-pci driver)
  c2. The controller needs to store the IP addresses in some persistence layer, make the SetupUnderlay a stateful operation

enhancements/grout-dpdk-underlay.md
zeeke: ok, removing

enhancements/grout-dpdk-underlay.md
zeeke: > > In real applications, this lets you bind PFs without needing to use SR-IOV or VFs.
> 
> not sure I understood what you mean by "needing to user sr-iov"
> 
> > for testing purposes, this will become useful when we'll deal with qEMU environment.

A DPDK application can run on a Physical Function by using its PCI address. The fact that the SR-IOV technology has been invented is to let multiple DPDK application use the same NIC, each of them using a different VF. But it's not mandatory.  

Also, there are NICs that does not support SR-IOV (i.e. VFs) but can be used by Grout and DPDK.
@rjarry do you mind checking this last statement?

enhancements/grout-dpdk-underlay.md
zeeke: updated the description

enhancements/grout-dpdk-underlay.md
fedepaol: fine, but then we need to reword where we talk about vfs because it's not really only about vfs.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,359 @@
...
+#### Story 2: VF Selection by PCI Address
+As an operator who pre-provisions VFs, I want to specify the VF's PCI
+address in the Underlay spec so the controller binds the correct device.
+
+#### Story 3: VF Selection by PF + Index
+As an operator on a multi-NIC node, I want to specify the PF name and VF
+index so that the controller resolves the correct VF without requiring
+me to look up PCI addresses.
+
+#### Story 4: VF Selection by Netlink Name
```
fedepaol: is this vf only?

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+## Design Details
+
+### API Types
+
+```go
+// +union
+type UnderlayInterface struct {
+	// +kubebuilder:validation:Enum=NetworkDevice;CNI;GroutDevice
```
fedepaol: wrong indentation? github is showing this differently

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,359 @@
...
+
+After the controller-provisioned-underlay-interfaces enhancement, the
+`UnderlayInterface` union has two modes: `NetworkDevice` and `CNIDevice`.
+This enhancement adds a third: `GroutDevice`.
+
+| Mode | Behavior | IPAM | Datapath |
+|------|----------|------|----------|
+| `NetworkDevice` | Moves host device; grout creates TAP+`remote=` | Native (CIDR-derived) | TAP PMD (kernel TC redirect) |
+| `CNIDevice` | Invokes CNI plugin | Delegated to CNI | Kernel |
+| `GroutDevice` | Binds a network device to grout as a DPDK port | Scraped from kernel netdev | DPDK PMD (user-space) |
```
fedepaol: I am having second thoughs about this. We are mixing how we are selecting the interface and what we are going to do with them.
Basically, when we run in grout mode, the differences between networkdevice and groutdevice.netlink name is only how we are going to operate it.

I wonder if the selection mechanism should be part of networkdevice in any case, with networkdevice being expanded as what you are suggesting for groutdevice, with an underlayacceleratedmode knob that tells how to use it (with the proper validation of what can and can't be done on each mode).

enhancements/grout-dpdk-underlay.md
zeeke: fixing

enhancements/grout-dpdk-underlay.md
zeeke: rewording

enhancements/grout-dpdk-underlay.md
zeeke: If I got your idea, it would become:
```
Underlay
 .Spec
  .NetworkDevice
     // selects the device by netlink name. kernel mode stays like it is right now. 
     // grout mode would create a TAP device over it unless .AcceleratedDatapath.Enabled is active)
    .InterfaceName     

     // select the device by PF netlink name and VF index number. 
     // Once the device is selected, kernel and grout mode acts as per InterfaceName
    .PFName, .VFIndex

     // select the device via its PCI address. in kernel mode, it's mandatory that 
     //  the device binds a kernel driver. in grout mode, it can even be a DPDK driver. 
     //(though if it's bound to a kernel driver, the controller can steal the addresses)
    .PCIAddress         

   .AcceleratedDatapath   
       .Enabled        // Valid only if in grout mode. if true, create grout PCI ports instead of TAP devices
       .MTU, RXQueues, QSize // tune grout interface creation
```

I don't see any advantage or drawback in moving to this format.

zeeke: @maiqueb, any thoughts on these API shapes?

maiqueb: shouldn't `PFName` and `VFIndex` be attributes of `AcceleratedDatapath` ? 

also, I think the enabled attribute redundant - if you're specifying the `AcceleratedDatapath` configuration, then, you want it enabled. If you don't, well ... you don't want it.

zeeke: > shouldn't PFName and VFIndex be attributes of AcceleratedDatapath ?

it selects a device, so it can also be used for the kernel datapath, in case we want to use a VF. But it's probably a scenario that has absolutely no meaning in the real world.

What if we start with just the netlink name only (maybe specified via .NetworkDevice.InterfaceName) ?

fedepaol: in general, the semantics of networkdevice is "I want to select this device"
on the other hand, the previous version had the device selection semantic (pci address, netlink name, etc) and also the implicit semantic of doing acceletared datapath for that.

My proposal was to keep the concerns separate, the selection on one side and the acceleration on the other.

enhancements/grout-dpdk-underlay.md
rjarry: Sorry I had missed this notification. Yes it would make sense you allow selecting interfaces by explicit PCI address for qemu testing purposes (qemu igc emulated sriov devices do not handle VF to VF communication unless the same vlan id is explicitly configured on the *two* VFs hardware filters (which grout does not do). Since we need one of the VFs to be a "trunk", we cannot use that solution to run virtualized tests.

The alternative solution is to use plain virtio interfaces and give them to grout as PFs. Then plug the host side tap interfaces to a Linux bridge that will perform the vlan switching (emulating the internal igc VF switch).

enhancements/grout-dpdk-underlay.md
zeeke: I update the doc to meet this new approach. I dropped the PCIAddress and PFName+VFIndex device selectors. 

I think this is a way simpler initial approach. the user has to put the ip address on the netlink device, as he would for the kernel datapath.

If the `Underlay.Spec.Interfaces[].NetworkDevice.Grout` field is not nil, the controller knows it has to set up a DPDK port.

I chose `Grout` as the field name because `AcceleratedDatapath` is too generic. subfields there are meaningful to grout, and other accelerated implementations (if there will be any) will probably need other knobs.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+## Design Details
+
+### API Types
+
+```go
+// NetworkDevice moves an existing host network device into the router netns.
+type NetworkDevice struct {
+	// interfaceName is the name of the host network device to move into
```
fedepaol: nit: gh is highlighting the indentatio, not sure what it means

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,339 @@
...
+	InterfaceName string `json:"interfaceName,omitempty"`
+
+	// grout, when set, binds the device as a DPDK port instead of
+	// creating a TAP+remote= bridge. Only valid when --datapath=grout.
+	// +optional
+	Grout *GroutPortOptions `json:"grout,omitempty"`
+}
+
+type GroutPortOptions struct {
+	// +kubebuilder:validation:Minimum=68
```
fedepaol: same here

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,339 @@
...
+	// +kubebuilder:validation:Maximum=9702
+	MTU *int `json:"mtu,omitempty"`
```
fedepaol: can you document the fields?

enhancements/grout-dpdk-underlay.md
fedepaol: Fair enough, I like this more

fedepaol:
A few very small nits, but LGTM

fedepaol (approved):

enhancements/grout-dpdk-underlay.md
fedepaol: @maiqueb please have a look, I think it's reasonable

enhancements/grout-dpdk-underlay.md
zeeke: fixed the indentation replacing tabs with 2 spaces

enhancements/grout-dpdk-underlay.md
zeeke: added doc. mostly taken from https://www.mankier.com/1/grcli-interface

@rjarry can you review it?

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,358 @@
...
+- **Managing VF creation on the host.** The user creates VFs 
+  (`echo N > /sys/.../sriov_numvfs`) before the controller consumes them.
+- **Workload-facing VF pairs for L2VNI HW acceleration.** Covered by
+  milestone M3a in [grout-dataplane](grout-dataplane.md).
+- **Replacing CNIDevice mode.** CNIDevice serves kernel-based NIC
+  sharing (macvlan/ipvlan). GroutDevice is for DPDK-bound VFs.
+
+## User Stories
+
+#### Story 1: High-Throughput Underlay
```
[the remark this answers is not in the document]
maiqueb: yes, this should be using level 3 headings (remove one hash).

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+As a network operator with DPDK-capable NICs, I want grout to attach
+a device directly via DPDK so that underlay traffic avoids the kernel
+entirely and achieves line-rate forwarding.
+
+#### Story 2: Opt-In Acceleration
+As an operator running grout, I want to add `grout: {}` to an existing
+`NetworkDevice` interface to switch it from TAP mode to DPDK mode
+without changing anything else in my Underlay spec.
```
maiqueb: this user story gets into the "how". 

Let's describe what we want (have per device DPDK mode via the underlay spec) instead. Or something similar.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+
+#### Story 3: Tuning Port Parameters
+As an operator, I want to set MTU, RX queue count, and queue size on
+a DPDK-bound underlay port to match my NIC capabilities and traffic
+profile.
```
maiqueb: these params could be useful for other device types (maybe we can exclude CNI from it, since I assume each plugin can allow mtu configuration).

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Proposal
+
+### Overview
+
+The existing `NetworkDevice` type gains an optional `grout` field. When
```
maiqueb: Could this mechanism might be useful / used in the future for other attachment types ? 

Can we use any other name ? I am trying to avoid coupling the API with a specific implementation. Having a more generic name could allow the attribute to be re-used in the future. Maybe `userSpaceConfig`, or `accelerationConfig`.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,358 @@
...
+     `/sys/bus/pci/devices/<addr>/net/`.
+   - `pfName` + `vfIndex`: read symlink
+     `/sys/class/net/<pfName>/device/virtfn<vfIndex>`, extract PCI
+     address, then look up netlink name as above.
+   - `netlinkName`: look up the PCI address via
+     `/sys/class/net/<name>/device` symlink.
+2. **Save original state** — Read the IP addresses from the netlink
+   device and the current kernel driver name (from
+   `/sys/bus/pci/devices/<addr>/driver`). Write both to a per-device
+   state file at `/var/run/openperouter/grout/<pci-address>.json`.
```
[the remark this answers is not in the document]
maiqueb: this is needed IMHO.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,339 @@
...
+reconcile time via `/sys/class/net/<name>/device`.
+
+### Port Creation Flow
+
+1. **Resolve device** — Read the PCI address from
+   `/sys/class/net/<interfaceName>/device` symlink.
+2. **Save original state** — Read the IP addresses from the netlink
+   device and the current kernel driver name (from
+   `/sys/bus/pci/devices/<addr>/driver`). Write both to a per-device
+   state file at `/var/run/openperouter/grout/<pci-address>.json`.
```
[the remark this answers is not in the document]
maiqueb: this one also sounds required.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+  // +required
+  InterfaceName string `json:"interfaceName,omitempty"`
+
+  // grout, when set, binds the device as a DPDK port instead of
+  // creating a TAP+remote= bridge. Only valid when --datapath=grout.
+  // +optional
+  Grout *GroutPortOptions `json:"grout,omitempty"`
+}
+
+type GroutPortOptions struct {
```
maiqueb: I would not use the grout prefixes.

Plus, maybe it makes sense to also allow configuring the `tx` counterpart - `txQueues` and `txQSize`.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+  // +kubebuilder:validation:Maximum=64
+  // +optional
+  RXQueues *int `json:"rxQueues,omitempty"`
+  // qSize is the descriptor ring size for each receive queue. Larger
+  // rings absorb traffic bursts at the cost of memory.
+  // +kubebuilder:validation:Minimum=64
+  // +kubebuilder:validation:Maximum=32768
+  // +optional
+  QSize *int `json:"qSize,omitempty"`
+}
```
maiqueb, line 234: maybe we should take the opportunity to also define L2 Addressing and Filtering attributes.

Given PE routers interact heavily with routing protocols (OSPF, IS-IS, BGP), L2 hardware filtering could be useful.

Maybe the following attributes: 
- macAddress
- promiscuous
- allMulticast

This might be useful because routing protocols rely heavily on multicast (e.g., OSPF uses 224.0.0.5/6, IS-IS uses specific L2 multicasts). 

If the PMD drops these at the hardware level because the exact MAC isn't programmed into the NIC's filter, routing adjacencies will fail.

I'm also fine to postpone it until actually needed.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+addresses. It is deleted after successful restoration.
+
+### Datapath Validation
+
+`KernelDatapathConfigValidator` is extended to reject `NetworkDevice`
+entries that have `grout` set.
+
+### Device Resolution
+
+The device is identified by `interfaceName` (the kernel netlink name).
```
maiqueb, line 295: this makes me think if the `Saved Device State` should be indexed by this attribute instead. 

like, align everywhere what's the attribute to index - iface name / pci address / ...

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+  Type          UnderlayInterfaceType `json:"type,omitempty"`
+  NetworkDevice *NetworkDevice        `json:"networkDevice,omitempty"`
+  CNIDevice     *CNIDevice            `json:"cniDevice,omitempty"`
+}
+```
+
+### Saved Device State
+
+Before binding the DPDK driver the controller writes a per-device state
+file to `/var/run/openperouter/grout/<pci-address>.json`:
```
maiqueb: is `/var/run/openperouter/grout/` a volume already ? We must ensure this persists across controller and router restarts.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,358 @@
...
+
+On Underlay deletion or netns rebuild:
+
+1. `grcli interface del u_<name>` — removes the DPDK port.
+2. Read the saved state file from
+   `/var/run/openperouter/grout/<pci-address>.json`.
+3. **Restore driver** — Rebind the device to the original kernel driver
+   recorded in the state file (`echo <pci> > /sys/bus/pci/drivers/<orig>/bind`).
+   For bifurcated drivers, move the netlink device back to the host
+   namespace.
```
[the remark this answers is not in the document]
maiqueb: @zeeke could you re-phrase it, with this in mind ?

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+5. Delete the state file.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| Device not available (no kernel netdev) | Clear error at reconcile with device name |
+| No IP address configured on the device | The grout port is configured with zero IP addresses |
+| `vfio-pci` module not loaded | Check before binding; surface actionable error in status |
+| State file lost (e.g. `/var/run` cleared) | Log warning; operator must manually rebind driver and restore IP |
```
maiqueb: how will the operator know to which IP to use ? Can it be recovered from elsewhere ?...

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+| Driver rebind fails at teardown | Retry with backoff; leave state file for manual recovery |
+| Multiple Underlays claim the same device | `grcli interface add` fails with "device busy"; surfaced in status |
```
maiqueb, line 334: we can't have multiple underlays in the same node, can we ? I think this is a non-issue.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+
+### Test Plan
+
+- **E2E tests / Kind**: Against grout in test-mode (no hugepages),
+  verify port creation with `net_tap` devargs (no VF hardware in CI),
+  address assignment, and teardown.
+- **E2E tests / QEMU**: Deploy a cluster based on KVM / QEMU with emulated
+  SR-IOV NICs. Running the entire e2etest suite is hard, as the same clab
+  topology can't be implemented with VMs. A small set of test cases will be
+  implemented for this lane, using a simple FRR BGP peer in a container.
```
maiqueb, line 344: this will be hard, but I agree we can't rely on actual HW for these tests.

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+| Risk | Mitigation |
+|------|------------|
+| Device not available (no kernel netdev) | Clear error at reconcile with device name |
+| No IP address configured on the device | The grout port is configured with zero IP addresses |
+| `vfio-pci` module not loaded | Check before binding; surface actionable error in status |
+| State file lost (e.g. `/var/run` cleared) | Log warning; operator must manually rebind driver and restore IP |
+| Driver rebind fails at teardown | Retry with backoff; leave state file for manual recovery |
+| Multiple Underlays claim the same device | `grcli interface add` fails with "device busy"; surfaced in status |
+
+### Test Plan
```
maiqueb, line 336: I believe we should also test the resiliency of the solution, for instance, checking if we can recover from losing the grout state file (this iface has these IPs, and this PCI addr).

enhancements/grout-dpdk-underlay.md
```diff
@@ -0,0 +1,346 @@
...
+## Alternatives
+
+### Alternative 1: Use CNIDevice with SR-IOV CNI
+
+Use existing `CNIDevice` mode with an `sriov` CNI plugin config.
+
+**Why not chosen:** The SR-IOV CNI moves a kernel netdev into a
+container namespace — it does not hand off to grout's DPDK port creation.
+IPAM via CNI is meaningless for DPDK-bound interfaces (no kernel netdev
+to assign the IP to).
```
maiqueb, line 359: so I hear that this alternative is just not feasible ?

maiqueb (changes requested):
Some questions.

enhancements/grout-dpdk-underlay.md
zeeke: MTU is kind of particular, and I agree it can be relevant for other device types. But RX queues and queue size are not. 
I'd drop the MTU field and keep the others under the same field grout {}.

Does it look good?

maiqueb: if you're certain you don't need to configure the MTU of the grout attachment, then by all means, remove it.

enhancements/grout-dpdk-underlay.md
zeeke: In my opinion, it's very unlikely to have another accelerated datapath other than Grout. And even if it exists in the future, I don't know if the `grout` subfields will have any meaning. So `acceleratedConfig`  might contain something meaningless to the new implementation.

BTW, if you have any preference in the name, I'm happy to follow that.

maiqueb: @fedepaol thoughts ? I'd really like to avoid having the name of another component in our public facing API.

enhancements/grout-dpdk-underlay.md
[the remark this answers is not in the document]
zeeke: are you sure? do we handle multiple Underlays host configuration in parallel for the kernel datapath? 
also, do we need to handle the case two different Underlays objects target the same device? Isn't it handled by webhooks?

enhancements/grout-dpdk-underlay.md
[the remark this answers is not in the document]
zeeke: this is valid. we can't save the device state by its pci_address.
We must use `Underlay.spec.interfaces[].networkDevice.interfaceName`
updating

enhancements/grout-dpdk-underlay.md
zeeke: changing to `PortOption`. 
for some reason, txQueues and txQSize are not valid arguments for `grcli interface add port`
https://www.mankier.com/1/grcli-interface
```
interface add port NAME devargs DEVARGS [[promisc (on|off)] [mtu MTU] ([vrf VRF]|[domain DOMAIN]) [description DESCR] [mac MAC] [rxqs N_RXQ] [qsize Q_SIZE] (up|down)]
```

enhancements/grout-dpdk-underlay.md
zeeke: madAddress and promiscuous are supported by `grcli interface add port`  https://www.mankier.com/1/grcli-interface 
So, we can easily add them now. 

`allMulticast` is not supported at the moment, I'd wait for a real need for that

enhancements/grout-dpdk-underlay.md
zeeke: I'm not sure to get this, but the filepath should contains the interfaceName only. not the PCI address. 
I'm changing that

enhancements/grout-dpdk-underlay.md
zeeke: yes, it needs to be a volume that survives perouter crashes/restart. I considered that an implementation, but it's worth mentioning in the enhancement

enhancements/grout-dpdk-underlay.md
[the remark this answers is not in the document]
maiqueb: > are you sure? do we handle multiple Underlays host configuration in parallel for the kernel datapath? also, do we need to handle the case two different Underlays objects target the same device? Isn't it handled by webhooks?

you're right to push back - we can't have multiple underlays targeting the same node. My bad on not realizing the implications of this comment.

enhancements/grout-dpdk-underlay.md
zeeke: the "operator" here refers to the user, the one who configured the cluster and the perouter deployment. he knew the IP address at the deployment time, so he could still get them from somewhere.

maiqueb: the "user" could actually be an automation.

IMHO, we need to highlight this in the documentation - the IP address for this attachment must be persisted somewhere , or be known to whoever setup the cluster.

Not sure where in the enhancement to actually add that info.

enhancements/grout-dpdk-underlay.md
zeeke: I think that state file is our last lifeline. I think there is no other way to recover the IP addresses.

what resilience case did you have in mind?

maiqueb: I just wanted to understand how could we ever recover from this scenario. 

I might be asking for too much, given that in non-grout mode we are in the same situation (can't recover the IP if we ever delete it).

enhancements/grout-dpdk-underlay.md
zeeke: - we would have to bind the right driver anyway, as the sriov-cni assumes the driver is already binded.
- In case of vfio-pci, we wouldn't know where the user should put the IP addresses of the port. 

so, I think it would create more problem than solutions

maiqueb: yes, I agree. I am pretty much asking if it should even be mentioned as an alternative. Because I don't see any good in it :)

enhancements/grout-dpdk-underlay.md
fedepaol: the problem is, at this point in time we don't know. The name is the name of the datapath implementation (similar to "kernel") and not a component, doesn't sound a big deal to me.
 
What we risk is having duplicated fields if:

- another datapath implementation comes
- the different datapath has those 3 fields, with the same semantic and meaning

all of this seems quite unlikely to me, and if it ever comes, we can have a new generic field and start the deprecation path for the grout one.

But, for the same reason (we won't have another datapath rather than grout, and if so there won't be overlap) I don't see any issue in naming it `fastDatapathConfig` or something like that.

enhancements/grout-dpdk-underlay.md
zeeke: but is it any different than kernelmode? 
For NetworkDevice and kernel datapath, the user (or any automation on behalf of him), sets the ip addresses on the underlay network device. Once the user creates an Underlay object targeting that netlink, the interface disappears from the host, and it comes back when the Underlay object is deleted. 

The same interaction is supposed to happen for grout. The fact that, under the hood, the controller strips the addresses out of the netlink to give them to grout, then saves them to a file, is a matter of implementation, isn't it?

If there are any details we want to nail down at this stage, I can add it to this enhancement

enhancements/grout-dpdk-underlay.md
zeeke: yep, in non-grout mode we are in a similar situation, though the controller does not strip the address out of the netlink, so it's less likely to lose the addresses in oblivion.

A similar problem is handled in e2etest by the check_veths utility.

enhancements/grout-dpdk-underlay.md
zeeke: it sounds good to me
going for `fastDatapathConfig` ? `acceleratedConfig` ?

zeeke:
Renamed  `Underlay.Spec.Interfaces[].NetworkDevice.Grout` to `Underlay.Spec.Interfaces[].NetworkDevice.AcceleratedConfig`.

@fedepaol @maiqueb please take a nother look when you have time

enhancements/grout-dpdk-underlay.md
maiqueb: Let's wait for @fedepaol s' oppinion. 

I am OK with deferring adding *any* of these attributes - we can add them later, once a use case that requires them appears.

enhancements/grout-dpdk-underlay.md
maiqueb: I do think we should try to account for that scenario. 

At the very least, by recording the limitation (both in our docs, and in an issue) for our future, smarter selves to address.

maiqueb (approved):
Thank you.

enhancements/grout-dpdk-underlay.md
zeeke: Sounds good. I opened an issue to track this problem
- https://github.com/openperouter/openperouter/issues/684

enhancements/grout-dpdk-underlay.md
christophefontaine: As of today, all multi is always enabled for all ports: https://github.com/DPDK/grout/blob/845c2591355715964136fc7235264a87017ac749/modules/infra/control/port.c#L383 

As not all drivers implement multicast filtering, and to avoid having to validate multiple code path / behaviors, we decided to always enable `all multi`.

fedepaol: FWIW, I agree with postponing if we'll ever need them

fedepaol:
Merging, thanks!

PR #564, opened 2026-07-10
fix(grout): disable rp_filter on grout ports

zeeke:
Grout creates NOARP kernel interfaces (u_<port>) for FRR's zebra daemon. BGP packets leave through the `main` TAP interface but return on the `u_<port>` interface. When rp_filter is enabled (as in OpenShift, where the default is set to 1), the kernel drops these return packets due to the asymmetric path, breaking BGP sessions.

Disable rp_filter on the underlay interfaces to allow this traffic.

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

**Release note**:

```release-note
Disable rp_filter sysctl on grout ports
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved networking reliability for passthrough and underlay interfaces by adjusting reverse-path filtering behavior.
  * Enhanced end-to-end test diagnostics with kernel message collection when failures occur.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -258,6 +258,12 @@ jobs:
           sudo apt-get update
           sudo apt-get install linux-modules-extra-$(uname -r)
 
+      # rp_filter is off on CI runners; enable it so kind nodes inherit a production-like default
+      - name: Enable rp_filter
+        run: |
+          sudo sysctl -w net.ipv4.conf.default.rp_filter=1
+          sudo sysctl -w net.ipv4.conf.default.log_martians=1
+
       - name: Disable containerd image store
         run: |
           sudo bash -c 'cat > /etc/docker/daemon.json << EOF
@@ -340,6 +346,12 @@ jobs:
         if: ${{ failure() }}
         run:
           make kind-export-logs
+
+      # Needed to debug kernel "martian source" packets
+      - name: Collect dmesg logs
+        if: ${{ failure() }}
+        run: |
+          sudo dmesg -T > /tmp/kind_logs/dmesg.log
 
       - name: Collect veth monitor logs
         if: ${{ failure() }}
--- a/internal/grout/passthrough.go
+++ b/internal/grout/passthrough.go
@@ -11,6 +11,8 @@ import (
 	"net"
 
 	"github.com/openperouter/openperouter/internal/hostnetwork"
+	"github.com/openperouter/openperouter/internal/netnamespace"
+	"github.com/openperouter/openperouter/internal/sysctl"
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 )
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -258,6 +258,13 @@ jobs:
           sudo apt-get update
           sudo apt-get install linux-modules-extra-$(uname -r)
 
+      # rp_filter is off on CI runners; enable it so kind nodes inherit a production-like default
+      - name: Enable rp_filter
+        run: |
+          sudo sysctl -w net.ipv4.conf.all.rp_filter=1
```
maiqueb: I'm confused - your commit msg says:
>  conf/all is set to 0 because the kernel uses max(all, iface) — a non-zero "all" would override the per-interface disable.

which is the opposite of what I see here.

internal/grout/passthrough.go
```diff
@@ -53,6 +55,18 @@ func SetupPassthrough(ctx context.Context, client *Client, params hostnetwork.Pa
 		return fmt.Errorf("failed to ensure IPs to grout port: %w", err)
 	}
 
+	if err := netnamespace.In(peRouterNs, func() error {
+		// Grout creates a NOARP kernel interface for each port. BGP packets leave
+		// through the `main` interface but return on the port's kernel interface (grout control plane tap),
+		// so rp_filter must be disabled to allow the asymmetric path.
```
maiqueb, line 61: shouldn't we want to have this also for the kernel datapath ?

.github/workflows/ci.yaml
zeeke: that message is a leftover from a previous implementation. rephrased the commit message.

`net.ipv4.conf.default.rp_filter=1` propagates to kind containers and netns

internal/grout/passthrough.go
zeeke: in kernel datapath, BGP connections are established using the involved kernel interfaces. no asymmetric traffic happens there.

grout support traffic from kernel based applications (bgpd in this case) by exposing one TUN device per VRF (`main`) and a TAP device for each port (this is an extra tap device to make bgpd happy):

```
# ip link
2: main: <POINTOPOINT,MULTICAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UNKNOWN mode DEFAULT group default qlen 500
    link/none 
4: p0: <BROADCAST,MULTICAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 06:c2:3a:4a:ba:c8 brd ff:ff:ff:ff:ff:ff
    alias Grout control plane interface
```

bgpd sends packets via `main`, but it might happen that responses come via `p0`.

maiqueb: ok ok. got it.

zeeke:
@fedepaol @maiqueb, mind taking another look at this?

PR #566, opened 2026-07-10
Fix L2 gateway IPs being overwritten when multiple L2VNIs share a VRF

### LINKED ISSUE #563: The helper function vrfsWithL2Gateways (defined at line 1017) overwrites the map entry res[vrf]
Taking this verbatim from Gemini (because I think it's right), Claude flagged the same issue. Unless we're not allowing multiple l2vnis with l2gateways inside the same VRF?

The helper function vrfsWithL2Gateways (defined in frr_conversion.go) overwrites the map entry res[vrf] if there are multiple L2VNI resources configured with the same VRF:
```
res[vrf] = l2vni.Spec.L2GatewayIPs
```

In a multi-tenant or multi-subnet environment, multiple L2VNIs can share the same L3 VRF. Overwriting the map entry means only the gateway IPs of the last processed L2VNI will be returned and advertised via BGP, while the gateway IPs of other L2VNIs in the same VRF will be silently omitted, leading to routing failures for those subnets.

andreaskaris:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

vrfsWithL2Gateways used assignment instead of append, so only the last
L2VNI's gateway IPs were kept per VRF. Switch to append to accumulate
all gateway IPs across L2VNIs sharing the same VRF.

Add unit tests for vrfsWithL2Gateways and an integration test case in
TestAPItoFRR covering the multi-L2VNI-per-VRF scenario.

Fixes https://github.com/openperouter/openperouter/issues/563

On EVPN, the issue never surfaced, because EVPN injects /32 routes per
host. But this will be important for SRv6.

**Release note**:

```release-note
Fix bug in vrfsWithL2Gateways
```

### DIFF
--- a/internal/conversion/frr_conversion.go
+++ b/internal/conversion/frr_conversion.go
@@ -963,8 +963,12 @@ func vrfsWithL2Gateways(l2vnis []v1alpha1.L2VNI) map[string][]string {
 			if l2vni.Spec.VRF != nil {
 				vrf = *l2vni.Spec.VRF
 			}
-			res[vrf] = l2vni.Spec.L2GatewayIPs
+			res[vrf] = append(res[vrf], l2vni.Spec.L2GatewayIPs...)
 		}
+	}
+	for vrf := range res {
+		slices.Sort(res[vrf])
+		res[vrf] = slices.Compact(res[vrf])
 	}
 	return res
 }
--- a/internal/conversion/frr_conversion_test.go
+++ b/internal/conversion/frr_conversion_test.go
@@ -2810,6 +2810,94 @@ func TestAPItoFRR(t *testing.T) {
 			},
 			wantErr: false,
 		},
+		{
+			name:      "multiple L2VNIs with same VRF accumulate gateway IPs",
+			nodeIndex: 0,
+			underlays: []v1alpha1.Underlay{
+				{
+					Spec: v1alpha1.UnderlaySpec{
+						ASN:          65000,
+						RouterIDCIDR: new("10.0.0.0/24"),
+						Neighbors: []v1alpha1.Neighbor{
+							{
+								Address: new("192.168.1.1"),
+								ASN:     new(int64(65001)),
+							},
+						},
+					},
+				},
+			},
+			vnis: []v1alpha1.L3VNI{
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "vni1"},
+					Spec: v1alpha1.L3VNISpec{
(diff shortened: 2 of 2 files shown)


internal/conversion/frr_conversion.go
```diff
@@ -963,7 +963,7 @@ func vrfsWithL2Gateways(l2vnis []v1alpha1.L2VNI) map[string][]string {
 			if l2vni.Spec.VRF != nil {
 				vrf = *l2vni.Spec.VRF
 			}
-			res[vrf] = l2vni.Spec.L2GatewayIPs
+			res[vrf] = append(res[vrf], l2vni.Spec.L2GatewayIPs...)
 		}
 	}
 	return res
```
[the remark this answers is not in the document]
andreaskaris: good catch

internal/conversion/frr_conversion.go
```diff
@@ -963,8 +963,12 @@ func vrfsWithL2Gateways(l2vnis []v1alpha1.L2VNI) map[string][]string {
 			if l2vni.Spec.VRF != nil {
 				vrf = *l2vni.Spec.VRF
 			}
-			res[vrf] = l2vni.Spec.L2GatewayIPs
+			res[vrf] = append(res[vrf], l2vni.Spec.L2GatewayIPs...)
 		}
 	}
+	for vrf := range res {
```
andreaskaris, line 969: Do we need to check in WithGatewayIPs for subnet overlaps? (if so, probably in a different commit / pr)
we might need to add that to the validation logic, as well?

internal/conversion/frr_conversion_test.go
```diff
@@ -3157,6 +3245,170 @@ func TestTunnelEndpointToFRRDualStack(t *testing.T) {
...
+						L2GatewayIPs: []string{"192.168.1.1/24"},
+					},
+				},
+			},
+			want: map[string][]string{
+				"red": {"192.168.1.1/24"},
+			},
+		},
+		{
+			name: "accumulated gateway IPs are sorted",
```
andreaskaris, line 3335: kind of redundant as sorting is implicit in one of the ealier ones already, just keeping it here because the test is easy to understand / follow

internal/conversion/frr_conversion.go
fedepaol: yeah it'd make sense I guess.

PR #596, opened 2026-07-21
CNI DHCP IPAM integration

maiqueb:

>
> /kind bug
> /kind cleanup

/kind feature

> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

Add DHCP IPAM support for CNI-provisioned underlay interfaces. 

The controller now supervises the upstream CNI dhcp daemon as a child process, re-acquires leases after daemon restarts, and reconciles the assigned addresses onto the correct kernel (or grout) interface. 

Includes a `dnsmasq` DHCP server in the containerlab topology and e2e tests covering lease acquisition, controller restart, and interface cleanup.

Depends-on: #543 

The controller restart scenario will be handled in a follow-up PR, which will issue CNI checks on the reconcile loop - right now, we rely exclusively on the CNI cache to know if an interface is configured. We also want to assure that the kernel state is appropriate, by issuing CNI CHECKs. On DHCP CNI, a CHECK cmd when the lease is not present in memory (this is what happens on controller restart) will lead to lease re-acquisition. 

The DHCP server will then reply with the same IP - if configured with sticky leases (a limitation / requirement we must document) and if the client has the same MAC address (which we currently document how for macvlan / ipvlan plugins). 

**Release note**:

```release-note
The controller now manages a DHCP daemon subprocess that handles lease acquisition and renewal for CNI-provisioned underlay interfaces, with automatic lease re-acquisition after controller restarts.
```

## Summary by CodeRabbit

* **New Features**
  * Added support for CNI-provisioned underlay interfaces, including raw CNI configuration, runtime options, and DHCP-based addressing.
  * Added validation to prevent mixing interface provisioning modes and protect CNI settings from in-place changes.
  * Added DHCP lease recovery and improved interface lifecycle handling across restarts.
  * Added an EVPN CNI-underlay example and lifecycle coverage.

* **Configuration**
  * Replaced the single CNI binary directory setting with configurable plugin directories.
  * Added optional DHCP plugin configuration.

* **Documentation**
  * Updated API, configuration, and Helm documentation for CNI underlays and related settings.

### DIFF
--- a/cmd/hostcontroller/main.go
+++ b/cmd/hostcontroller/main.go
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
@@ -52,6 +53,7 @@ import (
 	"github.com/openperouter/openperouter/internal/cniinvoker"
 	"github.com/openperouter/openperouter/internal/controller/nodeindex"
 	"github.com/openperouter/openperouter/internal/controller/routerconfiguration"
+	"github.com/openperouter/openperouter/internal/dhcp"
 	"github.com/openperouter/openperouter/internal/filewatcher"
 	"github.com/openperouter/openperouter/internal/hostnetwork"
 	"github.com/openperouter/openperouter/internal/logging"
@@ -62,10 +64,11 @@ import (
 )
 
 const (
-	datapathKernel = "kernel"
-	datapathGrout  = "grout"
-	modeK8s        = "k8s"
-	modeHost       = "host"
+	datapathKernel   = "kernel"
+	datapathGrout    = "grout"
+	modeK8s          = "k8s"
+	modeHost         = "host"
+	restartDHCPEvent = "dhcp-restart-trigger"
 )
 
 var (
@@ -230,13 +233,14 @@ func runK8sMode(
 		os.Exit(1)
 	}
 
-	cniinvoker.Init(args.cniPluginDirs.values, args.cniCacheDir, args.nodeName)
+	dhcpSupervisor := dhcp.NewLazySupervisor(logger)
+	cniinvoker.Init(args.cniPluginDirs.values, args.cniCacheDir, args.nodeName, dhcpSupervisor)
 	setupLog.Info("CNI plugin invoker initialized for k8s mode",
 		"pluginDirs", args.cniPluginDirs.values,
 		"cacheDir", args.cniCacheDir)
 
 	// runK8sConfigReconciler is blocking so when running in k8s mode we should stop here
-	if err := runK8sConfigReconciler(ctx, args, k8sConfig, logger, args.probeAddr); err != nil {
+	if err := runK8sConfigReconciler(ctx, args, k8sConfig, logger, args.probeAddr, dhcpSupervisor); err != nil {
 		logger.Error("failed to enable k8s reconciler", "error", err)
 		os.Exit(1)
 	}
@@ -259,7 +263,8 @@ func runHostMode(
 		os.Exit(1)
 	}
 
-	cniinvoker.Init(args.cniPluginDirs.values, args.cniCacheDir, args.nodeName)
+	dhcpSupervisor := dhcp.NewLazySupervisor(logger)
+	cniinvoker.Init(args.cniPluginDirs.values, args.cniCacheDir, args.nodeName, dhcpSupervisor)
 	setupLog.Info("CNI plugin invoker initialized for host mode",
 		"pluginDirs", args.cniPluginDirs.values,
 		"cacheDir", args.cniCacheDir)
@@ -272,7 +277,7 @@ func runHostMode(
 		defer close(staticDone)
 		logger.Info("creating static configuration controller for host mode")
 		err := runStaticConfigReconciler(
-			staticControllerCtx, args, hostModeParams, nodeConfig, logger, args.probeAddr,
+			staticControllerCtx, args, hostModeParams, nodeConfig, logger, args.probeAddr, dhcpSupervisor,
 		)
 		if errors.Is(err, context.Canceled) {
 			logger.Info("static config reconciler stopped (API became available)")
@@ -314,7 +319,7 @@ func runHostMode(
 
 	// Start API reconciler in main thread (blocking) - keeps process alive
 	if err := runK8sConfigReconcilerHostMode(
-		ctx, args, hostModeParams, nodeConfig, k8sConfig, logger,
+		ctx, args, hostModeParams, nodeConfig, k8sConfig, logger, dhcpSupervisor,
 	); err != nil {
 		logger.Error("failed to enable k8s reconciler", "error", err)
 		os.Exit(1)
@@ -326,7 +331,8 @@ func runK8sConfigReconcilerHostMode(ctx context.Context,
 	hostModeParams hostModeParameters,
 	nodeConfig *static.NodeConfig,
 	k8sConfig *rest.Config,
-	logger *slog.Logger) error {
+	logger *slog.Logger,
+	dhcpSupervisor *dhcp.Supervisor) error {
 
 	mgr, err := createK8sManager(k8sConfig, args.nodeName, args.namespace, func(opts *ctrl.Options) {
 		opts.HealthProbeBindAddress = args.probeAddr
@@ -351,6 +357,16 @@ func runK8sConfigReconcilerHostMode(ctx context.Context,
 	if args.datapath == datapathGrout {
 		datapathConfigurator = routerconfiguration.NewGroutConfigurator(args.groutSocketPath)
 	}
+
+	dhcpSupervisor.OnRestart = triggerKubernetesReconcile(triggerChan, types.NamespacedName{
+		Namespace: restartDHCPEvent,
+		Name:      args.namespace,
+	})
+
+	if err := mgr.Add(dhcpSupervisor); err != nil {
+		return fmt.Errorf("unable to add DHCP supervisor: %w", err)
+	}
+
 	apiReconciler := &routerconfiguration.PERouterReconciler{
 		Client:               mgr.GetClient(),
 		Scheme:               mgr.GetScheme(),
@@ -427,7 +443,8 @@ func runK8sConfigReconciler(ctx context.Context,
 	args parameters,
 	k8sConfig *rest.Config,
 	logger *slog.Logger,
-	probeAddr string) error {
+	probeAddr string,
+	dhcpSupervisor *dhcp.Supervisor) error {
 
 	mgr, err := createK8sManager(k8sConfig, args.nodeName, args.namespace, func(opts *ctrl.Options) {
 		opts.HealthProbeBindAddress = probeAddr
@@ -448,6 +465,16 @@ func runK8sConfigReconciler(ctx context.Context,
 		datapathConfigurator = routerconfiguration.NewGroutConfigurator(args.groutSocketPath)
 	}
 
+	triggerChan := make(chan event.GenericEvent, 1)
+	dhcpSupervisor.OnRestart = triggerKubernetesReconcile(triggerChan, types.NamespacedName{
+		Namespace: args.namespace,
+		Name:      restartDHCPEvent,
+	})
+
+	if err := mgr.Add(dhcpSupervisor); err != nil {
+		return fmt.Errorf("unable to add DHCP supervisor: %w", err)
+	}
+
 	apiReconciler := &routerconfiguration.PERouterReconciler{
 		Client:               mgr.GetClient(),
 		Scheme:               mgr.GetScheme(),
@@ -459,6 +486,7 @@ func runK8sConfigReconciler(ctx context.Context,
 		RouterProvider:       routerProvider,
 		MyNamespace:          args.namespace,
 		DatapathConfigurator: datapathConfigurator,
+		TriggerChan:          triggerChan,
 	}
 
 	if err := apiReconciler.SetupWithManager(mgr); err != nil {
@@ -477,7 +505,8 @@ func runStaticConfigReconciler(ctx context.Context,
 	hostModeParams hostModeParameters,
 	nodeConfig *static.NodeConfig,
 	logger *slog.Logger,
-	probeAddr string) error {
+	probeAddr string,
+	dhcpSupervisor *dhcp.Supervisor) error {
 	mgr, err := ctrl.NewManager(&rest.Config{}, ctrl.Options{
 		Scheme:                 scheme,
 		HealthProbeBindAddress: probeAddr,
@@ -518,6 +547,15 @@ func runStaticConfigReconciler(ctx context.Context,
 	}
 	if err = staticReconciler.SetupWithManager(mgr); err != nil {
 		return fmt.Errorf("unable to create controller: %w", err)
+	}
+
+	dhcpSupervisor.OnRestart = func() {
+		slog.Info("triggered reconciliation after DHCP daemon restart")
+		staticReconciler.TriggerReconcile()
+	}
+
+	if err := mgr.Add(dhcpSupervisor); err != nil {
+		return fmt.Errorf("unable to add DHCP supervisor: %w", err)
 	}
 
 	if err := staticRouterProvider.StartFRRRestartWatcher(ctx, func() {
@@ -700,6 +738,27 @@ func fanOut(ctx context.Context, in <-chan event.GenericEvent, outs ...chan<- ev
 			for _, out := range outs {
 				out <- evt
 			}
+		}
+	}
+}
+
+func triggerKubernetesReconcile(
+	triggerChan chan event.GenericEvent,
+	name types.NamespacedName,
+) func() {
+	return func() {
+		select {
+		case triggerChan <- event.GenericEvent{
+			Object: &metav1.PartialObjectMetadata{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      name.Name,
+					Namespace: name.Namespace,
+				},
+			},
+		}:
+			slog.Info("triggered reconciliation after DHCP daemon restart")
+		default:
+			slog.Debug("reconciliation already queued, skipping DHCP restart trigger")
 		}
 	}
 }
--- a/config/samples/underlay-cni-dhcp.yaml
+++ b/config/samples/underlay-cni-dhcp.yaml
@@ -1,0 +1,34 @@
+apiVersion: network.openperouter.io/v1alpha1
+kind: Underlay
+metadata:
+  name: underlay
+  namespace: openperouter-system
+spec:
+  asn: 64514
+  nodeSelector:
+    matchLabels:
+      kubernetes.io/hostname: node-1
+  tunnelEndpoint:
+    cidrs:
+    - 100.65.0.0/24
+  interfaces:
+    - type: CNIDevice
+      cniDevice:
+        type: RawConfig
+        interfaceName: net1
+        rawConfig:
+          cniVersion: "1.0.0"
+          name: macvlan-underlay
+          plugins:
+            - type: macvlan
+              master: eth0
+              mode: bridge
+              capabilities:
+                mac: true
+              ipam:
+                type: dhcp
+        runtimeConfig:
+          mac: "aa:bb:cc:00:00:01"
+  neighbors:
+    - asn: 64512
+      address: 192.168.11.2
--- a/e2etests/pkg/openperouter/interface.go
+++ b/e2etests/pkg/openperouter/interface.go
@@ -3,6 +3,7 @@
 package openperouter
 
 import (
+	"fmt"
 	"regexp"
 	"sort"
 	"strings"
@@ -50,6 +51,31 @@ func InterfaceIPAddresses(nodeName, intf string) (string, error) {
 	}
 	sort.Strings(addrs)
 	return strings.Join(addrs, "\n"), nil
+}
+
+// InterfaceIPv4InNetns returns the first global-scope IPv4 address (without
+// prefix length) assigned to intf inside the named netns on nodeName, or an
+// error if none is found.
+func InterfaceIPv4InNetns(nodeName, intf, ns string) (string, error) {
+	exec := executor.ForContainer(nodeName)
+	out, err := exec.Exec("ip", "netns", "exec", ns, "ip", "-4", "-o", "a", "ls", "dev", intf, "scope", "global")
+	if err != nil {
+		return "", err
+	}
+	for line := range strings.SplitSeq(out, "\n") {
+		m := addrRegexp.FindStringSubmatch(line)
+		if len(m) < 2 {
+			continue
+		}
+		// m[1] is e.g. "inet 192.168.11.100/24"; extract the bare IP.
+		fields := strings.Fields(m[1])
+		if len(fields) < 2 {
+			continue
+		}
+		ip, _, _ := strings.Cut(fields[1], "/")
+		return ip, nil
+	}
+	return "", fmt.Errorf("no global IPv4 address on %s/%s in netns %s", nodeName, intf, ns)
 }
 
 // InterfaceIsUp checks whether the interface in the default netns
--- a/internal/cniinvoker/cni.go
+++ b/internal/cniinvoker/cni.go
@@ -15,11 +15,18 @@
 package cniinvoker
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/cni/pkg/version"
 )
+
+// DHCPEnabler starts and waits for the DHCP daemon when a CNI config uses DHCP
+// IPAM. Implementations must be safe for concurrent use.
+type DHCPEnabler interface {
+	EnsureUp(ctx context.Context) error
+}
 
 const (
 	// minSupportedCNIVersion is the minimum CNI spec version accepted for the
--- a/internal/cniinvoker/invoker.go
+++ b/internal/cniinvoker/invoker.go
@@ -13,6 +13,8 @@ import (
 	"github.com/containernetworking/cni/libcni"
 )
 
+const ipamTypeDHCP = "dhcp"
+
 // Invoker is the node-level CNI plugin invoker singleton, nil until Init is
 // called.
 var Invoker *invoker
@@ -22,15 +24,18 @@ var Invoker *invoker
 type invoker struct {
 	cniConfig   *libcni.CNIConfig
 	containerID string
+	dhcpEnabler DHCPEnabler
 }
 
 // Init sets the Invoker singleton. The cache directory must be stable across
 // controller restarts so Del keeps working for attachments created by
-// previous invocations.
-func Init(pluginDirs []string, cacheDir, nodeName string) {
+// previous invocations. If dhcpEnabler is nil, any CNI config that uses DHCP
+// IPAM will fail with an error.
+func Init(pluginDirs []string, cacheDir, nodeName string, dhcpEnabler DHCPEnabler) {
 	Invoker = &invoker{
 		cniConfig:   libcni.NewCNIConfigWithCacheDir(pluginDirs, cacheDir, nil),
 		containerID: containerIDForNode(nodeName),
+		dhcpEnabler: dhcpEnabler,
 	}
 }
 
@@ -78,6 +83,10 @@ func (inv *invoker) Add(ctx context.Context, p AddParams) error {
 	confList, err := libcni.NetworkConfFromBytes(p.Config)
 	if err != nil {
 		return err
+	}
+
+	if err := inv.ensureDHCPForConfList(ctx, confList); err != nil {
+		return fmt.Errorf("ensuring dhcp daemon for add %q: %w", p.IfName, err)
 	}
 
 	rt := &libcni.RuntimeConf{
@@ -132,6 +141,10 @@ func (inv *invoker) Check(ctx context.Context, ifName string) error {
 		return fmt.Errorf("failed to parse cached cni config for network %q: %w", attachment.Network, err)
 	}
 
+	if err := inv.ensureDHCPForConfList(ctx, confList); err != nil {
+		return fmt.Errorf("ensuring dhcp daemon for add %q: %w", ifName, err)
+	}
+
 	if err := inv.cniConfig.CheckNetworkList(ctx, confList, &libcni.RuntimeConf{
 		ContainerID:    attachment.ContainerID,
 		NetNS:          attachment.NetNS,
@@ -155,10 +168,16 @@ func (inv *invoker) Del(ctx context.Context, ifNameToDelete string) error {
 	if attachmentToDelete == nil {
 		return nil
 	}
+
 	confListToDelete, err := libcni.NetworkConfFromBytes(attachmentToDelete.Config)
 	if err != nil {
 		return fmt.Errorf("failed to parse cached cni config for network %q: %w", attachmentToDelete.Network, err)
 	}
+
+	if err := inv.ensureDHCPForConfList(ctx, confListToDelete); err != nil {
+		return fmt.Errorf("ensuring dhcp daemon for del %q: %w", ifNameToDelete, err)
+	}
+
 	if err := inv.cniConfig.DelNetworkList(ctx, confListToDelete, &libcni.RuntimeConf{
 		ContainerID:    attachmentToDelete.ContainerID,
 		NetNS:          attachmentToDelete.NetNS,
@@ -217,4 +236,17 @@ func capabilityArgsEqual(a, b map[string]any) bool {
 		return true
 	}
 	return reflect.DeepEqual(a, b)
+}
+
+func (inv *invoker) ensureDHCPForConfList(ctx context.Context, confList *libcni.NetworkConfigList) error {
+	for _, plugin := range confList.Plugins {
+		if plugin.Network == nil || plugin.Network.IPAM.Type != ipamTypeDHCP {
+			continue
+		}
+		if inv.dhcpEnabler == nil {
+			return fmt.Errorf("IPAM type is %q but DHCP support is not enabled", ipamTypeDHCP)
+		}
+		return inv.dhcpEnabler.EnsureUp(ctx)
+	}
+	return nil
 }
--- a/internal/dhcp/supervisor.go
+++ b/internal/dhcp/supervisor.go
@@ -1,0 +1,192 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package dhcp
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log/slog"
+	"net"
+	"os"
+	"os/exec"
+	"strings"
+	"sync"
+	"syscall"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	// DefaultSocketPath is the unix socket the dhcp daemon listens on and the
+	// dhcp plugin connects to. It matches the plugin's compiled-in default, so
+	// EnsureIPAM/ADD reach this daemon without extra config.
+	DefaultSocketPath = "/run/cni/dhcp.sock"
+
+	// DefaultBinPath is the well-known location of the dhcp daemon binary
+	// inside the controller container image.
+	DefaultBinPath = "/opt/openperouter/cni/bin/dhcp"
+
+	// IPAMTypeDHCP is the IPAM type string that triggers DHCP lease management.
+	IPAMTypeDHCP = "dhcp"
+
+	socketReadyTimeout = 10 * time.Second
+	socketPollInterval = 100 * time.Millisecond
+	restartBackoff     = 500 * time.Millisecond
+	shutdownGrace      = 5 * time.Second
+)
+
+// Supervisor runs and supervises the CNI dhcp IPAM daemon. It registers
+// with the controller-runtime manager immediately but blocks in Start until
+// EnsureUp is called — typically when the CNI invoker first processes an
+// ADD/DEL/CHECK for a config using DHCP IPAM. Once started, the daemon runs
+// for the lifetime of the manager; crashes are restarted automatically.
+// The supervisor opts out of leader election so the daemon runs on every node.
+type Supervisor struct {
+	// Logger receives the daemon's stdout/stderr and supervisor events.
+	Logger *slog.Logger
+	// SocketPath is the daemon's listening socket; defaults to DefaultSocketPath.
+	SocketPath string
+	// Broadcast sets the daemon's -broadcast flag (request broadcast replies),
+	// which clients without an address typically need.
+	Broadcast bool
+	// OnRestart is called when the daemon exits unexpectedly (not due to
+	// context cancellation). Callers typically use it to trigger
+	// reconciliation; the reconciler's EnsureUp call handles waiting for
+	// the restarted daemon's socket.
+	OnRestart func()
+
+	binPath string
+	startCh chan struct{}
+	once    sync.Once
+}
+
+// NewLazySupervisor creates a Supervisor. Register it with the manager via
+// mgr.Add(); it will block in Start until EnsureUp is called.
+func NewLazySupervisor(logger *slog.Logger) *Supervisor {
+	return &Supervisor{
+		Logger:  logger,
+		binPath: DefaultBinPath,
+		startCh: make(chan struct{}),
+	}
+}
+
+// NeedLeaderElection returns false so the daemon runs on every node regardless
+// of leader election.
+func (l *Supervisor) NeedLeaderElection() bool { return false }
+
+// Start blocks until EnsureUp is called, then runs the daemon and restarts it
+// on crash. Start only returns when ctx is cancelled (manager shutdown).
+func (l *Supervisor) Start(ctx context.Context) error {
+	logger := l.logger()
+	socketPath := l.socketPath()
+
+	select {
+	case <-l.startCh:
+	case <-ctx.Done():
+		return nil
+	}
+	logger.Info("DHCP supervisor started", "bin", l.binPath, "socket", socketPath)
+
+	return l.runDaemonWithRestart(ctx, logger, socketPath)
+}
+
+func (l *Supervisor) runDaemonWithRestart(ctx context.Context, logger *slog.Logger, socketPath string) error {
+	wait.UntilWithContext(
+		ctx,
+		func(ctx context.Context) {
+			err := l.runDaemon(ctx, logger, socketPath)
+			if ctx.Err() != nil {
+				return
+			}
+
+			logger.Error("dhcp daemon exited, restarting", "error", err, "backoff", restartBackoff)
+
+			if l.OnRestart != nil {
+				l.OnRestart()
+			}
+		},
+		restartBackoff,
+	)
+	logger.Info("DHCP daemon supervisor stopped", "error", ctx.Err())
+	return nil
+}
+
+// EnsureUp enables the supervisor and blocks until the DHCP daemon's socket is
+// accepting connections (or ctx is cancelled). It is safe for concurrent use;
+// when the daemon is already running the socket Dial succeeds immediately.
+func (l *Supervisor) EnsureUp(ctx context.Context) error {
+	// Signal Start() to proceed by closing startCh. sync.Once guarantees
+	// this happens exactly once even when multiple goroutines call EnsureUp
+	// concurrently — a second close would panic.
+	l.once.Do(func() { close(l.startCh) })
+	return l.waitForSocket(ctx)
+}
+
+func (l *Supervisor) waitForSocket(ctx context.Context) error {
+	socketPath := l.socketPath()
+	err := wait.PollUntilContextTimeout(ctx, socketPollInterval, socketReadyTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			dialer := net.Dialer{Timeout: socketPollInterval / 2}
+			conn, err := dialer.DialContext(ctx, "unix", socketPath)
+			if err != nil {
+				return false, nil
+			}
+			if err := conn.Close(); err != nil {
+				l.logger().Warn("failed to close socket", "error", err)
+			}
+			return true, nil
+		})
+	if err != nil {
+		return fmt.Errorf("dhcp daemon socket %s not ready after %s: %w", socketPath, socketReadyTimeout, err)
+	}
+	return nil
+}
+
+func (l *Supervisor) runDaemon(ctx context.Context, logger *slog.Logger, socketPath string) error {
+	if err := os.Remove(socketPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		logger.Warn("failed to remove stale dhcp socket", "socket", socketPath, "error", err)
+	}
+
+	args := []string{"daemon", "-socketpath", socketPath}
+	if l.Broadcast {
+		args = append(args, "-broadcast=true")
+	}
+	cmd := exec.CommandContext(ctx, l.binPath, args...)
+	cmd.Cancel = func() error { return cmd.Process.Signal(syscall.SIGTERM) }
+	cmd.WaitDelay = shutdownGrace
+	cmd.Stdout = logWriter{logger: logger, level: slog.LevelInfo}
+	cmd.Stderr = logWriter{logger: logger, level: slog.LevelWarn}
+
+	return cmd.Run()
+}
+
+func (l *Supervisor) logger() *slog.Logger {
+	if l.Logger != nil {
+		return l.Logger
+	}
+	return slog.Default()
+}
+
+func (l *Supervisor) socketPath() string {
+	if l.SocketPath != "" {
+		return l.SocketPath
+	}
+	return DefaultSocketPath
+}
+
+// logWriter forwards a subprocess's output to slog, one line per log record.
+type logWriter struct {
+	logger *slog.Logger
+	level  slog.Level
+}
+
+func (w logWriter) Write(p []byte) (int, error) {
+	for line := range strings.SplitSeq(strings.TrimRight(string(p), "\n"), "\n") {
+		if line != "" {
+			w.logger.Log(context.Background(), w.level, "DHCP daemon", "log", line)
+		}
+	}
+	return len(p), nil
+}
--- a/internal/hostnetwork/underlay_cnidev_test.go
+++ b/internal/hostnetwork/underlay_cnidev_test.go
@@ -114,7 +114,7 @@ var _ = Describe("Underlay CNI configuration", func() {
 		logPath = filepath.Join(testDir, "invocations.log")
 		Expect(os.Setenv("CNI_TEST_LOG", logPath)).To(Succeed())
 
-		cniinvoker.Init([]string{binDir}, filepath.Join(testDir, "cache"), "testnode")
+		cniinvoker.Init([]string{binDir}, filepath.Join(testDir, "cache"), "testnode", nil)
 	})
 
 	AfterEach(func() {
--- a/clab/singlecluster/dhcp1/dnsmasq.sh
+++ b/clab/singlecluster/dhcp1/dnsmasq.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+set -eu
+
+apk add --no-cache dnsmasq
+
+while ! ip link show eth1 >/dev/null 2>&1; do
+  sleep 0.1
+done
+
+ip addr add 192.168.11.1/24 dev eth1
+ip link set eth1 up
+
+exec dnsmasq \
+  --no-daemon \
+  --interface=eth1 \
+  --bind-interfaces \
+  --dhcp-range=192.168.11.100,192.168.11.150,255.255.255.0,2m \
+  --dhcp-option=option:router,192.168.11.1 \
+  --no-resolv \
+  --no-hosts \
+  --log-dhcp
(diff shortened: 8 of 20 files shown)


internal/hostnetwork/underlay.go
```diff
@@ -110,19 +126,47 @@ func SetupUnderlay(ctx context.Context, params UnderlayParams) error {
...
+	if err := cniinvoker.Invoker.Add(ctx, cniinvoker.AddParams{
+		Config:         iface.CNI.Config,
+		NetNS:          ns,
+		IfName:         iface.InterfaceName,
+		CapabilityArgs: iface.CNI.CapabilityArgs,
+	}); err != nil {
+		return fmt.Errorf("failed to setup underlay cni device %s: %w", iface.InterfaceName, err)
+	}
+	return nil
 }
```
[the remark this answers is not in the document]
maiqueb: @qinqon I think the bot's right - we should align the behavior between the 2 interface type removal funcs. 

IMHO we should collect errors and continue on both scenarios. Keep in mind DEL is idempotent, thus, shouldn't be that much of a problem.

internal/hostnetwork/underlay_cnidev_test.go
```diff
@@ -0,0 +1,252 @@
...
+			return nil
+		}
+		Expect(err).NotTo(HaveOccurred())
+		return strings.Split(strings.TrimSpace(string(content)), "\n")
+	}
+
+	BeforeEach(func() {
+		cleanTest(underlayCNITestNS)
+		testNs = createTestNS(underlayCNITestNS)
+
```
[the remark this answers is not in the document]
maiqueb: @qinqon ^ I also think this one should be acted upon.

internal/cniinvoker/ipam.go
```diff
@@ -0,0 +1,181 @@
...
+
+	cmd := exec.CommandContext(ctx, binPath)
+	cmd.Stdin = strings.NewReader(string(netconf))
+	cmd.Env = []string{
+		"CNI_COMMAND=ADD",
+		"CNI_CONTAINERID=" + inv.containerID,
+		"CNI_NETNS=" + netNS,
+		"CNI_IFNAME=" + ifName,
+		"CNI_PATH=" + strings.Join(inv.cniConfig.Path, ":"),
+	}
```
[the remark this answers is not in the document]
maiqueb: I'll use libcni instead of plain exec.

internal/hostnetwork/underlay_cnidev_test.go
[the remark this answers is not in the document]
maiqueb: we're using cleanTest between executions (after each) which cleans up the netns.

config/samples/underlay-cni-dhcp.yaml
```diff
@@ -0,0 +1,32 @@
...
+          cniVersion: "1.0.0"
+          name: macvlan-underlay
+          plugins:
+            - type: macvlan
+              master: eth0
+              mode: bridge
+              ipam:
+                type: dhcp
+        runtimeConfig:
+          mac: "aa:bb:cc:00:00:01"
```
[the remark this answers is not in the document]
maiqueb: Done.

internal/webhooks/underlay_webhook.go
```diff
@@ -144,3 +150,55 @@ var getUnderlays = func() (*v1alpha1.UnderlayList, error) {
...
+			return fmt.Errorf("cni rawConfig for interface %q is immutable, delete and recreate the Underlay to change it",
+				ifName)
+		}
+		if !reflect.DeepEqual(oldDevice.RuntimeConfig, iface.CNIDevice.RuntimeConfig) {
+			return fmt.Errorf("cni runtimeConfig for interface %q is immutable, delete and recreate the Underlay to change it",
+				ifName)
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
maiqueb: This is no longer present on this PR - it was on an old version of the dependent PR.

examples/evpn/cni-underlay/workload.yaml
```diff
@@ -0,0 +1,78 @@
...
+  containers:
+    - name: agnhost
+      image: k8s.gcr.io/e2e-test-images/agnhost:2.45
+      command: ["/bin/sh", "-c", "ip r del default dev eth0 && /agnhost netexec --http-port=8090"]
+      ports:
+      - containerPort: 8090
+        name: http
+      securityContext:
+        capabilities:
+          add: ["NET_ADMIN"]
```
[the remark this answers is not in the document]
maiqueb: This was introduced in another PR, it's not related to this one.

internal/hostnetwork/underlay_cnidev_test.go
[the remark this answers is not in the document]
maiqueb: This is now tracked in issue https://github.com/openperouter/openperouter/issues/602

cmd/hostcontroller/main.go
```diff
@@ -177,6 +179,8 @@ func main() {
 		fmt.Sprintf(
 			"comma-separated list of directories to search for CNI plugin binaries, can be specified just once, defaults to %v",
 			args.cniPluginDirs.values))
+	flag.StringVar(&args.dhcpBin, "dhcp-bin", "",
```
qinqon: Why do we need specific dhcp-bin argument ? the dhcp bin is the plugin-dirs + "dhcp"

cmd/hostcontroller/main.go
```diff
@@ -351,6 +355,20 @@ func runK8sConfigReconcilerHostMode(ctx context.Context,
 	if args.datapath == datapathGrout {
 		datapathConfigurator = routerconfiguration.NewGroutConfigurator(args.groutSocketPath)
 	}
+
+	var dhcpRestartChan chan event.GenericEvent
+	if args.dhcpBin != "" {
```
qinqon: We just if there is any "dhcp" command at any of the cni plugins dirs

cmd/hostcontroller/main.go
```diff
@@ -351,6 +355,20 @@ func runK8sConfigReconcilerHostMode(ctx context.Context,
...
+		if err := mgr.Add(&dhcp.Supervisor{
+			BinPath:       args.dhcpBin,
+			Logger:        logger,
+			RestartNotify: dhcpRestartChan,
+		}); err != nil {
+			return fmt.Errorf("unable to add DHCP supervisor: %w", err)
+		}
+		setupLog.Info("DHCP daemon supervisor registered", "bin", args.dhcpBin)
+	}
+
```
qinqon: wrap this over a setupDHCPSupervisor function so we can do later on this on the two places we configure it

```golang
setupDHCPSupervisor(apiReconciler, args...) 
```

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -371,6 +375,10 @@ func (r *PERouterReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		builder = builder.WatchesRawSource(source.Channel(r.TriggerChan, &handler.EnqueueRequestForObject{}))
 	}
 
+	if r.DHCPRestartChan != nil {
```
qinqon: Isn't reconciling the whole vni controllers on DHCP daemon restarts too much ?

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,169 @@
+// SPDX-License-Identifier:Apache-2.0
+
+// Package dhcp supervises the upstream CNI "dhcp" IPAM daemon as a child
+// process of the controller. The daemon shares the controller's mount namespace
+// and privileges, so it reaches the router netns directly and its socket is the
+// in-container default; the dhcp plugin (exec'd by libcni from the controller)
+// connects to the same socket. The supervisor restarts the daemon on exit and,
+// after each (re)start, signals a channel so a watcher can re-establish leases
+// (the daemon keeps leases in memory and forgets them across restarts).
```
qinqon: I don't think the package "dhcp" doc has to be here, I understand that is fixing the "package without doc" warning, but if for example we add something like leases.go tomorrow we will have to update the docs at supervisor.go ? move the package docs to doc.go

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,169 @@
...
+		logger.Error("dhcp daemon exited, restarting", "error", err, "backoff", restartBackoff)
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(restartBackoff):
+		}
+	}
+}
+
+func (s *Supervisor) runOnce(ctx context.Context, logger *slog.Logger, socketPath string) error {
```
qinqon: the runOnce is a pretty missleading method name, you are expecting this to have something like the golang RunOnce but what is this doing is starting the dhcp daemon, so this method should be call "startDaemon"

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,169 @@
...
+	logger := s.logger()
+	socketPath := s.socketPath()
+	logger.Info("starting dhcp daemon supervisor", "bin", s.BinPath, "socket", socketPath)
+	for {
+		err := s.runOnce(ctx, logger, socketPath)
+		if ctx.Err() != nil {
+			logger.Info("dhcp daemon supervisor stopped")
+			return nil
+		}
+		logger.Error("dhcp daemon exited, restarting", "error", err, "backoff", restartBackoff)
```
qinqon: We should not loop forever on errors that we know are permanent, like for example bad argument (socket path does not exist) on those we should make the whole chain fail so we end up with broken router pods and is quite clear we have an issue, not sure we can differenciate between transitory and permanent errors

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -371,6 +375,10 @@ func (r *PERouterReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		builder = builder.WatchesRawSource(source.Channel(r.DHCPRestartChan, &handler.EnqueueRequestForObject{}))
```
qinqon: From the state of this PR the new Reconcile will not refresh the leases since CNI ADD sill be skipped on cache, so let's just remove all this channel thing from the PR since it's doing nothing.

qinqon (changes requested):
The restart channel added to refresh DHCP leases ends up doing nothing in practice. So for the purposes of this PR — and until we implement CNI CHECK (where we'll need to be careful with grout, by the way) — let's just remove the channel to keep the PR smaller, and reintroduce it once we know it will actually do something. I haven't gone deeper into the review yet, since I'd prefer the PR to be simplified further before I continue.

~One idea: maybe we could tweak something in the CNI cache file so that libcni ends up invoking the DHCP IPAM plugin again.~

maiqueb:
> I was thinking that maybe we can hack something at the cni cache fiel so libcni end up calling DHCP IPAM again.

Let's not hack anything. Let's integrate properly, using the API.

cmd/hostcontroller/main.go
maiqueb: In case the user doesn't want to run this feature, they can opt out. 

> the dhcp bin is the plugin-dirs + "dhcp"

What if the user wants to run other dhcp CNI ? for instance, their own patched version.

cmd/hostcontroller/main.go
maiqueb: I want a way to opt out of the feature.

internal/controller/routerconfiguration/underlay_vni_controller.go
maiqueb: We have one reconcile loop, with plenty of events that trigger reconcile.

This is just one more event.

Having different scoped loops seems like an orthogonal task, which will require an enhancement.

internal/dhcp/supervisor.go
maiqueb: I can, but, I don't see that done anywhere else in this project. 

What's so special about the dhcp package ? Why is this a concern only now ?

internal/dhcp/supervisor.go
maiqueb: I got lost in the middle of the argument, and I don't know what's the action item.

What I understood: there might be recoverable errors, and non-recoverable errors. You're not sure we can differentiate them.

cmd/hostcontroller/main.go
qinqon: > In case the user doesn't want to run this feature, they can opt out.
> 

Thing is that we have to manage to start dhcp daemon only if needed, not unconditionally, meaning that 
it does not start until Underlay with "dhcp" pops up.

> > the dhcp bin is the plugin-dirs + "dhcp"
> 
> What if the user wants to run other dhcp CNI ? for instance, their own patched version.

It will have the cni type with the sepcial plugin name, same we do for bridge-cnv, and the daemon should be started with that, for that we should not start the DHCP daemon unconditioncally at openperouter start.

cmd/hostcontroller/main.go
qinqon: > I want a way to opt out of the feature.

As mentioned, I think opt-out or opt-in should be a consequence of Underlay config, so we should start DHCP daemon only if really needed.

internal/controller/routerconfiguration/underlay_vni_controller.go
qinqon: > We have one reconcile loop, with plenty of events that trigger reconcile.
> 
> This is just one more event.
> 
> Having different scoped loops seems like an orthogonal task, which will require an enhancement.

As mentin I think we should not implement this in the PR since we are not doing the lease dance without CNI CHECK

internal/dhcp/supervisor.go
qinqon: Let's do what is done at the whole project, don't add docs to the package, that should be done to the whole project at a PR, maybe just open an issue.

But if we do it doing it at a specific .go does not make much sense.

internal/dhcp/supervisor.go
qinqon: > I got lost in the middle of the argument, and I don't know what's the action item.
> 
> What I understood: there might be recoverable errors, and non-recoverable errors. You're not sure we can differentiate them.

We should not loop forever on non recoverable errors that should be fixed by users, like a bad socket path, on those case it has to fully pass error so the controller pod fails.

cmd/hostcontroller/main.go
maiqueb: I'm going to create a lazy supervisor, which'll block until we've seen an underlay config w/ DHCP IPAM configuration.

cmd/hostcontroller/main.go
maiqueb: now it's just 2 lines. I think there's no point writing an helper for that.

internal/dhcp/supervisor.go
maiqueb: Done.

internal/controller/routerconfiguration/underlay_vni_controller.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: how do you propose we differentiate those type of errors ?

maiqueb: cheapest I could come up with is https://github.com/openperouter/openperouter/pull/596/commits/0f416485822c86981709166f06e4fb3be4edb41a , but I don't see how I can further understand the underlaying cause of the error. 

Far from sure this is right.

internal/dhcp/supervisor.go
maiqueb: Done, removed it.

internal/controller/routerconfiguration/underlay_vni_controller.go
maiqueb: Removed it.

internal/controller/routerconfiguration/underlay_vni_controller.go
```diff
@@ -162,6 +169,29 @@ func (r *PERouterReconciler) reconcile(ctx context.Context, logger *slog.Logger)
 	return ctrl.Result{}, nil
 }
 
+func (r *PERouterReconciler) ensureDHCPSupervisor(underlays []v1alpha1.Underlay, logger *slog.Logger) {
```
qinqon: Keep `LazySupervisor` as the lifecycle Runnable, but move the trigger into `cniinvoker` — inject the supervisor at `Init` as a small `interface { EnsureUp(ctx) error }` (Enable + wait-for-socket), and call it from `Add` when the config has a dhcp-ipam stanza and from `Del` when the **cached attachment's** config does. Then this function, the `DHCPSupervisor` reconciler field, and the reconciler-side conflist parsing can all go away.

This also stays compatible with the CNI CHECK follow-up: the lease refresh lands in `Add`'s cached branch right after `EnsureUp` (correct daemon-ready → CHECK ordering at the call site), and the restart-notification channel wiring on the supervisor is untouched.

cmd/hostcontroller/main.go
```diff
@@ -351,6 +352,12 @@ func runK8sConfigReconcilerHostMode(ctx context.Context,
 	if args.datapath == datapathGrout {
 		datapathConfigurator = routerconfiguration.NewGroutConfigurator(args.groutSocketPath)
 	}
+
+	dhcpSupervisor := dhcp.NewLazySupervisor(logger)
```
qinqon: `runStaticConfigReconciler` (main.go:489) registers no supervisor and `StaticConfigReconciler` never enables it, yet its reconcile drives the same `Reconcile` → `SetupUnderlay` → `invoker.Add` path — a static-file underlay with dhcp IPAM would dial-fail forever there.

With the invoker-level `EnsureUp` (see my comment on `ensureDHCPSupervisor`) that mode is covered by construction; the supervisor just needs registering with the static-mode manager too.

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,192 @@
...
+
+	for {
+		err := l.startDaemon(ctx, logger, socketPath)
+		if ctx.Err() != nil {
+			logger.Info("dhcp daemon supervisor stopped")
+			return nil
+		}
+		var nre *NonRecoverableError
+		if errors.As(err, &nre) {
+			return err
```
qinqon: I think I am going to retract on this, we don't want to blow the whole openperouter business only because we have an issue with DHCP initial args, what would be nice is this kind of non recoverable errors to appear at the Status CRD on nodes.

internal/dhcp/ipam.go
```diff
@@ -0,0 +1,56 @@
...
+		var p struct {
+			IPAM struct {
+				Type string `json:"type"`
+			} `json:"ipam"`
+		}
+		if err := json.Unmarshal(plugin, &p); err != nil {
+			continue
+		}
+		if p.IPAM.Type != "" {
+			return p.IPAM.Type, nil
```
qinqon: nit: this returns the first non-empty `ipam.type`, so a conflist like `[plugin with static ipam, plugin with dhcp ipam]` mis-detects and the daemon never starts. Should be "any plugin has `ipam.type == dhcp`" — and it becomes trivial if detection moves into the invoker per-stanza.

qinqon (changes requested):
Let's move the enabling to cniinvoker (which sees both the desired configs at ADD and the provisioned attachments at DEL). Detecting DHCP from the reconciler over the desired underlays has two coverage gaps (see inline comments): the DEL/teardown path and the static-config mode.

Also enabling the dhcp supervisor at cniinvoker fits well with future CNI check, since the daemon restart channel will call reconcile and and we will end up at the cni add/check

internal/dhcp/supervisor.go
maiqueb: that's a follow-up probably requiring an enhancement for extra status conditions.

internal/dhcp/ipam.go
maiqueb: I don't think it's valid to have that configuration - i.e. if you have a CNI chain, the only plugins that are not the first must be meta plugins - they don't invoke IPAM.

internal/dhcp/supervisor.go
maiqueb: https://github.com/openperouter/openperouter/issues/612

internal/controller/routerconfiguration/underlay_vni_controller.go
maiqueb: Done.

cmd/hostcontroller/main.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: Resolving this comment since #612 will tackle it in a follow-up.

cmd/hostcontroller/main.go
maiqueb: I'm now lazily provisioning the dhcp daemon depending on the underlay configuration.

examples/evpn/cni-underlay/prepare.sh
```diff
@@ -0,0 +1,12 @@
...
+set -euo pipefail
+set -x
+CURRENT_PATH=$(dirname "$0")
+
+source "${CURRENT_PATH}/../../common.sh"
+
+DEMO_MODE=true make deploy
+export KUBECONFIG=$(pwd)/bin/kubeconfig
+
+apply_manifests_with_retries openpe.yaml workload.yaml
```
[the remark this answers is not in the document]
maiqueb: We can tackle this out of band in a separate PR.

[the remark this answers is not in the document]
maiqueb: Implemented in https://github.com/openperouter/openperouter/pull/620

Resolving this thread.

internal/hostnetwork/underlay_dhcp_test.go
```diff
@@ -0,0 +1,179 @@
...
+					break
+				}
+			}
+			if hasLinkLocal {
+				Expect(hasLinkLocal).To(BeTrue(), "IPv6 link-local should survive IPv4 reconciliation")
+			}
+			return nil
+		})
+		Expect(err).NotTo(HaveOccurred())
+	})
```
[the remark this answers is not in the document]
maiqueb: this test was removed

internal/cniinvoker/invoker.go
```diff
@@ -187,3 +200,17 @@ func capabilityArgsEqual(a, b map[string]any) bool {
...
+		return nil
+	}
+	ipamType, err := dhcp.IPAMType(config)
+	if err != nil {
+		return fmt.Errorf("failed to detect IPAM type: %w", err)
+	}
+	if ipamType == dhcp.IPAMTypeDHCP {
+		return inv.dhcpEnabler.EnsureUp(ctx)
+	}
+	return nil
```
qinqon: If IPAM type is DHCP and dhcpEnabler == nil we should fail, if not user will be activating DHCP thinking that is is going to work fine.

internal/cniinvoker/invoker.go
```diff
@@ -124,6 +132,11 @@ func (inv *invoker) Del(ctx context.Context, ifNameToDelete string) error {
 	if attachmentToDelete == nil {
 		return nil
 	}
+
+	if err := inv.ensureDHCP(ctx, attachmentToDelete.Config); err != nil {
```
qinqon: shouldn't we kill dhcp daemon if there is no more interfaces with DHCP ? 

Also same as add we can parse with NetworkConfFromBytes before calling this

internal/cniinvoker/invoker.go
```diff
@@ -63,6 +67,10 @@ func (e ConfigMismatchError) Error() string {
 // same config skips the invocation, one with a different config returns a
 // ConfigMismatchError leaving the interface untouched.
 func (inv *invoker) Add(ctx context.Context, p AddParams) error {
+	if err := inv.ensureDHCP(ctx, p.Config); err != nil {
```
qinqon: We can call `NetworkConfFromBytes` fore this so we don't have to parse things twice the first time the interface is created

internal/dhcp/ipam.go
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package dhcp
```
qinqon: I think the package should be cnidhcp, since it's not general DHCP package is very specific of how CNI DHCP is working, maybe we need to go back to a cni package where we put invoker and dhcp supevisor together

internal/dhcp/ipam.go
```diff
@@ -0,0 +1,56 @@
...
+	var c struct {
+		IPAM struct {
+			Type string `json:"type"`
+		} `json:"ipam"`
+	}
+	if err := json.Unmarshal(config, &c); err != nil {
+		return "", fmt.Errorf("failed to parse CNI config: %w", err)
+	}
+	return c.IPAM.Type, nil
+}
```
qinqon: If we pre-parse add CNI invokations we don't need all this or maybe we have something but simpler

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,192 @@
...
+	socketPollInterval = 100 * time.Millisecond
+	restartBackoff     = 2 * time.Second
+	shutdownGrace      = 5 * time.Second
+)
+
+// LazySupervisor runs and supervises the CNI dhcp IPAM daemon. It registers
+// with the controller-runtime manager immediately but blocks in Start until
+// Enable is called — typically when the reconciler first detects an underlay
+// using DHCP IPAM. It opts out of leader election so the daemon runs on every
+// node.
```
qinqon: Change docs, now we integrate it at cni add/del not reconciler

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,192 @@
...
+	logger := l.logger()
+	socketPath := l.socketPath()
+	logger.Info("DHCP supervisor enabled by underlay configuration", "bin", l.binPath, "socket", socketPath)
+
+	for {
+		err := l.startDaemon(ctx, logger, socketPath)
+		if ctx.Err() != nil {
+			logger.Info("dhcp daemon supervisor stopped")
+			return nil
+		}
```
qinqon: Don't we print the error ?

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,192 @@
...
+		logger.Error("dhcp daemon exited, restarting", "error", err, "backoff", restartBackoff)
```
qinqon: Is it worth it to differenciate between gracefull restart and not (meaning err == nil)

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,192 @@
...
+		if time.Now().After(deadline) {
+			return fmt.Errorf("dhcp daemon socket %s not ready after %s", socketPath, socketReadyTimeout)
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(socketPollInterval):
+		}
+	}
+}
```
qinqon: We usually use the wait package for this things https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,192 @@
...
+
+// NeedLeaderElection returns false so the daemon runs on every node regardless
+// of leader election.
+func (l *LazySupervisor) NeedLeaderElection() bool { return false }
+
+// Start blocks until Enable is called, then runs the daemon and restarts it on
+// exit until ctx is cancelled.
+func (l *LazySupervisor) Start(ctx context.Context) error {
```
qinqon: You can simplify this code using the wait pakage https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,192 @@
...
+		return fmt.Errorf("failed to start dhcp daemon: %w", err)
+	}
+	go func() {
+		if err := l.waitForSocket(ctx); err != nil {
+			logger.Warn("dhcp daemon socket not ready before timeout", "socket", socketPath, "error", err)
+			return
+		}
+		logger.Info("dhcp daemon ready", "socket", socketPath)
+	}()
+	return cmd.Wait()
```
qinqon: We don't need to wait for socket here, just call cmd.Run() instead of cmd.Start,waitForSocket,cmd.Wait

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (w logWriter) Write(p []byte) (int, error) {
+	for line := range strings.SplitSeq(strings.TrimRight(string(p), "\n"), "\n") {
+		if line != "" {
+			w.logger.Info("dhcp daemon", "log", line)
+		}
+	}
+	return len(p), nil
+}
```
qinqon, line 192: If you want a prefix just do something like Logger.WithGroup("dhcpo daemon"), I don't think we need to do this logWriter thinkg also to pipe in into stderr or stdout

qinqon (changes requested):
I think we can reduce the channel code by using the "wait" package

internal/dhcp/ipam.go
maiqueb: IMHO, since there aren't any other dhcp implementations in this project, I think its fairly straight-forward to understand what the package is about - there's nothing else to cause confusion.

internal/cniinvoker/invoker.go
RamLavi: +1
Also, `ensureDHCP` calls `waitForSocket`, which can be slow (up to `socketReadyTimeout`, currently 10s). 

Doing it this early means we pay that cost on every `Add()`, even when the interface is already cached and about to take the no-op fast path - adding latency (and a failure mode) for no reason.

Consider only calling `ensureDHCP` when the interface is not cached, i.e. move it into the cache-miss branch, right before we actually invoke the plugin.

maiqueb: We want to tackle a scenario where the interface is cached, but the lease needs to be renewed, don't we ? 

For that to happen, we need the DHCP daemon alive.
@RamLavi

internal/cniinvoker/invoker.go
maiqueb: IMHO Enable/Disable lifecycle is **not** required. 

The simple path forward is to start the daemon on first EnsureUp and leave it running.

Allowing it to shutdown - which indeed does address an edge case (underlay config transitions away from DHCP) - isn't worth the complexity, given the DHCP daemon is lightweight and idle when unused.

internal/cniinvoker/invoker.go
maiqueb: Now calling `NetworkConfFromBytes` before ensuring DHCP daemon's up.

internal/dhcp/supervisor.go
maiqueb: Done.

internal/dhcp/ipam.go
maiqueb: Done.

internal/cniinvoker/invoker.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: IMHO, no.

In both cases the right thing to do is identical: log, set afterSocketReady, backoff, restart. 
A clean exit from a long-running daemon is still unexpected - it shouldn't have stopped - so restarting with the same urgency makes sense. 

Splitting the two would add a branch that changes nothing about the behavior, or would need a justification for treating them differently (e.g. longer backoff for crashes, no backoff for clean exits), which I don't see a case for here.

internal/dhcp/supervisor.go
maiqueb: Done

internal/dhcp/supervisor.go
maiqueb: I am using this split pattern of start + wait to be able to invoke the `afterSocketReady` callback in between. 

This callback will cause the DHCP daemon to restart.

Let me see if I can simplify it ...

internal/dhcp/supervisor.go
maiqueb: cmd.Stdout / cmd.Stderr are `io.Writer` interfaces - they must implement the `Write` function:
```go
type Writer interface {
	Write(p []byte) (n int, err error)
}
```

The logger doesn't. Hence, this is needed.

qinqon: Right, I assumed the slog.Logger has some io.Writer implementation at the lib, but that's not the case, maybe I have different logging framework in mind ? controller-runtime or klog.

internal/dhcp/supervisor.go
maiqueb: I've applied it where I think it could make sense. But here, where in reality, what I want is an infinite loop that restarts something if the <previous something> broke, I don't think the wait pkg fits.

internal/dhcp/supervisor.go
qinqon: Ok let's forget about the WithGroup, is not really worth it but we have one thing we can improve, it would be good idea to dump stdout -> Info and stderr -> Error, meaning that add a flag to the struct to do one of the other or directly pass the Info or Error loggers depending on what stdout or stderr are we wiring up.

maiqueb: Done (stderr => logged as warn ; stdout => logged as info).

internal/dhcp/supervisor.go
maiqueb: how about now ?

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,185 @@
...
+				return false, nil
+			}
+			_ = conn.Close()
+			return true, nil
+		})
+	if err != nil {
+		return fmt.Errorf("dhcp daemon socket %s not ready after %s: %w", socketPath, socketReadyTimeout, err)
+	}
+	return nil
+}
```
qinqon: I have just found that there is a net.DialTimeout, maybe we can just use that for this ?

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,185 @@
...
+
+	args := []string{"daemon", "-socketpath", socketPath}
+	if l.Broadcast {
+		args = append(args, "-broadcast=true")
+	}
+	cmd := exec.CommandContext(ctx, l.binPath, args...)
+	cmd.Cancel = func() error { return cmd.Process.Signal(syscall.SIGTERM) }
+	cmd.WaitDelay = shutdownGrace
+	cmd.Stdout = logWriter{logger: logger, level: slog.LevelInfo}
+	cmd.Stderr = logWriter{logger: logger, level: slog.LevelError}
```
qinqon: Weren't we going to use LevelWarning for this ?

internal/dhcp/supervisor.go
qinqon: > I've applied it where I think it could make sense. But here, where in reality, what I want is an infinite loop that restarts something if the <previous something> broke, I don't think the wait pkg fits.

I still see this simpler, in fact I will move the the the last half of code to a l.runDaemonWithRestart, so the Start function just have about waiting for it to be enabled and calling runDaemonWithRestart
```golang
func (l *LazySupervisor) Start(ctx context.Context) error {
	logger := l.logger()
	socketPath := l.socketPath()

	select {
	case <-l.startCh:
	case <-ctx.Done():
		return nil
	}
	
        return l.runDaemonWithRestart(ctx, logger, socketPath, restartBackoff)
}

func(l *LazySueprvisor) runDaemonWithRestart(...) .. {
       logger.Info("DHCP daemon supervisor started", "bin", l.binPath, "socket", socketPath) 
	var lastErr error
	err :=  wait.UntilWithContext(ctx, func(ctx context.Context) {
		lastErr = l.runDaemon(ctx, logger, socketPath)
		if ctx.Err() != nil {
			return
		}

		if l.OnRestart != nil {
			l.OnRestart()
		}

		logger.Error("dhcp daemon exited, restarting", "error", lastErr, "backoff", restartBackoff)
	}, restartBackoff)
        logger.Info("DHCP daemon supervisor stopped", "error", lastErr)
	return err
}
```

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,185 @@
...
+	case <-ctx.Done():
+		return nil
+	}
+	logger.Info("DHCP supervisor started", "bin", l.binPath, "socket", socketPath)
+
+	for {
+		err := l.runDaemon(ctx, logger, socketPath)
+		if ctx.Err() != nil {
+			logger.Info("dhcp daemon supervisor stopped", "error", err)
+			return nil
```
qinqon: Is ctx.Err() the same as err ? like we are checking ctx.Err() but pprinting "err"

internal/dhcp/supervisor.go
qinqon: > IMHO, no.
> 
> In both cases the right thing to do is identical: log, set afterSocketReady, backoff, restart. 
> A clean exit from a long-running daemon is still unexpected - it shouldn't have stopped - so restarting with the same urgency makes sense. 
> 
> Splitting the two would add a branch that changes nothing about the behavior, or would need a justification for treating them differently (e.g. longer backoff for crashes, no backoff for clean exits), which I don't see a case for here.

Fair enough

internal/dhcp/ipam.go
qinqon: > Done.

Mean that we no longer need the ipam.go and ipam_test.go

internal/cniinvoker/cni.go
```diff
@@ -15,12 +15,19 @@
...
+	"context"
 	"fmt"
 
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/cni/pkg/version"
 )
 
+// DHCPEnabler starts and waits for the DHCP daemon when a CNI config uses DHCP
+// IPAM. Implementations must be safe for concurrent use.
+type DHCPEnabler interface {
```
qinqon, line 27: Maybe we can call the whole DHCPEnabler dhcpEnabler and frieneds DHCPDaemon, at the end this is what it is a DHCP daemon that we start so dhcpDaemon.EnsureUP is more clear,

internal/cniinvoker/invoker.go
```diff
@@ -218,3 +231,18 @@ func capabilityArgsEqual(a, b map[string]any) bool {
 	}
 	return reflect.DeepEqual(a, b)
 }
+
+const ipamTypeDHCP = "dhcp"
```
qinqon: Shouldn't this need to go up

cmd/hostcontroller/main.go
```diff
@@ -448,6 +474,27 @@ func runK8sConfigReconciler(ctx context.Context,
 		datapathConfigurator = routerconfiguration.NewGroutConfigurator(args.groutSocketPath)
 	}
 
+	dhcpTriggerChan := make(chan event.GenericEvent, 1)
```
qinqon: Call the channel just "triggerChan" we should not reference the user to the name

cmd/hostcontroller/main.go
```diff
@@ -448,6 +474,27 @@ func runK8sConfigReconciler(ctx context.Context,
...
+					Name:      "dhcp-restart-trigger",
+					Namespace: "default",
+				},
+			},
+		}:
+			slog.Info("triggered reconciliation after DHCP daemon restart")
+		default:
+			slog.Debug("reconciliation already queued, skipping DHCP restart trigger")
+		}
+	}
```
qinqon: this anonymous function is duplicated at the main, just create a named function with it and use it

qinqon (changes requested):
Nothing super big

cmd/hostcontroller/main.go
maiqueb: no, I'd rather be able to identify at a glance what a trigger channel is required. It just costs 4 characters.

qinqon: fair enough, I was just following the patter we have already https://github.com/openperouter/openperouter/blob/main/cmd/hostcontroller/main.go#L347-L366

maiqueb: you can't have a pattern with a single instance of something imho. 

~~I still think being able to differentiate the trigger channel for systemd from the one from dhcp at a glance is good.
I'll leave it as is.~~

OK, I take back my words - I now realize we are re-using the same channel on systemd mode. I can rename the variable in the k8s path.

cmd/hostcontroller/main.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: Done.

internal/dhcp/ipam.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: Sorted this out. Let me know what you think.

internal/dhcp/supervisor.go
maiqueb: I adapted this snippet into the code. Hope you like it now.

internal/dhcp/supervisor.go
qinqon: > Done

I have check this out, DialTimeout give us near nothing since we have to keep the wai.Poll... around it, just use net.DialContext.

internal/cniinvoker/invoker.go
qinqon, line 142: We are missing the ensure call here here, if we reconcile con DHCP daemon retart we should be sure the damon is ready before calling the actual ChecknetworkList.

```golang
if err := inv.ensureDHCPForConfList(ctx, confList); err != nil {
		return fmt.Errorf("ensuring dhcp daemon for add %q: %w", p.IfName, err)
	}
```

internal/cniinvoker/invoker.go
```diff
@@ -155,10 +162,16 @@ func (inv *invoker) Del(ctx context.Context, ifNameToDelete string) error {
 	if attachmentToDelete == nil {
 		return nil
 	}
+
 	confListToDelete, err := libcni.NetworkConfFromBytes(attachmentToDelete.Config)
 	if err != nil {
 		return fmt.Errorf("failed to parse cached cni config for network %q: %w", attachmentToDelete.Network, err)
 	}
+
+	if err := inv.ensureDHCPForConfList(ctx, confListToDelete); err != nil {
```
qinqon, line 177: So at the end do we really need the daemon on del ? I though we found that is not really needed.

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,186 @@
...
+			if l.OnRestart != nil {
+				l.OnRestart()
+			}
+
+			logger.Error("dhcp daemon exited, restarting", "error", err, "backoff", restartBackoff)
+		},
+		restartBackoff,
+	)
+	logger.Info("DHCP daemon supervisor stopped", "error", ctx.Err())
+	return ctx.Err()
```
qinqon: We should not return ctx.Err() if we do so a normal sigterm will end up making controller-runtime propagate it so we do os.Exit(1) wich is wrong, this is a normal shutdown.

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,186 @@
...
+			err := l.runDaemon(ctx, logger, socketPath)
+			if ctx.Err() != nil {
+				return
+			}
+
+			if l.OnRestart != nil {
+				l.OnRestart()
+			}
+
+			logger.Error("dhcp daemon exited, restarting", "error", err, "backoff", restartBackoff)
```
qinqon: nit: let's move the log before OnRestart so logs reads causally

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,186 @@
...
+
+	return l.runDaemonWithRestart(ctx, logger, socketPath)
+}
+
+func (l *LazySupervisor) runDaemonWithRestart(ctx context.Context, logger *slog.Logger, socketPath string) error {
+	wait.UntilWithContext(
+		ctx,
+		func(ctx context.Context) {
+			err := l.runDaemon(ctx, logger, socketPath)
+			if ctx.Err() != nil {
```
qinqon, line 100: Add a comment that this means normal shutdown, the controller runtime manager cancel the context after catching SIGTERM, so the runnables are canceled, also add a log with the ctx.Err() before return

qinqon (changes requested):
Also I am not super fan of the package naming though, the stuff about DHCP we are doing is specific to the cni dhcp plugin so maybe the package "dhcp" should be "cnidhcp"

internal/cniinvoker/invoker.go
maiqueb: Why is it not needed ? Can you explain why ? 

How can we release the lease on underlay delete if we don't send the DHCP release msg ?

internal/cniinvoker/invoker.go
maiqueb: yes, indeed - fixed!

Good catch !

internal/cniinvoker/invoker.go
maiqueb: up .... as in ?...

internal/dhcp/supervisor.go
maiqueb: isn't this the standard way a context works ? Why comment on it ?

> add a log with the ctx.Err() before return

It's already there, in line 112.

internal/cniinvoker/invoker.go
qinqon: right, disregard.

internal/dhcp/supervisor.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: But where does it hurt ? Is there something that can go wrong by using this ?

internal/dhcp/supervisor.go
maiqueb: Done.

maiqueb:
> Also I am not super fan of the package naming though, the stuff about DHCP we are doing is specific to the cni dhcp plugin so maybe the package "dhcp" should be "cnidhcp"

If you find another integration point in this project about dhcp, I'll rename it.

maiqueb (commented):
@qinqon please take another look.

internal/cniinvoker/invoker.go
qinqon: Usually constants are declared/defined at the beginning of the file.

internal/dhcp/supervisor.go
qinqon: If we don't use DialContext passing the controller-runtime manager context we can end up hanging here after sigterm, just replace DialTimeout with DialContext

internal/cniinvoker/invoker.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: so what I hear is this was better *before* than it is now. Should have asked what was the point of using this. 

Resetting this back to its original version.

internal/dhcp/supervisor.go
maiqueb: having pursued this was a mistake, as per https://github.com/openperouter/openperouter/pull/596/changes#r3689265125

internal/cniinvoker/cni.go
maiqueb: That's debatable. 

Maybe we could, but, for now, unless another maintainer wants it, I won't follow through.

internal/dhcp/supervisor.go
maiqueb: now using a dialer with context ...

qinqon (approved):
LGTM

e2etests/pkg/openperouter/interface.go
```diff
@@ -52,6 +53,31 @@ func InterfaceIPAddresses(nodeName, intf string) (string, error) {
 	return strings.Join(addrs, "\n"), nil
 }
 
+// InterfaceIPv4InNetns returns the first global-scope IPv4 address (without
+// prefix length) assigned to intf inside the named netns on nodeName, or an
+// error if none is found.
+func InterfaceIPv4InNetns(nodeName, intf, ns string) (string, error) {
```
fedepaol, line 59: this will collide with https://github.com/openperouter/openperouter/pull/618 , might make sense to run a pod to do this.

internal/cniinvoker/invoker.go
```diff
@@ -218,3 +237,16 @@ func capabilityArgsEqual(a, b map[string]any) bool {
 	}
 	return reflect.DeepEqual(a, b)
 }
+
+func (inv *invoker) ensureDHCPForConfList(ctx context.Context, confList *libcni.NetworkConfigList) error {
+	for _, plugin := range confList.Plugins {
+		if plugin.Network == nil || plugin.Network.IPAM.Type != ipamTypeDHCP {
```
fedepaol, line 243: should we cry if we have an ipam we don't support? Or will it work out of the box?

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,187 @@
...
+	OnRestart func()
+
+	binPath string
+	startCh chan struct{}
+	once    sync.Once
+}
+
+// NewLazySupervisor creates a LazySupervisor. Register it with the manager via
+// mgr.Add(); it will block in Start until EnsureUp is called.
+func NewLazySupervisor(logger *slog.Logger) *LazySupervisor {
```
fedepaol: might be me, but I think Lazy is redundant here as it's an implementation detail (and we don't have a non lazy one)

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,187 @@
...
+	)
+	logger.Info("DHCP daemon supervisor stopped", "error", ctx.Err())
+	return nil
+}
+
+// EnsureUp enables the supervisor and blocks until the DHCP daemon's socket is
+// accepting connections (or ctx is cancelled). It is safe for concurrent use;
+// when the daemon is already running the socket Dial succeeds immediately.
+func (l *LazySupervisor) EnsureUp(ctx context.Context) error {
+	l.once.Do(func() { close(l.startCh) })
```
fedepaol, line 123: let's add a comment on what this is about as it's pretty condensed

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,187 @@
...
+func (l *LazySupervisor) waitForSocket(ctx context.Context) error {
+	socketPath := l.socketPath()
+	err := wait.PollUntilContextTimeout(ctx, socketPollInterval, socketReadyTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			dialer := net.Dialer{Timeout: socketPollInterval / 2}
+			conn, err := dialer.DialContext(ctx, "unix", socketPath)
+			if err != nil {
+				return false, nil
+			}
+			_ = conn.Close()
```
fedepaol: log the error?

enhancements/controller-provisioned-underlay-interfaces.md
fedepaol, line 1: not sure I like touching enhancements. My point of view is that they reflect the status at the time of discussion, which can drift at implementation stage

internal/cniinvoker/cni.go
fedepaol: I'd just call it "dhcp", but enabler is fine too

internal/dhcp/supervisor.go
```diff
@@ -0,0 +1,187 @@
...
+
+const (
+	// DefaultSocketPath is the unix socket the dhcp daemon listens on and the
+	// dhcp plugin connects to. It matches the plugin's compiled-in default, so
+	// EnsureIPAM/ADD reach this daemon without extra config.
+	DefaultSocketPath = "/run/cni/dhcp.sock"
+
+	// DefaultBinPath is the well-known location of the dhcp daemon binary
+	// inside the controller container image.
+	DefaultBinPath = "/opt/openperouter/cni/bin/dhcp"
```
fedepaol, line 29: is this provided together with cni binaries? I didn't see changes in the dockerfile

maiqueb: it's already being shipped: https://github.com/openperouter/openperouter/blob/ba8837816b7bbd8eaad5fda27008048c5c45ae78/Dockerfile#L66

internal/cniinvoker/invoker.go
maiqueb: we'll do CNI ADD / CHECK / DEL as part of the reconcile. 

Depending on the plugin, it'll work or not. 

I originally wanted to enforce only allowing the plugins known to work (dhcp + static), but @phoracek highlighted that we would probably make the community happier if we chose flexibility (while offering no guarantees these other plugins would work).

My opinion: we should not do anything if we see another IPAM type.

e2etests/pkg/openperouter/interface.go
maiqueb: I think #618 needs to sort that out somehow, across the entire code base.

internal/dhcp/supervisor.go
maiqueb: Done - removed that prefix.

internal/dhcp/supervisor.go
maiqueb: Done.

internal/dhcp/supervisor.go
maiqueb: Done.

enhancements/controller-provisioned-underlay-interfaces.md
maiqueb: I dropped those changes. IMHO, it could have gone either way.

maiqueb (commented):
@fedepaol

Issue #597, opened 2026-07-21
grout datapath: underlay BGP sessions never recover after a router pod restart


qinqon:
In the grout datapath, underlay setup moves the interface addresses into grout and removes them from the kernel interfaces (`migrateAddressesToGrout`). Grout's state is in-memory only and the reverse migration (`migrateAddressesToKernel`) only runs on deliberate underlay removal — there is no shutdown hook, and none would survive a crash.

When the router pod restarts (upgrade, eviction, crash), the grout container dies with it: the addresses are then gone everywhere — not in grout, not on the kernel interfaces, and not re-derivable from the Underlay API (it carries no interface address). The next reconcile recreates the grout ports address-less, so the underlay BGP sessions never re-establish.

Affects both NetworkDevice and CNI-provisioned underlays; no e2e restarts router pods under grout, which is why it went unnoticed (surfaced by the CNI lifecycle test in #543, where the spec is skipped under grout).

Possible directions:
- persist the migrated addresses (e.g. under /var/lib/openperouter) and restore them when the grout port is recreated
- keep the addresses on the kernel interfaces if grout can tolerate it
- for CNI interfaces, restore from the libcni cached result (it records the assigned IPs)

zeeke:
Persisting the addresses in a file looks good to me. I'd use `/var/run/...`, as the information should be cleared after reboot

PR #509, opened 2026-06-17
feature: Implement route reflector

qinqon:
**Is this a BUG FIX or a FEATURE ?**:

/kind feature

**What this PR does / why we need it**:
Implement route reflector functionality into the Underlay as specified from the [enhancement](https://github.com/openperouter/openperouter/blob/main/enhancements/route-reflector.md)

It also adds a commit to deploy the clab singlecluster with a pair of kind workers to properly test
route reflection: `workers_for_cluster()` in `clab/scripts/03-kind-configs.sh` wires two workers for the
`pe-kind` cluster, so every singlecluster CI lane now runs with 3 nodes (control-plane + 2 workers).

The route reflector e2e test skips itself when the cluster has fewer than 3 nodes.

**Special notes for your reviewer**:

Depends-On:
- https://github.com/openperouter/openperouter/pull/669

Configure route reflector client at neighbor level on the underlay has not being implemented, if needed this will be done as a follow up, frr only supports route reflector client per AF.

**Release note**:

```release-note
Introduce route reflector support: some nodes can be instructed to have route reflector clients. This enables scenarios where the TOR can't be modified in order to propagate routes across the nodes.

Also, support listenRange to configure neighbors.
```

**AI Guidelines Acknowledgment**:

- [x] I have reviewed all changes in this PR, including any AI-generated content, and I take full responsibility for its accuracy and correctness.

## Summary by CodeRabbit

* **New Features**
  * Added BGP route-reflector support for Underlay, including `routeReflector.clusterID`, neighbor `listenRange`, and per-address-family `routeReflectorClient`.
  * Enabled dynamic iBGP peering via `listenRange`.
  * Introduced configurable dynamic BGP session caps via `bgpListenLimit` (FRR `bgp listen limit`).
* **Documentation**
  * Updated API/config docs and added route-reflector + listen-range examples and a runnable demo target.
* **Bug Fixes**
  * Strengthened CRD admission validations for neighbor addressing modes and `routeReflectorClient` applicability.
* **Tests**
  * Expanded E2E coverage for route reflection and listen-range, plus FRR/unit golden tests.

PR #614, opened 2026-07-26
feat(cni): invoke CNI CHECK during underlay reconciliation

RamLavi:

/kind feature

Following #543, this PR wires CNI CHECK into the underlay reconciliation path. On every
reconcile, the controller runs CHECK against the cached attachment before trusting it. 
When CHECK fails, the interface is torn down (DEL) and re-provisioned (ADD), so the 
kernel state - not just the cache - determines whether an interface needs rebuilding.

- The `internal/grout` datapath shares the same `SetupUnderlayCNIDevInterface`
  function, so the fix covers both the kernel and grout datapaths without
  separate changes.

**Release note**:

```release-note
CNI-provisioned underlay interfaces are now validated with a CNI CHECK
on every reconcile. If an interface was removed or misconfigured
outside of OpenPERouter, the next reconcile tears it down and
re-provisions it instead of trusting a stale cache entry.
```

## Summary by CodeRabbit

* **New Features**
  * Added CNI `CHECK`-based validation for cached CNI-provisioned underlay interfaces during reconcile.
  * On failed `CHECK` or detected drift, OpenPERouter now tears down via CNI `DEL` and re-provisions via CNI `ADD` so CNI sessions can re-establish (including after controller/router restarts).
  * Expanded end-to-end scenarios for DEL/ADD reprovisioning on interface changes and out-of-band interface deletion; enabled router pod restart coverage in Grout mode.
* **Documentation**
  * Updated docs to describe `CHECK`-driven drift detection and automatic teardown/reprovisioning.
* **Bug Fixes**
  * Under grout datapath, underlay specs backed by `CNIDevice`-based CNIs are now rejected.

### DIFF
--- a/e2etests/tests/cni_lifecycle.go
+++ b/e2etests/tests/cni_lifecycle.go
@@ -46,7 +46,7 @@ const (
 // CNI DEL / ADD reprovisioning on interface changes and the teardown
 // through CNI DEL. The traffic coverage runs in the EVPN routes suites,
 // parameterized by underlay flavor.
-var _ = Describe("CNI underlay lifecycle", GroutSupport, Ordered, func() {
+var _ = Describe("CNI underlay lifecycle", Ordered, func() {
 	var cs clientset.Interface
 	nodes := []corev1.Node{}
 
@@ -184,15 +184,6 @@ var _ = Describe("CNI underlay lifecycle", GroutSupport, Ordered, func() {
 	})
 
 	It("keeps the CNI interfaces when the router pods are restarted", func() {
-		if GroutMode {
-			// Restarting the router wipes the grout state and the underlay
-			// addresses live in grout (they are removed from the kernel
-			// interfaces), so the sessions cannot re-establish. This is a
-			// grout datapath limitation independent of how the underlay
-			// interface is provisioned.
-			Skip("the grout datapath does not recover the underlay addresses after a router restart, " +
-				"see https://github.com/openperouter/openperouter/issues/597")
-		}
 		indexesBefore, err := cniInterfaceIndexes(nodes, infra.CNIUnderlayInterface)
 		Expect(err).NotTo(HaveOccurred())
 
--- a/e2etests/tests/cni_lifecycle.go
+++ b/e2etests/tests/cni_lifecycle.go
@@ -43,9 +43,10 @@ const (
 // The CNI underlay lifecycle coverage exercises the behaviors specific to
 // CNI provisioned underlay interfaces: the libcni result cache driving the
 // reconciliation across controller and router restarts, the in-place
-// CNI DEL / ADD reprovisioning on interface changes and the teardown
-// through CNI DEL. The traffic coverage runs in the EVPN routes suites,
-// parameterized by underlay flavor.
+// CNI DEL / ADD reprovisioning on interface changes, CNI CHECK validation
+// of cached state on every reconciliation, and the teardown through CNI DEL.
+// The traffic coverage runs in the EVPN routes suites, parameterized by
+// underlay flavor.
 var _ = Describe("CNI underlay lifecycle", Ordered, func() {
 	var cs clientset.Interface
 	nodes := []corev1.Node{}
@@ -195,6 +196,35 @@ var _ = Describe("CNI underlay lifecycle", Ordered, func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(indexesAfter).To(Equal(indexesBefore),
 			"the interfaces should survive the router restart untouched")
+
+		By("checking the sessions re-establish")
+		validateSessionUp()
+	})
+
+	It("reprovisions the CNI interface after external drift is caught by the next reconcile", func() {
+		indexesBefore, err := cniInterfaceIndexes(nodes, infra.CNIUnderlayInterface)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("deleting the CNI interface directly, behind the controller's back")
+		for _, node := range nodes {
+			exec := executor.ForContainer(node.Name)
+			_, err := exec.Exec("ip", "netns", "exec", openperouter.NamedNetns, "ip", "link", "del", infra.CNIUnderlayInterface)
+			Expect(err).NotTo(HaveOccurred())
+		}
+		validateCNIInterfacesGone(infra.CNIUnderlayInterface)
+
+		// Reconciliation is event-driven, not on a timer: the dangling cache
+		// entry is only caught on the next reconcile. Restarting the
+		// controllers forces one without touching the Underlay.
+		By("forcing a reconcile by restarting the controllers")
+		restartControllers()
+
+		By("checking cni check detects the drift and the interface is reprovisioned")
+		validateCNIInterfacesPresent(infra.CNIUnderlayInterface)
+		indexesAfter, err := cniInterfaceIndexes(nodes, infra.CNIUnderlayInterface)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(indexesAfter).NotTo(Equal(indexesBefore),
+			"the interface should have been recreated with a new ifindex")
 
 		By("checking the sessions re-establish")
 		validateSessionUp()
--- a/internal/hostnetwork/underlay.go
+++ b/internal/hostnetwork/underlay.go
@@ -133,9 +133,19 @@ func SetupUnderlayNetDevInterface(ctx context.Context, ns netns.NsHandle,
 	return nil
 }
 
-// SetupUnderlayCNIDevInterface provisions a single underlay cni dev interface
+// SetupUnderlayCNIDevInterface provisions a single underlay cni dev interface.
+// A CNI CHECK runs first and, if it fails, the interface is torn down so the
+// subsequent Add provisions it fresh.
 func SetupUnderlayCNIDevInterface(ctx context.Context, ns string,
 	iface UnderlayInterface) error {
+	if err := cniinvoker.Invoker.Check(ctx, iface.InterfaceName); err != nil {
+		slog.WarnContext(ctx, "cni check failed, rebuilding underlay cni device",
+			"interface", iface.InterfaceName, "error", err)
+		if err := cniinvoker.Invoker.Del(ctx, iface.InterfaceName); err != nil {
+			return fmt.Errorf("failed to delete unhealthy underlay cni device %s: %w", iface.InterfaceName, err)
+		}
+	}
+
 	if err := cniinvoker.Invoker.Add(ctx, cniinvoker.AddParams{
 		Config:         iface.CNI.Config,
 		NetNS:          ns,
--- a/internal/hostnetwork/underlay_cnidev_test.go
+++ b/internal/hostnetwork/underlay_cnidev_test.go
@@ -43,9 +43,11 @@ var underlayCNITestConfig = fmt.Sprintf(`{
 }`, underlayCNITestPlugin)
 
 // fakeCNIPlugin creates a dummy link with a fixed address inside CNI_NETNS on
-// ADD and deletes it on DEL, recording every invocation in CNI_TEST_LOG. It
-// lets the tests exercise the full CNI provisioning path without shipping
-// real plugins.
+// ADD and deletes it on DEL, recording every invocation in CNI_TEST_LOG. On
+// CHECK it verifies the link still exists in the netns, so tests can drive a
+// real CHECK failure by deleting the link behind the invoker's back. It lets
+// the tests exercise the full CNI provisioning path without shipping real
+// plugins.
 const fakeCNIPlugin = `#!/bin/sh
 set -e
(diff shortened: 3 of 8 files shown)


internal/hostnetwork/underlay_cnidev_test.go
```diff
@@ -151,12 +159,31 @@ var _ = Describe("Underlay CNI configuration", func() {
...
-	It("is idempotent through the cni cache", func() {
+	It("is idempotent through the cni cache when the interface is healthy", func() {
+		params := cniParams("net1")
+		Expect(SetupUnderlay(context.Background(), params)).To(Succeed())
+		Expect(SetupUnderlay(context.Background(), params)).To(Succeed())
+
+		Expect(loggedCommands()).To(HaveExactElements("ADD net1", "CHECK net1"),
+			"the second setup should pass cni check and be served from the cache, without a new ADD")
+		validateCNIInterfaceInNS(testNs, "net1")
+	})
```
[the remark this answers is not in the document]
RamLavi: perhaps in a follow up, this it not in the PR scope

RamLavi:
failures do not seem related to this change (either grout https://github.com/openperouter/openperouter/issues/597 related, infra (docker hub [timeout](https://github.com/openperouter/openperouter/actions/runs/30202016566/job/89793949085?pr=614#step:10:152)) related, or network curl related)

maiqueb (commented):
I think this PR does what it needs to, but we need to coordinate with #596 - which also relies on CNI check to ensure the dhcp daemon is available. 

We should sync out of band to decide which goes in first, then ensure the DHCP flows work as expected.

maiqueb:
CI's failing quite hard in this PR as well @fedepaol . Maybe the route reflector PR is not to blame for CI ? #509

e2etests/tests/cni_lifecycle.go
```diff
@@ -43,9 +43,10 @@ const (
 // The CNI underlay lifecycle coverage exercises the behaviors specific to
 // CNI provisioned underlay interfaces: the libcni result cache driving the
 // reconciliation across controller and router restarts, the in-place
-// CNI DEL / ADD reprovisioning on interface changes and the teardown
-// through CNI DEL. The traffic coverage runs in the EVPN routes suites,
-// parameterized by underlay flavor.
+// CNI DEL / ADD reprovisioning on interface changes, self-healing of
+// externally caused drift via CNI CHECK, and the teardown through CNI DEL.
+// The traffic coverage runs in the EVPN routes suites, parameterized by
```
maiqueb, line 48: I don't think we should mention things like self-healing. 

We don't reconcile on externally caused drift (we're clueless about that) and we only reconcile when something changes in the API, or when the static configuration changes.

internal/hostnetwork/underlay.go
```diff
@@ -133,9 +133,23 @@ func SetupUnderlayNetDevInterface(ctx context.Context, ns netns.NsHandle,
...
+// unhealthy, it is torn down so the subsequent Add provisions it fresh
+// instead of trusting the stale cache entry.
 func SetupUnderlayCNIDevInterface(ctx context.Context, ns string,
 	iface UnderlayInterface) error {
+	if err := cniinvoker.Invoker.Check(ctx, iface.InterfaceName); err != nil {
+		slog.WarnContext(ctx, "cni check failed, rebuilding underlay cni device",
+			"interface", iface.InterfaceName, "error", err)
+		if err := cniinvoker.Invoker.Del(ctx, iface.InterfaceName); err != nil {
+			return fmt.Errorf("failed to delete unhealthy underlay cni device %s: %w", iface.InterfaceName, err)
+		}
```
maiqueb, line 146: the interface is torn down based on the cached values. I think it's fine, but I'd like to get more feedback from @RamLavi and @fedepaol

maiqueb (commented):
minor comments.

internal/hostnetwork/underlay.go
RamLavi: FWIW I don't see an issue. if we got here, then the kernel data is already done (interface deleted) so it makes sense to delete using the cached data (which we got from the ADD).

internal/hostnetwork/underlay.go
```diff
@@ -133,9 +133,23 @@ func SetupUnderlayNetDevInterface(ctx context.Context, ns netns.NsHandle,
...
+// SetupUnderlayCNIDevInterface provisions a single underlay cni dev interface.
+// Every time the underlay is reconciled, a CNI CHECK runs against the cached
+// attachment first: the libcni cache can drift from the real namespace state
+// (e.g. the interface was deleted externally), so the cache alone cannot be
+// trusted as the source of truth. When CHECK reports the interface
+// unhealthy, it is torn down so the subsequent Add provisions it fresh
+// instead of trusting the stale cache entry.
 func SetupUnderlayCNIDevInterface(ctx context.Context, ns string,
 	iface UnderlayInterface) error {
+	if err := cniinvoker.Invoker.Check(ctx, iface.InterfaceName); err != nil {
```
qinqon, line 141: We should skip CNI check for grout, when using grout addresses are removed from interfaces and moved to the grout user space process, so if we use CNI CHECK with a static ip IPAM cni check will fail.

internal/hostnetwork/underlay.go
```diff
@@ -133,9 +133,23 @@ func SetupUnderlayNetDevInterface(ctx context.Context, ns netns.NsHandle,
...
 }
 
-// SetupUnderlayCNIDevInterface provisions a single underlay cni dev interface
+// SetupUnderlayCNIDevInterface provisions a single underlay cni dev interface.
+// Every time the underlay is reconciled, a CNI CHECK runs against the cached
+// attachment first: the libcni cache can drift from the real namespace state
+// (e.g. the interface was deleted externally), so the cache alone cannot be
+// trusted as the source of truth. When CHECK reports the interface
+// unhealthy, it is torn down so the subsequent Add provisions it fresh
+// instead of trusting the stale cache entry.
```
qinqon: Not sure we have to go that deep on CNI check implementation, just mention the interface
is reconstructed if CNI check fails, we should not document what is already documented at Invoker.* functions

e2etests/tests/cni_lifecycle.go
RamLavi: DONE

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/b14140d5b466d718fbe402bb92e901341653cbd7..3603980a6fb67cde818b7125aaaaaad8150a6af1):
 Commit 3 (test/e2e): 
  - e2etests/tests/cni_lifecycle.go: Replaced "self-healing of externally caused drift via CNI CHECK" with "CNI CHECK validation of cached state on every
  reconciliation" in the suite comment block
  - Commit message reworded: "cover CNI CHECK validation on external interface drift"

e2etests/tests/cni_lifecycle.go
```diff
@@ -209,6 +210,35 @@ var _ = Describe("CNI underlay lifecycle", GroutSupport, Ordered, func() {
...
+
+		By("checking cni check detects the drift and the interface is reprovisioned")
+		validateCNIInterfacesPresent(infra.CNIUnderlayInterface)
+		indexesAfter, err := cniInterfaceIndexes(nodes, infra.CNIUnderlayInterface)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(indexesAfter).NotTo(Equal(indexesBefore),
+			"the interface should have been recreated with a new ifindex")
+
+		By("checking the sessions re-establish")
+		validateSessionUp()
```
qinqon, line 230: Not sure if first we need to check if the bgp session is gone at interface del, problem though is that I think when we do this kind of things we have to wait for the BGP session to timeout sice explicitly is not closed without BFD.

e2etests/tests/cni_lifecycle.go
```diff
@@ -209,6 +210,35 @@ var _ = Describe("CNI underlay lifecycle", GroutSupport, Ordered, func() {
...
+			Expect(err).NotTo(HaveOccurred())
+		}
+		validateCNIInterfacesGone(infra.CNIUnderlayInterface)
+
+		// Reconciliation is event-driven, not on a timer: the dangling cache
+		// entry is only caught on the next reconcile. Restarting the
+		// controllers forces one without touching the Underlay.
+		By("forcing a reconcile by restarting the controllers")
+		restartControllers()
+
```
qinqon, line 221: This is interesting, maybe we should reconcile on CNI changes or netlink changes ? not sure if that's even possible.

qinqon (changes requested):
Nothing big

e2etests/tests/cni_lifecycle.go
maiqueb: it's clearly possible, but I don't think we want to do that *now*. 

If we do - we should have an enhancement first, design exactly what we want to have, then implement it. 

Let's not lose focus. Please open an issue about that for tracking.

internal/hostnetwork/underlay.go
maiqueb: skip ? No. I'd say we fail it, and report an error to the user. 
This use-case can't be implemented for that platform. Open an issue to tackle it as a follow-up, *or* tackle it in this PR (use a dummy interface, DOWN state to serve as a placeholder for the IPAM configuration).

But silently doing something other than what the user requested is *not* right imho.

RamLavi: hmm, so from what I understand from you, returning an error here would break the entire grout CNI dev underlay flow on the CHECK, as SetupUnderlayCNIDevInterface runs on every reconcile tick.

I generally agree that skipping is bad, but returning FAIL just breaks the entire grout path.. I don't like to open a follow up and merge the PR if it's s breaking grout in such a way.

@maiqueb can you elaborate on the dummy DOWN solution you suggest? IIUC it would mean keeping the interface on the kernel just so it won't fail on the CHECK. that feels very hacky imo - it might bite us in the next corner. 

perhaps skipping is the lesser evil here, though I'm open for suggestions. thoughts?

qinqon: All this feels like we have to document that CNI is not supported to grout and fail if user try to configure boths, if we need to support it a enhancement is needed.

maiqueb: that's the point I'm trying to make: the user is providing a configuration that is simply not supported on that platform. They can only use the `NetworkDevice` option. 

As @qinqon says, we require an enhancement for that integration point.

RamLavi: OK, I'll add a validation check so make sure the user understand this is an invalid config.
And opened issue for future developement https://github.com/openperouter/openperouter/issues/629

e2etests/tests/cni_lifecycle.go
RamLavi: hmm I don't get what you mean. AFAIU `validateSessionUp` only checks that sessions are established after the interface is reprovisioned - it doesn't assert the session goes down during the drift window.
Can you explain?

e2etests/tests/cni_lifecycle.go
RamLavi: opened https://github.com/openperouter/openperouter/issues/631

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/3603980a6fb67cde818b7125aaaaaad8150a6af1..4c284caad4246901cd3af840e8fff9a2fdfffe7b): Address @qinqon review

  Commit 2 (feat/controller): 9f9fe7ec
  - internal/hostnetwork/underlay.go: Trimmed function comment to 2 lines per Q2 review

  Commit 3 (fix/grout): 7363a5f1 -- NEW
  - internal/conversion/validate_grout.go: Reject CNI dev underlays at validation time on grout
  - internal/conversion/validate_grout_test.go: Table-driven test asserting rejection and NetworkDevice acceptance

internal/hostnetwork/underlay.go
RamLavi: fixed

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/4c284caad4246901cd3af840e8fff9a2fdfffe7b..303e04936aafbd33bb2dbb4cab97b69ff8eafbf9):

  Commit 3 (fix/grout): 
  - e2etests/tests/cni_lifecycle.go: Remove GroutSupport label and dead GroutMode skip

e2etests/tests/cni_lifecycle.go
qinqon: You can disregard my comment it will take the test to strict.

qinqon (approved):
LGTM

maiqueb (approved):
Nice. Let's wait for CI ...

maiqueb:
@please rebase the PR - I've rebased it using the button, but, we'd like to have the history properly recorded.

RamLavi:
[Change](https://github.com/openperouter/openperouter/compare/2ac9007a143bd2958277121a062ed626bd220e89..4c21dc35dedc031ad92fc35d527b71cdfbbeb407): Rebase (no conflicts)

PR #618, opened 2026-07-27
adapt e2e suite to run on other platforms

oribon:

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

The existing suite relies on the cluster being kind (where nodes are containers), making it impossible to run the suite on other k8s clusters. Here we adapt the suite to support generic k8s clusters, under the assumption that the clab is deployed the way (w.r.t to the cluster).

**Release note**:

```release-note
NONE
```

## Summary by CodeRabbit

* **New Features**
  * Added a node-exec helper for node-scoped E2E command execution with automated setup/teardown.
  * Added JSON-driven node/fabric link registration (including a new default links file) and suite flags for link config and FRR-K8s namespace.

* **Bug Fixes**
  * Improved E2E configuration namespace handling during updates and cleanup.
  * FRR config injection now uses the configured container runtime.
  * Enhanced error handling across interface, netns, and underlay checks.

* **Tests**
  * Updated E2E scenarios to run commands on real Kubernetes nodes and improved polling/error surfacing.

### DIFF
--- a/e2etests/nodelink-default.json
+++ b/e2etests/nodelink-default.json
@@ -1,0 +1,34 @@
+{
+  "nodes": {
+    "pe-kind-control-plane": {
+      "ipForKindLeaf": "192.168.11.3",
+      "ipForKindLeaf2": "192.168.12.3",
+      "ipv6ForKindLeaf": "2001:db8:11::3",
+      "ipv6ForKindLeaf2": "2001:db8:12::3",
+      "ifaceForKindLeaf": "toleafkind1",
+      "ifaceForKindLeaf2": "toleafkind2",
+      "leafIfaceForKindLeaf": "tokindctrlpl",
+      "leafIfaceForKindLeaf2": "tokindctrlpl"
+    },
+    "pe-kind-worker": {
+      "ipForKindLeaf": "192.168.11.4",
+      "ipForKindLeaf2": "192.168.12.4",
+      "ipv6ForKindLeaf": "2001:db8:11::4",
+      "ipv6ForKindLeaf2": "2001:db8:12::4",
+      "ifaceForKindLeaf": "toleafkind1",
+      "ifaceForKindLeaf2": "toleafkind2",
+      "leafIfaceForKindLeaf": "tokindworker",
+      "leafIfaceForKindLeaf2": "tokindworker"
+    },
+    "pe-kind-worker2": {
+      "ipForKindLeaf": "192.168.11.5",
+      "ipForKindLeaf2": "192.168.12.5",
+      "ipv6ForKindLeaf": "2001:db8:11::5",
+      "ipv6ForKindLeaf2": "2001:db8:12::5",
+      "ifaceForKindLeaf": "toleafkind1",
+      "ifaceForKindLeaf2": "toleafkind2",
+      "leafIfaceForKindLeaf": "tokindworker2",
+      "leafIfaceForKindLeaf2": "tokindworker2"
+    }
+  }
+}
--- a/e2etests/nodelink-default.json
+++ b/e2etests/nodelink-default.json
@@ -1,0 +1,24 @@
+{
+  "nodes": {
+    "pe-kind-control-plane": {
+      "ipForKindLeaf": "192.168.11.3",
+      "ipForKindLeaf2": "192.168.12.3",
+      "ipv6ForKindLeaf": "2001:db8:11::3",
+      "ipv6ForKindLeaf2": "2001:db8:12::3",
+      "ifaceForKindLeaf": "toleafkind1",
+      "ifaceForKindLeaf2": "toleafkind2",
+      "leafIfaceForKindLeaf": "tokindctrlpl",
+      "leafIfaceForKindLeaf2": "tokindctrlpl"
+    },
+    "pe-kind-worker": {
+      "ipForKindLeaf": "192.168.11.4",
+      "ipForKindLeaf2": "192.168.12.4",
+      "ipv6ForKindLeaf": "2001:db8:11::4",
+      "ipv6ForKindLeaf2": "2001:db8:12::4",
+      "ifaceForKindLeaf": "toleafkind1",
+      "ifaceForKindLeaf2": "toleafkind2",
+      "leafIfaceForKindLeaf": "tokindworker",
+      "leafIfaceForKindLeaf2": "tokindworker"
+    }
+  }
+}
--- a/e2etests/pkg/config/update.go
+++ b/e2etests/pkg/config/update.go
@@ -26,11 +26,12 @@ type Resources struct {
 }
 
 type Updater struct {
-	cli       client.Client
-	namespace string
+	cli             client.Client
+	namespace       string
+	frrk8sNamespace string
 }
 
-func UpdaterForCRs(r *rest.Config, ns string) (*Updater, error) {
+func UpdaterForCRs(r *rest.Config, ns, frrk8sNs string) (*Updater, error) {
 	myScheme := runtime.NewScheme()
 
 	if err := v1alpha1.AddToScheme(myScheme); err != nil {
@@ -54,8 +55,9 @@ func UpdaterForCRs(r *rest.Config, ns string) (*Updater, error) {
 	}
 
 	return &Updater{
-		cli:       cl,
-		namespace: ns,
+		cli:             cl,
+		namespace:       ns,
+		frrk8sNamespace: frrk8sNs,
 	}, nil
 }
 
@@ -105,7 +107,13 @@ func (o Updater) Update(r Resources) error {
 
 	// Iterating over the map will return the items in a random order.
 	for i, obj := range objects {
-		obj.SetNamespace(o.namespace)
+		switch obj.(type) {
+		case *frrk8sv1beta1.FRRConfiguration:
+			obj.SetNamespace(o.frrk8sNamespace)
+		default:
+			obj.SetNamespace(o.namespace)
+		}
+
 		_, err := controllerutil.CreateOrUpdate(context.Background(), o.cli, obj, func() error {
 			// the mutate function is expected to change the object when updating.
 			// we always override with the old version, and we change only the spec part.
@@ -182,7 +190,7 @@ func (o Updater) CleanButUnderlay() error {
 		return err
 	}
 	if err := o.cli.DeleteAllOf(context.Background(), &frrk8sv1beta1.FRRConfiguration{},
-		client.InNamespace(o.namespace)); err != nil {
+		client.InNamespace(o.frrk8sNamespace)); err != nil {
 		return err
 	}
 	return nil
--- a/e2etests/pkg/config/update.go
+++ b/e2etests/pkg/config/update.go
@@ -26,11 +26,12 @@ type Resources struct {
 }
 
 type Updater struct {
-	cli       client.Client
-	namespace string
+	cli             client.Client
+	openpeNamespace string
+	frrk8sNamespace string
 }
 
-func UpdaterForCRs(r *rest.Config, ns string) (*Updater, error) {
+func UpdaterForCRs(r *rest.Config, openpeNs, frrk8sNs string) (*Updater, error) {
 	myScheme := runtime.NewScheme()
 
 	if err := v1alpha1.AddToScheme(myScheme); err != nil {
@@ -54,8 +55,9 @@ func UpdaterForCRs(r *rest.Config, ns string) (*Updater, error) {
 	}
 
 	return &Updater{
-		cli:       cl,
-		namespace: ns,
+		cli:             cl,
+		openpeNamespace: openpeNs,
+		frrk8sNamespace: frrk8sNs,
 	}, nil
 }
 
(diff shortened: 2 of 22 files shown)


e2etests/pkg/config/update.go
```diff
@@ -182,7 +185,7 @@ func (o Updater) CleanButUnderlay() error {
 		return err
 	}
 	if err := o.cli.DeleteAllOf(context.Background(), &frrk8sv1beta1.FRRConfiguration{},
-		client.InNamespace(o.namespace)); err != nil {
+		client.InNamespace(frrk8s.Namespace)); err != nil {
```
zeeke: can we add the frrk8s.Namespace field to the struct and pass it in the `UpdateForCRs(...)` ?

e2etests/suite/suite_test.go
```diff
@@ -34,6 +36,8 @@ func handleFlags() {
 	flag.BoolVar(&tests.HostMode, "systemdmode", false, "tells if openperouter is running on the host")
 	flag.BoolVar(&tests.GroutMode, "groutmode", false, "tells if openperouter is running with grout dataplane")
 	flag.BoolVar(&tests.SkipUnderlayPassthrough, "skip-underlay-passthrough", false, "skip creating underlay in passthrough tests")
+	flag.StringVar(&nodeLinkConfigPath, "nodelink-config", "../nodelink-default.json", "path to node links config JSON")
+	flag.StringVar(&frrk8s.Namespace, "frrk8s-namespace", frrk8s.Namespace, "namespace where FRR-K8s pods run")
```
zeeke, line 40: I think we should also have a flag to specify the openperouter namespace. As of now, it's hardcoded on L67

e2etests/pkg/config/update.go
oribon: done

e2etests/suite/suite_test.go
oribon: done

e2etests/tests/cni_lifecycle.go
zeeke: does this need to become `executor.ForNode(...)`

zeeke: same,
does this need to become executor.ForNode(...)

e2etests/pkg/executor/nodeexec.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	return nil, fmt.Errorf("no node-exec-helper pod found on node %s", nodeName)
+}
+
+func (e *nodeExecutor) Exec(cmd string, args ...string) (string, error) {
+	if Kubectl == "" {
+		return "", errors.New("the kubectl parameter is not set")
+	}
+	nsenterArgs := []string{"exec", e.podName, "-n", e.namespace, "-c", "nsenter", "--",
+		"chroot", "/proc/1/root", "nsenter", "-t", "1", "-m", "-u", "-i", "-n", cmd}
```
zeeke: mind using the long args for nsenter?

```suggestion
	nsenterArgs := []string{
		"exec", e.podName, "-n", e.namespace, "-c", "nsenter", "--",
		"chroot", "/proc/1/root",
		"nsenter",
		"--target", "1",
		"--mount", "-u",
		"--ipc",
		"--net",
		cmd}
```

e2etests/pkg/executor/nodeexec.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list helper pods: %w", err)
+	}
+	for i := range pods.Items {
+		if pods.Items[i].Spec.NodeName == nodeName {
+			return &nodeExecutor{namespace: forNodeNamespace, podName: pods.Items[i].Name}, nil
+		}
+	}
+	return nil, fmt.Errorf("no node-exec-helper pod found on node %s", nodeName)
+}
```
zeeke: just dropping an idea:
in this function we can just same the nodeName and then lazyloading the podName during the exec.
This would remove the `error` return variable and make the call site cleaner.

I'm not sure it's worth  the added complexity, though

zeeke (commented):
left few comments.

I'm not sure about the use of `busybox:1.36` (as it's not configurable), but I see we already have other tests using that image. So, I guess we can live with it

e2etests/pkg/executor/nodeexec.go
oribon: done (good idea, this also makes this more robust for pods deleting during test)

e2etests/tests/cni_lifecycle.go
oribon: yep (done)

oribon: yep (done)

e2etests/pkg/executor/nodeexec.go
oribon: done

oribon:
> I'm not sure about the use of `busybox:1.36` (as it's not configurable), but I see we already have other tests using that image. So, I guess we can live with it

added a `node-exec-image` flag that defaults to busybox to handle this, gives us some flexibility with minimal cost

zeeke (approved):
@fedepaol @maiqueb please take a look a this when you have time.

e2etests/pkg/config/update.go
```diff
@@ -26,11 +26,12 @@ type Resources struct {
...
 type Updater struct {
-	cli       client.Client
-	namespace string
+	cli             client.Client
+	namespace       string
+	frrk8sNamespace string
 }
 
-func UpdaterForCRs(r *rest.Config, ns string) (*Updater, error) {
+func UpdaterForCRs(r *rest.Config, ns, frrk8sNs string) (*Updater, error) {
```
maiqueb: I think that you need to disambiguate the "old" `ns` variable, given that now you add another namespace to the input attributes.

What's the "old" `ns` variable going to be used for now ? 
IMHO, the call `config.UpdaterForCRs(clientconfig, openperouter.Namespace, frrk8s.Namespace)` is strange to read - from the caller, IDK why we need to feed it two namespaces.

e2etests/suite/suite_test.go
maiqueb: one question though ... in a "real" deployment, what would this value be ? the "real" openshift-openperouter netns ?

e2etests/nodelink-default.json
```diff
@@ -0,0 +1,34 @@
+{
+  "nodes": {
+    "pe-kind-control-plane": {
+      "ipForKindLeaf": "192.168.11.3",
+      "ipForKindLeaf2": "192.168.12.3",
+      "ipv6ForKindLeaf": "2001:db8:11::3",
+      "ipv6ForKindLeaf2": "2001:db8:12::3",
+      "ifaceForKindLeaf": "toleafkind1",
+      "ifaceForKindLeaf2": "toleafkind2",
```
maiqueb, line 9: these 2 are the same for all 3 nodes.

Do we really need to parametrize them ? Will this grow in the future ? Do we envision a real need to change these values going forward ?

e2etests/suite/suite_test.go
```diff
@@ -73,6 +76,11 @@ var _ = ginkgo.BeforeSuite(func() {
 	tests.K8sReporter = reporter
 
 	cs := k8sclient.New()
+
+	ginkgo.By("Registering fabric and node links from " + nodeLinkConfigPath)
+	err = infra.RegisterLinks(nodeLinkConfigPath)
+	Expect(err).NotTo(HaveOccurred(), "failed to register links")
```
maiqueb: nit: you can inline these and use provide positive logic - i.e. `Expect(IPForKindLeaf).To(Succeed())`

e2etests/systemd_static_suite/suite_test.go
```diff
@@ -50,4 +54,12 @@ var _ = ginkgo.BeforeSuite(func() {
 	if kubeconfig == "" {
 		ginkgo.Fail("KUBECONFIG not set")
 	}
+
+	cs := k8sclient.New()
+	err := executor.SetupNodeExec(cs, frrk8s.Namespace, nodeExecImage)
+	Expect(err).NotTo(HaveOccurred(), "failed to setup node-exec-helper")
```
maiqueb:
```suggestion
	Expect(executor.SetupNodeExec(k8sclient.New(), frrk8s.Namespace, nodeExecImage)).To(Succeed(), "failed to setup node-exec-helper")
```

maiqueb (commented):
Just some questions.

I've asked @RamLavi to also review this PR.
Looks we're close.

e2etests/nodelink-default.json
oribon: per cluster, I expect these to be the same for all nodes (but the names themselves can be different across deployments). for the sake of explicitness and visibility (and some extensibility) I think it's nice to have them as part of the struct given that the cost is pretty minimial.

e2etests/suite/suite_test.go
oribon: done

e2etests/systemd_static_suite/suite_test.go
oribon: done

e2etests/pkg/config/update.go
oribon: the updater needs to distinguish between resources belonging to the openpe namespace (underlays, etc.) and those belonging to frr-k8s namespace (frrconfigurations), I renamed ns to openpeNamespace to reflect that

oribon:
> one question though ... in a "real" deployment, what would this value be ? the "real" openshift-openperouter netns ?

right, for frr-k8s it would be the ocp specific one controlled by cno, and for openpe whatever we decide (idk if we already have, openshift-* makes sense)

e2etests/pkg/config/update.go
maiqueb: OK, thank you for explaining.

We (CNV) have historically never concerned ourselves with the passthrough use cases, and as a result, I didn't even notice we monitored the `FrrConfiguration` CRs.

Makes sense now.

maiqueb (approved):
@andreaskaris could you take a look, please ?

e2etests/tests/evpn_l2.go
```diff
@@ -82,12 +81,12 @@ var _ = Describe("Routes between bgp and the fabric with Underlay in ipv4", Orde
 		})
 		Expect(err).NotTo(HaveOccurred())
 
-		// Create pre-existing OVS bridges on Kind nodes for testing
-		nodes, err = k8s.GetNodes(cs)
+		// Create pre-existing OVS bridges on cluster nodes for testing
+		ovsNodes, err := k8s.GetNodes(cs)
```
fedepaol, line 85: why this change?

e2etests/suite/suite_test.go
fedepaol: I guess whatever we deploy openperouter in, and it makes sense to make the name customizable

fedepaol (approved):
one small question, LGTM

e2etests/tests/evpn_l2.go
oribon: probably a leftover from the 10 rebases, when I opened the PR the kind node names were hardcoded here 😛 I can revert this now if you'd like

e2etests/nodelink-default.json
```diff
@@ -0,0 +1,34 @@
+{
```
andreaskaris, line 1: q: would yaml be better than json? less prone to typos, etc.? I'm indifferent, but elsewhere we use yaml configs (I wouldn't be surprised if we have a mix and yet in other locations we use json)

e2etests/pkg/infra/config.go
```diff
@@ -0,0 +1,47 @@
...
+	kindLeafIPv6  = "2001:db8:11::2"
+	kindLeaf2IPv6 = "2001:db8:12::2"
+)
+
+type nodeLinksConfig struct {
+	Nodes map[string]nodeConfig `json:"nodes"`
+}
+
+type nodeConfig struct {
+	IPForKindLeaf         string `json:"ipForKindLeaf"`
```
andreaskaris, line 23: nit: is this less ambiguous? "IPToKindLeaf"?   (because it's the IP on the node TO the kindleaf node)
Just because I had to think for a moment why we were having kindLeafIP and IPForKindLeaf

andreaskaris (approved):
/lgtm  a couple of questions / nits

maiqueb:
From my perspective, I am OK with merging this PR.

@oribon let me know if you want to address @andreaskaris 's nits, or if we should tackle them in a follow-up.

Tagging @fedepaol as well for visibility / asking his opinion.

oribon:
@maiqueb I prefer merging this and tackling in a follow-up, they are mostly about style and would require me to change some of the other stuff that I have using this 😛 (+ risking more rebases)

maiqueb:
> @maiqueb I prefer merging this and tackling in a follow-up, they are mostly about style and would require me to change some of the other stuff that I have using this 😛 (+ risking more rebases)

Let's get it in then.

PR #627, opened 2026-07-28
API: replace BFD passiveMode with a sessionMode enum

qinqon:

/kind feature

Implements part of the [Replace Boolean Fields](https://github.com/openperouter/openperouter/blob/main/enhancements/api-improvements.md#replace-boolean-fields)
section of the API improvements enhancement, for `BFDSettings`.

- `passiveMode` becomes a `sessionMode` enum, `Active` (the default when
  omitted) or `Passive`. Passive stays a BFD attribute rather than a BGP
  neighbor property: it is BFD `passive-mode`, which FRR accepts per
  peer, and is distinct from BGP `neighbor X passive`.
- `echoMode` and its companion `echoInterval` are removed rather than
  carried forward. Echo mode relies on the peer looping echo packets
  back through its forwarding plane, so it only works between FRR
  instances; against a physical fabric or ToR it is inert, and it is
  restricted to single hop sessions anyway. `echoInterval` only has
  meaning with echo enabled.

```yaml
# before
bfd:
  passiveMode: true
  echoMode: true
  echoInterval: 50

# after
bfd:
  sessionMode: Passive
```

N/S

**Release note**:

```release-note
The `passiveMode` field on `Underlay` `neighbors[].bfd` has been replaced by a `sessionMode` enum: replace `passiveMode: true` with `sessionMode: Passive`. The `echoMode` and `echoInterval` fields have been removed; BFD echo mode only works between FRR instances and is inert against a physical fabric.
```


RamLavi (approved):
LGTM

api/v1alpha1/neighbor.go
```diff
@@ -140,25 +154,11 @@ type BFDSettings struct {
...
-	// +kubebuilder:validation:Maximum:=60000
-	// +kubebuilder:validation:Minimum:=10
-	// +optional
-	EchoInterval *int32 `json:"echoInterval,omitempty"`
-
-	// echoMode enables or disables the echo transmission mode.
-	// This mode is disabled by default, and not supported on multi
-	// hops setups.
-	// +optional
-	EchoMode *bool `json:"echoMode,omitempty"`
```
maiqueb, line 155: for these I need approval from @fedepaol / @andreaskaris . 

Are these used for anything ? Do your use-cases require any of these attributes ?

maiqueb (approved):
LGTM from me, but I need to have validation from the telco side that we can actually remove these other knobs.

api/v1alpha1/neighbor.go
qinqon: @maiqueb this was commented on the enhancement, I remember @fedepaol commenting that we don't really need them, this is already cecided.

maiqueb: Can you link the relevant section in the enhancement ?

qinqon: https://github.com/openperouter/openperouter/blob/main/enhancements/api-improvements.md#replace-boolean-fields


On BFDSettings, **remove EchoMode (and its companion EchoInterval)** and replace PassiveMode *bool with a sessionMode enum. Passive stays a BFD attribute (it is BFD passive-mode, which FRR accepts per-peer; the hostcontroller already emits one BFD profile per neighbor, so all BFD knobs are per-neighbor) and is not a BGP neighbor property — it is distinct from BGP neighbor X passive. Echo mode relies on the peer looping echo packets through its forwarding plane and only works between FRR instances; against a physical fabric / ToR it is inert, and it is restricted to single-hop sessions anyway — so it is dropped rather than carried forward. **EchoInterval only has meaning with echo enabled and is removed with it**. Passive is the only remaining BFD mode, so it is modelled as a small enum (active is the default) rather than an over-engineered single-element set:

maiqueb: thank you !

maiqueb:
Seems we're just following what we said we would do in the enhancement.

PR #628, opened 2026-07-28
API: replace Neighbor.ebgpMultiHop with a properties list

qinqon:

/kind feature

Implements part of the [Replace Boolean Fields](https://github.com/openperouter/openperouter/blob/main/enhancements/api-improvements.md#replace-boolean-fields)
section of the API improvements enhancement, for `Neighbor.ebgpMultiHop`.

`ebgpMultiHop` becomes a session level `properties` list keyed by type.
It is modelled as a discriminated struct list rather than a plain enum
set because a property may need typed parameters: `ebgpMultiHop` takes
an optional `ttl`, which a string enum cannot express and which
Kubernetes structural schemas will not let us mix into one list. This
also keeps the configuration close to the FRR model, where toggles and
their parameters live on the neighbor.

```yaml
# before
neighbors:
  - asn: 64520
    address: 2001:db8:1234::1
    ebgpMultiHop: true

# after
neighbors:
  - asn: 64520
    address: 2001:db8:1234::1
    properties:
      - type: ebgpMultiHop
        ebgpMultiHop:
          ttl: 5          # optional, FRR defaults to 255
```

N/S

**Release note**:

```release-note
The `ebgpMultiHop` field on `Underlay` `neighbors[]` has been replaced by a session level `properties` list. Replace `ebgpMultiHop: true` with a `properties` entry of `type: ebgpMultiHop`, which now also accepts an optional `ttl` (1-255, FRR defaults to 255).
```

## Summary by CodeRabbit

* **New Features**
  * Added a structured neighbor properties model for configuring optional session features.
  * Added eBGP multihop configuration with an optional TTL value.
  * Generated FRR configuration now includes the configured multihop TTL.

* **Changes**
  * Replaced the legacy `ebgpMultiHop` boolean with the new `properties` format.
  * Updated CRDs, examples, and configuration manifests to use the new format.

* **Documentation**
  * Updated API reference and SRv6 documentation with the new configuration format and TTL behavior.


RamLavi:
@qinqon 
Right now it's possible to set `type: ebgpMultiHop` with no `ebgpMultiHop` struct at all - e.g.:
```yaml
properties:
  - type: ebgpMultiHop
```
vs.
```yaml
properties:
  - type: ebgpMultiHop
    ebgpMultiHop: {}
```
Both are valid today and mean the same thing (enabled, TTL defaults to 255 in FRR). I understand this is what the enhancement shows as well , but is that intentional?  i.e. do we want two ways to spell "enabled with default TTL"?

I would think that there should be a CEL rule to require `ebgpMultiHop` to be present (even if empty) whenever `type: ebgpMultiHop` is set, so there's exactly one way to express it... just my opinion though. wdyt?

p.s. I'm not trying to be phlosophical here, I just bumped into the question when I wanted to suggest a helper function and noticed we need to check twice for the actual existence, which I found weird, e.g.
```
  func ebgpMultiHopForNeighbor(n v1alpha1.Neighbor) (bool, *int32) {
        prop := findProperty(n.Properties, v1alpha1.NeighborPropertyEBGPMultiHop)
        if prop == nil {
                return false, nil
        }
        if prop.EBGPMultiHop == nil { // if type is already confirmed - why do we need to check if the struct is also there, is my point
                return true, nil
        }
        return true, prop.EBGPMultiHop.TTL
  }

```

qinqon:
> @qinqon Right now it's possible to set `type: ebgpMultiHop` with no `ebgpMultiHop` struct at all - e.g.:
> 
> ```yaml
> properties:
>   - type: ebgpMultiHop
> ```
> 
> vs.
> 
> ```yaml
> properties:
>   - type: ebgpMultiHop
>     ebgpMultiHop: {}
> ```
> 
> Both are valid today and mean the same thing (enabled, TTL defaults to 255 in FRR). I understand this is what the enhancement shows as well , but is that intentional? i.e. do we want two ways to spell "enabled with default TTL"?
> 
> I would think that there should be a CEL rule to require `ebgpMultiHop` to be present (even if empty) whenever `type: ebgpMultiHop` is set, so there's exactly one way to express it... just my opinion though. wdyt?
> 
> p.s. I'm not trying to be phlosophical here, I just bumped into the question when I wanted to suggest a helper function and noticed we need to check twice for the actual existence, which I found weird, e.g.
> 
> ```
>   func ebgpMultiHopForNeighbor(n v1alpha1.Neighbor) (bool, *int32) {
>         prop := findProperty(n.Properties, v1alpha1.NeighborPropertyEBGPMultiHop)
>         if prop == nil {
>                 return false, nil
>         }
>         if prop.EBGPMultiHop == nil { // if type is already confirmed - why do we need to check if the struct is also there, is my point
>                 return true, nil
>         }
>         return true, prop.EBGPMultiHop.TTL
>   }
> ```

No we do't want to force an empty strict this is an anty pattern that resonotes to kubevirt bridge binding things to have the default TTL we just set `Type: bgpMultiHop`.

api/v1alpha1/neighbor.go
```diff
@@ -114,6 +118,43 @@ type Neighbor struct {
...
+	// NeighborPropertyEBGPMultiHop enables eBGP multihop on the neighbor
+	// session, rendered as "neighbor X ebgp-multihop [ttl]".
+	NeighborPropertyEBGPMultiHop NeighborPropertyType = "ebgpMultiHop"
+)
+
+// EBGPMultiHopProperties holds parameters for the ebgpMultiHop property.
+type EBGPMultiHopProperties struct {
+	// ttl is the maximum number of hops for the eBGP multihop session.
+	// When omitted, FRR defaults to 255.
+	// +kubebuilder:validation:Minimum=1
```
maiqueb, line 149: is one really the minimum ? like a TTL of 1 is just one hop away, isn't it ? 

Is that considered multi-hop ?...

internal/conversion/frr_conversion.go
```diff
@@ -1003,6 +1006,21 @@ func bfdProfileForNeighbor(n v1alpha1.Neighbor) *frr.BFDProfile {
 	return bfdProfile
 }
 
+// ebgpMultiHopForNeighbor returns whether the ebgpMultiHop property is set on
+// the neighbor session, together with its optional TTL parameter.
+func ebgpMultiHopForNeighbor(n v1alpha1.Neighbor) (bool, *int32) {
```
maiqueb, line 1009: would it make sense to change the signature, returning the `v1alpha1.NeighborProperty` instead ? 

Then the caller would extract the TTL from it. 

Just a nit you can ignore.

maiqueb:
...

> No we do't want to force an empty strict this is an anty pattern that resonotes to kubevirt bridge binding things to have the default TTL we just set `Type: bgpMultiHop`.

Hm, why is this an anti-pattern ? I agree it's redundant. Really just want to understand what's wrong in that approach.

maiqueb:
@qinqon you need to rebase.

qinqon:
> @qinqon you need to rebase.

Done

internal/conversion/frr_conversion.go
```diff
@@ -1001,6 +1004,21 @@ func bfdProfileForNeighbor(n v1alpha1.Neighbor) *frr.BFDProfile {
...
```
RamLavi, line 1009: please consider using a generic helper function that does the lookup
```
  func ebgpMultiHopForNeighbor(n v1alpha1.Neighbor) (bool, *int32) {
        prop := findProperty(n.Properties, v1alpha1.NeighborPropertyEBGPMultiHop)
        if prop == nil {
                return false, nil
        }
        if prop.EBGPMultiHop == nil {
                return true, nil
        }
        return true, prop.EBGPMultiHop.TTL
  }
```

qinqon:
> ...
> 
> > No we do't want to force an empty strict this is an anty pattern that resonotes to kubevirt bridge binding things to have the default TTL we just set `Type: bgpMultiHop`.
> 
> Hm, why is this an anti-pattern ? I agree it's redundant. Really just want to understand what's wrong in that approach.

The "anti-pattern" was about using an empty struct instead of enum, but here is like doing the worse of both worlds using a type AND empty struct.

maiqueb:
> > ...
> > > No we do't want to force an empty strict this is an anty pattern that resonotes to kubevirt bridge binding things to have the default TTL we just set `Type: bgpMultiHop`.
> > 
> > 
> > Hm, why is this an anti-pattern ? I agree it's redundant. Really just want to understand what's wrong in that approach.
> 
> The "anti-pattern" was about using an empty struct instead of enum, but here is like doing the worse of both worlds using a type AND empty struct.

I was actually thinking that this would break the [discriminated union pattern](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions), but it seems it doesn't:
> Important to note:

    ...
    Empty union members (discriminant values without a paired union member) are also permitted

Let's roll with it.

internal/conversion/frr_conversion.go
qinqon: Done

internal/conversion/frr_conversion.go
qinqon: Then we will complicate this simple code 
```golang
ebgpMultiHop, ebgpMultiHopTTL := ebgpMultiHopForNeighbor(n)

	res := &frr.NeighborConfig{
....
		EBGPMultiHop:          ebgpMultiHop,
		EBGPMultiHopTTL:       ebgpMultiHopTTL,
...
	}

```

Into nil checking and the like.

api/v1alpha1/neighbor.go
qinqon: I think you are right but frr range is 1-255

```
neighbor <peer> ebgp-multihop [(1-255)] 
```

internal/conversion/frr_conversion.go
maiqueb: /shrug ok.

api/v1alpha1/neighbor.go
maiqueb: can you try to understand what would that mean ?

qinqon: Has same effect as disabling it, since it will not really allow multihop, multihop with your own hop is like no multihop :-)

maiqueb: perfect, thanks for the explanation.

maiqueb (approved):
Let's ship it.

PR #630, opened 2026-07-29
fix: apply the FRR configuration before the datapath

### LINKED ISSUE #604: bgpd SIGSEGV in bgp_evpn_local_vni_add (NULL EVPN instance) when a VNI is created
**Description**:

`bgpd` crashes with SIGSEGV in the systemdmode e2e lane. Two core dumps were captured in nightly run [29063549595](https://github.com/openperouter/openperouter/actions/runs/29063549595) (Jul 10, artifact `kind-logs-systemdmode`, `core_dumps/` directory). Both crashes are identical, so this is a deterministic NULL-pointer dereference, not random corruption.

### Crash summary

|  | Crash 1 | Crash 2 |
|---|---|---|
| Node | pe-kind-control-plane | pe-kind-worker |
| Time (UTC) | 2026-07-10 02:27:56 | 2026-07-10 02:43:30 |
| Signal | SIGSEGV (11) | SIGSEGV (11) |
| `si_addr` | `0x1ae0` | `0x1ae0` |
| Thread | `zclient_read` (scheduled from `lib/zclient.c:5021 zclient_event()`) | same |

FRR version: `10.6.0_git` (image `quay.io/openperouter/router:main`), cmdline `/usr/lib/frr/bgpd -d -F traditional -A 127.0.0.1 --limit-fds 100000`.

```
Jul 10 02:27:56 pe-kind-control-plane frr[4905]: BGP: Received signal 11 at 1783650476 (si_addr 0x1ae0, PC 0x55a040fe003c); aborting...
Jul 10 02:27:56 pe-kind-control-plane frr[4905]: BGP: in thread zclient_read scheduled from lib/zclient.c:5021 zclient_event()
```

### Faulting location (resolved from the core)

Resolving the faulting PC against the bgpd load base (`0x55a040f93000`) gives file offset `0x4d03c`:

```
$ addr2line -f -e /usr/lib/frr/bgpd 0x4d03c
bgp_filter_evpn_routes_upon_martian_change
/dist/src/frr-10.6.0_git/bgpd/bgp_evpn.c:6974

   4d03c: mov 0x1ae0(%rdi),%rax    <-- %rdi (struct bgp *) == NULL, offset 0x1ae0 == si_addr
```

So `bgp_filter_evpn_routes_upon_martian_change()` is invoked with a NULL `bgp` instance while bgpd processes a zebra notification (martian/interface address change) — i.e. the EVPN BGP instance is NULL/gone at that point.

### Context at crash time

Both crashes occur while the openperouter controller is churning VNI interfaces and reloading FRR:

- Crash 1: during `setting up VNI red` reconcile + frr-reload; immediately preceded by `ZEBRA: toswitch1(57): Tx RA failed, socket 14 error 99 (Address not available)`
- Crash 2: during `setting up L2VNI 300` reconcile

`watchfrr` restarts bgpd afterwards (`bgpd state -> down : read returned EOF`), so the suite usually keeps running — but all BGP sessions flap, which likely contributes to the intermittent session-establishment timeouts seen in the nightly e2e lanes.

### Suggested next steps

- Reproduce / report upstream to FRR (`bgp_evpn.c:6974`, NULL `bgp` in `bgp_filter_evpn_routes_upon_martian_change` — likely a missing NULL check on `bgp_get_evpn()` when the EVPN instance is being deleted/reconfigured during interface address churn)
- Check whether the openperouter reconcile flow (removing/re-adding VNIs while zebra address events are in flight) can avoid triggering the window, or whether an FRR version bump/patch fixes it
- Consider making the e2e collect `show bgp` crash info + failing the lane on daemon core dumps so these don't go unnoticed

- [x] I have reviewed the content of this issue, including any AI-generated text, and I take full responsibility for its accuracy and relevance.

qinqon:

/kind bug

The datapath created the kernel objects before the FRR configuration referencing them was reloaded. A VXLAN interface showing up while FRR has no EVPN instance makes zebra send `ZEBRA_VNI_ADD`, and bgpd handles it in `bgp_evpn_local_vni_add()`, which dereferences the instance returned by `bgp_get_evpn()` without checking it for NULL, so bgpd crashes.

`Reconcile()` ran `datapathConfigurator.Configure()` and only then `frrConfigurator()`, so the kernel objects existed for as long as the FRR reload takes (~1.5s in CI) before the configuration declaring the EVPN instance was applied. Reloading the FRR configuration first closes that window.

The crash analysis and the CI occurrences are in #604, and the missing NULL check was reported upstream in https://github.com/FRRouting/frr/issues/22851 and fixed by [fc7ccc3](https://github.com/FRRouting/frr/commit/fc7ccc32aca5d8f9dd2db7bb52b0efdadfa2b421). That fix is in FRR `master` only, it is not in any release yet, and we are pinned to `frr:10.6.0`, so this reordering is what keeps us from hitting the crash until a backport is released.

Fixed #604

The second commit fixes a pre-existing bug that this reordering exposed. We were generating `disable-connected-check` for neighbors configured over an interface, which FRR rejects, so every reload of an unnumbered underlay failed:

```
toleafkind1 is directly connected peer, cannot accept disable-connected-check
line 2: Failure to communicate[13] to bgpd
```

That went unnoticed because the datapath was configured before the reload, so the interfaces were already in place and the sessions came up regardless, with the reconcile error only being requeued. The same message is present in the nightly artifacts of Jul 24, 25 and 27, before this branch. With the reordering the reload failure became fatal, since the datapath is no longer configured when FRR rejects the configuration.

Removals stay correct in this order: `ZEBRA_VNI_DEL` is handled by `bgp_evpn_local_vni_del()`, which does not touch the EVPN instance, and FRR stops referencing the objects before they are deleted.

The reordering is safe because `internal/conversion` builds the FRR configuration from the API config and the node index only, with no netlink reads, and the objects it names (`vrf`, `vni`, `router bgp ... vrf`, `interface`) are accepted by FRR before they exist and bind once they show up.

Two behaviour changes come with it:

- when the FRR reload fails, the datapath is no longer configured, which is what keeps the datapath from getting ahead of FRR. `TestReconcileSkipsDatapathOnFrrReloadFailure` covers it.
- when the datapath fails, FRR has already been reloaded. The configuration is declarative, so the next reconcile converges.

`TestReconcileAppliesFRRConfigBeforeDatapath` pins the ordering, since it is an invariant that a refactor could silently drop. Both tests fail without the fix.

The zebra crashes seen in the same runs are a different race and are not addressed here, they are tracked in #632.

**Release note**:

```release-note
Fixed a crash of the FRR bgpd daemon that could happen while VNIs were being created, by applying the FRR configuration before the corresponding kernel objects.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved routing configuration sequencing to apply FRR settings before datapath configuration, preventing potential VXLAN/BGP setup failures.
  * Datapath configuration is now skipped when FRR reload/configuration fails.
  * Corrected connected-check handling for unnumbered IPv6 eBGP neighbor sessions.

* **Tests**
  * Added new test coverage for configuration ordering and for skipping datapath steps on FRR failures.
  * Added a golden test for unnumbered IPv6 eBGP configuration generation.

### DIFF
--- a/internal/controller/routerconfiguration/reconcile.go
+++ b/internal/controller/routerconfiguration/reconcile.go
@@ -110,16 +110,13 @@ func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, nodeInde
 		RawFRRConfigs: apiConfig.RawFRRConfigs,
 	}
 
-	err = datapathConfigurator.Configure(ctx, interfacesConfiguration{
-		targetNamespace: targetNamespace,
-		APIConfigData:   config,
-		nodeIndex:       nodeIndex,
-	})
-	if openpeerrors.IsNonResourceError(err) {
-		return err
-	}
-	resourceErrors = append(resourceErrors, err)
-
+	// The FRR configuration must be applied before the datapath creates the kernel
+	// objects it references. bgpd handles ZEBRA_VNI_ADD in bgp_evpn_local_vni_add(),
+	// which dereferences the instance returned by bgp_get_evpn() without checking it
+	// for NULL, so a VXLAN interface showing up while FRR has no EVPN instance
+	// configured crashes bgpd. See https://github.com/FRRouting/frr/issues/22851.
+	// Removals keep working in this order because ZEBRA_VNI_DEL does not touch the
+	// EVPN instance, and FRR stops referencing the objects before they are deleted.
 	if err = frrConfigurator(ctx, frrConfigData{
 		configFile:    frrConfigPath,
 		updater:       updater,
@@ -129,6 +126,16 @@ func Reconcile(ctx context.Context, apiConfig conversion.APIConfigData, nodeInde
 	}); err != nil {
 		return err
 	}
+
+	err = datapathConfigurator.Configure(ctx, interfacesConfiguration{
+		targetNamespace: targetNamespace,
+		APIConfigData:   config,
+		nodeIndex:       nodeIndex,
+	})
+	if openpeerrors.IsNonResourceError(err) {
+		return err
+	}
+	resourceErrors = append(resourceErrors, err)
 
 	return errors.Join(resourceErrors...)
 }
--- a/internal/controller/routerconfiguration/reconcile_test.go
+++ b/internal/controller/routerconfiguration/reconcile_test.go
@@ -557,6 +557,62 @@ func TestReconcileFrrReloadFailure(t *testing.T) {
 	}
 }
 
+// recordingDatapathConfigurator appends to a shared list every time the datapath
+// is configured, so that tests can assert the order of the reconcile steps.
+type recordingDatapathConfigurator struct {
+	steps *[]string
+}
+
+func (r *recordingDatapathConfigurator) Configure(_ context.Context, _ interfacesConfiguration) error {
+	*r.steps = append(*r.steps, "datapath")
+	return nil
+}
+
+func (r *recordingDatapathConfigurator) Validate(_ conversion.APIConfigData) error {
+	return nil
+}
+
+// TestReconcileAppliesFRRConfigBeforeDatapath pins the ordering between the FRR
+// configuration and the datapath. The kernel objects must only be created once FRR
+// already references them: bgpd handles ZEBRA_VNI_ADD in bgp_evpn_local_vni_add(),
+// which dereferences bgp_get_evpn() without a NULL check, so a VXLAN interface
+// created while FRR has no EVPN instance crashes bgpd.
+// See https://github.com/FRRouting/frr/issues/22851.
+func TestReconcileAppliesFRRConfigBeforeDatapath(t *testing.T) {
+	var steps []string
+	recordingConfigureFRR := func(_ context.Context, _ frrConfigData) error {
+		steps = append(steps, "frr")
+		return nil
+	}
+
+	reconcileErr := Reconcile(context.Background(), conversion.APIConfigData{}, 0, "",
+		"", "", noopUpdater, &recordingDatapathConfigurator{steps: &steps}, recordingConfigureFRR)
+	if reconcileErr != nil {
+		t.Fatalf("unexpected reconcile error: %v", reconcileErr)
+	}
+
+	want := []string{"frr", "datapath"}
+	if diff := cmp.Diff(steps, want); diff != "" {
+		t.Errorf("reconcile steps out of order (-got, +want):\n%s", diff)
+	}
+}
+
+// TestReconcileSkipsDatapathOnFrrReloadFailure ensures no kernel object is created
+// when FRR could not be configured, so that the datapath never gets ahead of FRR.
+func TestReconcileSkipsDatapathOnFrrReloadFailure(t *testing.T) {
+	var steps []string
+
+	reconcileErr := Reconcile(context.Background(), conversion.APIConfigData{}, 0, "",
+		"", "", failingUpdater, &recordingDatapathConfigurator{steps: &steps}, configureFRR)
+	if reconcileErr == nil {
+		t.Fatal("expected error from FRR reload failure")
+	}
+
+	if len(steps) != 0 {
+		t.Errorf("expected the datapath not to be configured, got steps: %v", steps)
+	}
+}
+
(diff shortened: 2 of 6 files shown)


maiqueb (changes requested):
Seems OK, although I don't understand what the second commit has to do with this.

Taking it a step further, I actually don't understand why it is wrong - afaiu, when you use BGP unnumbered, you can only peer using link-local addresses, which cannot be routed beyond the local wire - the connection is guaranteed to be directly connected

IIUC, this would make the command redundant - but not wrong, or even lead to failures elsewhere.

qinqon:
> Seems OK, although I don't understand what the second commit has to do with this.
> 
> Taking it a step further, I actually don't understand why it is wrong - afaiu, when you use BGP unnumbered, you can only peer using link-local addresses, which cannot be routed beyond the local wire - the connection is guaranteed to be directly connected
> 
> IIUC, this would make the command redundant - but not wrong, or even lead to failures elsewhere.

Problem is that at FRR the combination of unnumered + disable-connect-check  it's not a no-op it rejected:

```c
/*
 * If 'neighbor <interface>', then this is for directly connected peers,
 * we should not accept disable-connected-check.
 */
if (peer->conf_if && (flag == PEER_FLAG_DISABLE_CONNECTED_CHECK)) {
	vty_out(vty, "%s is directly connected peer, cannot accept disable-connected-check\n", ip_str);
	return CMD_WARNING_CONFIG_FAILED;
}
```

It fails the same way whether or not the interface exists, since the check is only on peer->conf_if. And because we push the whole file through frr-reload.py, that one rejected line fails the entire reload:

```
toleafkind1 is directly connected peer, cannot accept disable-connected-check
line 2: Failure to communicate[13] to bgpd, line:  neighbor toleafkind1 disable-connected-check
[bgpd] Configuration file[/etc/frr/frr.conf] processing failure: 13
```

Why we never noticed. The unnumbered e2e cases do not check RouterNodeConfigurationStatus, so we were not seeing that they were already failing on main with:

{"type":"Ready","status":"False","reason":"ConfigurationFailed",
 "message":"failed to update the frr configuration: failed to reload against socket ... status 500"}
{"type":"Degraded","status":"True", ...}

This happens 40 to 80 times per nightly run (helm 40, manifests 44, systemdmode 80, all before this branch). The tests still passed because we configured the underlay before the FRR reload, so the interfaces were already there and the BGP sessions came up regardless of the failed reload.

Why it has to be in this PR. Now that the FRR configuration is applied first, a rejected configuration means we never configure the underlay, so the session cannot come up and the test fails. That is what made manifests, operator, systemdmode and helm fail on Unnumbered / passthrough and frr-k8s in the first push of this branch, and what the second commit fixes.

But we can move this it's own PR, although it did make sense to put it all together too.

maiqueb (approved):
Thank you.

fedepaol:
Did we file an issue against FRR / checked if this was fixed in newer FRR versions?

I am asking because inverting the order might raise new bugs which didn't happen before. In general, I would expect FRR to be happier if the datapath is laid down before applying a configuration.

fedepaol:
Ah, I see the issue filed (and fixed!). Still, my considerations apply. Given how FRR behaves, we are exposing ourselves to new unexplored behaviors. Maybe it'd be worth consuming the fix.

qinqon:
> Ah, I see the issue filed (and fixed!). Still
 
So as part of flaky CI investigations we openned a pair of issues, one issue for bgpd y and another for zebra, the one related to this PR is the bgpd one, looks like it got fixed at frr main and also backportes to `stable/10.7`, so we will need bump from 10.6 to 10.7.
- https://github.com/FRRouting/frr/issues/22852

> Given how FRR behaves, we are exposing ourselves to new unexplored behaviors. Maybe it'd be worth consuming the fix.

Let's monitor how CI looks like, and in parallel open PR to bump to 10.7.x.

About changing behaviour before doing this I tested the 30 configs we generate apply cleanly via frr-reload.py --test against a live FRR where only lo and eth0 exist, and they did look fine, but let's monitor the CI.

UPDATE: Is super early but I have check the nightly and PRs CI (10 runs, not much) after merging this and there is no coredumps zebra or bgpd.

PR #639, opened 2026-08-03
[DNM] route reflector CI run with patched FRR (10.6.0 + FRRouting/frr#22858)

qinqon:
**DO NOT MERGE** — CI experiment only.

This is PR #509 (route reflector) with a single extra commit that builds the
router image on FRR 10.6.0 plus [FRRouting/frr#22858](https://github.com/FRRouting/frr/pull/22858),
to check whether the e2e failures seen on #509 are all caused by that zebra bug.

### The bug

`zl3vni`/`zevpn` cache raw `struct interface *` that are only cleared through
`zebra_vxlan_svi_down()`, which is unreachable once the parent bridge is freed.
Deleting a bridge together with its SVI leaves them dangling, and openperouter
deletes and recreates that datapath on every config change.

The stale pointer kills zebra two ways: a dereference (SIGSEGV in
`if_is_operative()` / `neigh_update_internal()`), or, with `debug zebra kernel`
enabled as it is in CI, the assert in `ipaddr_family()` while formatting a
garbage `vtep_ip` (SIGABRT). watchfrr restarts zebra inside the container, so
the pod restart count stays at 0, but the node never registers the L3VNI again
and fabric type-5 routes stop importing into the VRF for the rest of the run.

Reproduced under ASAN and confirmed fixed by #22858:
https://github.com/FRRouting/frr/pull/22858#issuecomment-5130131005

Related: FRRouting/frr#22852, FRRouting/frr#21794 (ovn-kubernetes CI, same 10.6.0).

### Baseline being compared against

Run [30523670618](https://github.com/openperouter/openperouter/actions/runs/30523670618) on #509:

| Lane | zebra core (SIGABRT) | Failures | First failure |
| --- | --- | --- | --- |
| systemdmode | 08:02:09 | 1 | 08:16:00 |
| operator | 08:08:26 | 4 | 08:09:17 |
| helm | 08:11:38 (two nodes) | 6 | 08:13:53 |
| manifests | 08:37:22 | 4 | 08:50:16 |

5 crashes, 15 failures, none of them before a crash on their own lane. On the
crashed node the router logs show L2 VNI adds continuing while `Rx L3VNI ADD`
never appears again.

The same specs also fail on main, e.g. the Jul 27 nightly operator lane
(`SRV6 ... without gatewayIPs`, `Single Session Baseline`, `iBGP ...`,
`MacvlanStatic ...`, `... two pods l2 overlay`), so this is not specific to #509.
#509 only raises how often it happens, by adding a third node and two more
underlay/VNI lifecycles.

### What the result means

- Green, no cores: the failures on #509 are all this zebra bug.
- Cores gone but specs still failing: there is a second cause to chase.

Failures here are probabilistic (roughly one core per lane), so a single green
run is suggestive rather than conclusive; expect a couple of re-runs.

### Not for merge

The base image is a personal build and the clab fabric routers deliberately stay
on stock 10.6.0. The real fix is #22858 landing upstream and being backported to
`stable/10.6`.

## Summary by CodeRabbit

* **New Features**
  * Added BGP route-reflector support for underlay and EVPN routes.
  * Added dynamic BGP neighbor discovery through IPv4 and IPv6 listen ranges.
  * Added configurable BGP dynamic-session limits, defaulting to 65,535.
  * Added route-reflector and listen-range configuration options with validation.
* **Documentation**
  * Added API and configuration guidance, including a complete EVPN route-reflector example.
* **Tests**
  * Added coverage for route reflection, dynamic neighbors, session limits, and multi-worker connectivity.


qinqon:
## Result: the patch is not enough

Run [30793138449](https://github.com/openperouter/openperouter/actions/runs/30793138449), with the router image built on FRR 10.6.0 + FRRouting/frr#22858.

| | Baseline [30523670618](https://github.com/openperouter/openperouter/actions/runs/30523670618) | This run (patched FRR) |
| --- | --- | --- |
| zebra cores | 5, all SIGABRT | 3, all SIGABRT |
| e2e failures | 15 | 5 |
| e2e lanes red | 4/4 | 2/4 (helm, manifests) |

The patch was really in use: the image CI built and loaded contains `zebra_vxlan_if_ref_cleanup`, the symbol #22858 introduces.

The remaining cores have a backtrace **identical** to the pre-patch ones:

```
dplane_thread_loop -> kernel_dplane_process_func -> kernel_update_multi
  -> netlink_batch_add_msg -> netlink_macfdb_update_ctx
  -> zlog_ref -> ... -> printfrr_ia -> _zlog_assert_failed -> abort
```

So the garbage `vtep_ip.ipa_type` that trips the assert in `ipaddr_family()` comes from a different stale/uninitialised source than the interface pointers #22858 clears. I had assumed both signatures shared one root cause; that was wrong, and I have corrected it upstream.

### Careful about the improvement

15 failures down to 5 looks good but is probably luck, not the patch. Every core in both runs is the same SIGABRT class, and the patch demonstrably does not remove it. What changed is *where* the crashes landed: two of this run's three cores are on `pe-kind-worker2`, the node fewest specs touch, and the third hit control-plane late in the run. Blast radius depends on which node dies and when.

### Where this leaves things

- FRRouting/frr#22858 still fixes the use-after-free I reproduced under ASAN (SIGSEGV in `if_is_operative()`), and is still worth landing and backporting.
- It does **not** fix the `netlink_macfdb_update_ctx` assert, which is the signature that actually dominates our CI. That needs its own upstream issue and fix.
- Pinning a patched FRR image will therefore not make CI green on its own.

### Also worth noting

`e2etests (systemdmode)` and `e2etests (operator)` passed, both of which failed on the baseline. `e2etests (helm-grout)` failed on an unrelated spec; the grout image installs FRR from a DPDK/grout release RPM (`Dockerfile.grout`), so it is not affected by `FRR_IMAGE` and is outside this experiment.

Keeping this open until the macfdb assert has an upstream fix to test.

PR #643, opened 2026-08-03
Dhcp ipam lease release

maiqueb:

>

/kind bug

> /kind cleanup

> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

This PR pins CNI plugins to containernetworking/plugins@012159164d7f which includes the fix for [containernetworking/plugins#1278](https://github.com/containernetworking/plugins/issues/1278):
  - the DHCP plugin now sends DHCPRELEASE with the correct client IP as ciaddr, so dnsmasq actually honours the release.
  - Switch Dockerfile clone strategy from `git clone --branch` (which doesn't accept a commit SHA) to a shallow `git fetch` by SHA.
  - Update the e2e DHCP lifecycle test to assert the lease is gone after pod deletion, replacing the previous workaround that expected the lease to persist due to the upstream bug.

Once containernetworking/plugins cuts a release that includes the fix, we should bump back to a proper tag.

**Release note**:

```release-note
Fix DHCP lease cleanup on pod deletion by consuming the upstream ciaddr fix (containernetworking/plugins#1279).
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network component consistency by using fixed, reproducible CNI versions during environment setup.
  * Updated the default CNI plugin version to improve compatibility and reliability.
  * Corrected DHCP lifecycle handling so leases are expected to be removed immediately when the underlying network is deleted.
* **Tests**
  * Updated lifecycle validation to reflect the corrected DHCP cleanup behavior and provide more accurate results.


Dockerfile
```diff
@@ -9,7 +9,10 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /cni-plugins
-RUN git clone --branch ${CNI_PLUGINS_VERSION} --depth 1 https://github.com/containernetworking/plugins.git .
+RUN git init && \
+    git remote add origin https://github.com/containernetworking/plugins.git && \
+    git fetch --depth 1 origin ${CNI_PLUGINS_VERSION} && \
+    git checkout FETCH_HEAD
```
qinqon, line 15: nit: I have found out that we can directly download the tarball, so w just need to curl/unpack it. 

https://github.com/containernetworking/plugins/archive/012159164d7f552ee7a8ee840447c61611958e87.tar.gz

qinqon (changes requested):
just a nit to use curl directly.

Dockerfile
maiqueb: I just want to consume the proper binary, and to me, the safest way, is to build from scratch.

Feel free to push another PR to change this.

qinqon: > I just want to consume the proper binary, and to me, the safest way, is to build from scratch.
> 
> Feel free to push another PR to change this.

The tarball contains the source code to build not the binary, but I will not press on this it was just a interesting finding

qinqon (approved):
LGTM

PR #654, opened 2026-08-05
fix(controller): handle CNI config mismatch on underlay recreate

fedepaol:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

Error found in   [FAIL] Routes between bgp and the fabric with Underlay in ipv4 MacvlanDHCP with vnis and frr-k8s [BeforeEach] translates EVPN incoming routes as BGP routes

https://github.com/openperouter/openperouter/actions/runs/30977939400/job/92217350377?pr=509

When an underlay is deleted and recreated with a different CNI config (e.g. static -> dhcp IPAM) for the same interface name, the reconciler can miss the cleanup path. The test deletes old underlays and creates new ones sequentially, but the reconciler is async: by the time it processes the queued delete event, getConfigFromAPI() reads current state which already has the new underlays. Since len(config.Underlays) > 0, restoreUnderlay() is never called and the stale CNI cache entry is never removed.

From the CI logs on worker2: the reconcile for underlay-cni-1 (deleted) found underlay-dhcp-* already in the API. SetupUnderlay tried to provision "underlay0" with DHCP config but the cache held the old static config, producing 265 ConfigMismatchError loops that prevented VRF creation and blocked all BGP sessions on that node.

Handle this by deleting the old attachment and retrying Add when a config mismatch is detected.

The way we handle the scenario is debatable, and the scenario itself is related to a very specific edge case which is, quick deletion and recreation of two underlays with the same interface name. We could probably workaround this by using a different interface name per test

**Release note**:

```release-note
Bug: if the delete of a previous underlay and the add of a new one are squashed together, and both use CNI but with a different configuration, the underlay is not processed.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically detects mismatched network interface configurations.
  * Removes stale interface settings and provisions the interface again with the updated configuration.
  * Reports errors if cleanup or re-provisioning fails.

* **Tests**
  * Added coverage for updating an interface after its configuration changes.

### DIFF
--- a/internal/hostnetwork/underlay_cnidev_test.go
+++ b/internal/hostnetwork/underlay_cnidev_test.go
@@ -243,6 +243,33 @@ var _ = Describe("Underlay CNI configuration", func() {
 		Expect(loggedCommands()).To(ConsistOf("ADD net1", "DEL net1"))
 	})
 
+	It("rebuilds the cni interface when the config changes", func() {
+		Expect(SetupUnderlay(context.Background(), cniParams("net1"))).To(Succeed())
+
+		// Build a second config with a different network name so the
+		// cached attachment does not match, simulating a delete-and-
+		// recreate with different IPAM (e.g. static -> dhcp).
+		altConfig := fmt.Sprintf(`{
+		  "cniVersion": "1.0.0",
+		  "name": "underlay-cni-alt",
+		  "plugins": [{"type": %q}]
+		}`, underlayCNITestPlugin)
+		altParams := UnderlayParams{
+			TargetNS: underlayCNITestNSPath(),
+			UnderlayInterfaces: []UnderlayInterface{{
+				InterfaceName: "net1",
+				Kind:          UnderlayInterfaceCNIDev,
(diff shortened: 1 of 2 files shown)


internal/hostnetwork/underlay_cnidev_test.go
```diff
@@ -243,6 +243,33 @@ var _ = Describe("Underlay CNI configuration", func() {
 		Expect(loggedCommands()).To(ConsistOf("ADD net1", "DEL net1"))
 	})
 
+	It("rebuilds the cni interface when the config changes", func() {
```
maiqueb, line 246: I think we literally said we don't want to handle configuration changes in CNI interfaces in the enhancement. 

While I understand that in an eventually consistent world a DELETE + a CREATE (the same key, but with different config) is an UPDATE, and that our docs instruct the user to re-create the configuration, I think we need to understand better what we want to do.

Issue #659, opened 2026-08-05
Flake: Routes between bgp and the fabric with Underlay in ipv4 MacvlanDHCP with vnis and frr-k8s


fedepaol:
Error found in [FAIL] Routes between bgp and the fabric with Underlay in ipv4 MacvlanDHCP with vnis and frr-k8s [BeforeEach] translates EVPN incoming routes as BGP routes

https://github.com/openperouter/openperouter/actions/runs/30977939400/job/92217350377?pr=509

When an underlay is deleted and recreated with a different CNI config (e.g. static -> dhcp IPAM) for the same interface name, the reconciler can miss the cleanup path. The test deletes old underlays and creates new ones sequentially, but the reconciler is async: by the time it processes the queued delete event, getConfigFromAPI() reads current state which already has the new underlays. Since len(config.Underlays) > 0, restoreUnderlay() is never called and the stale CNI cache entry is never removed.

From the CI logs on worker2: the reconcile for underlay-cni-1 (deleted) found underlay-dhcp-* already in the API. SetupUnderlay tried to provision "underlay0" with DHCP config but the cache held the old static config, producing 265 ConfigMismatchError loops that prevented VRF creation and blocked all BGP sessions on that node.

PR #660, opened 2026-08-05
fix(e2e): use distinct CNI interface names per underlay flavor

fedepaol:

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
/kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

Related to https://github.com/openperouter/openperouter/issues/659 , which requires broader discussion, but removes the flake. The scenario is very unlikely to happen.

When a static CNI underlay test runs before a DHCP one, event coalescing can cause the reconciler to miss the cleanup of the old CNI cache entry. If both flavors use the same interface name, the stale cache blocks provisioning with ConfigMismatchError.

Give each flavor its own interface name (ul-static, ul-dhcp, ul-rename) so UnderlayInterfacesToRemove detects the old name and cleans it up regardless of event ordering.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/tests/cni_lifecycle.go
+++ b/e2etests/tests/cni_lifecycle.go
@@ -25,9 +25,11 @@ import (
 )
 
 const (
-	// cniUnderlayInterfaceRenamed is used to exercise the in-place
+	lcUnderlayStaticInterface = "lc-static"
+	lcUnderlayDHCPInterface   = "lc-dhcp"
+	// cniUnderlayInterfaceRenamed is used to exercise the
 	// reprovisioning flow by changing the desired CNI interface name.
-	cniUnderlayInterfaceRenamed = "underlay1"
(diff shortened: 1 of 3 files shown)


e2etests/tests/cni_lifecycle.go
```diff
@@ -25,9 +25,11 @@ import (
 )
 
 const (
-	// cniUnderlayInterfaceRenamed is used to exercise the in-place
+	lcUnderlayStaticInterface = "lc-static"
+	lcUnderlayDHCPInterface   = "lc-dhcp"
```
maiqueb, line 29: what does lc stand for ?

fedepaol: lifecycle!

maiqueb (changes requested):
I think the commit msg and the commit's contents do not match.

PR #663, opened 2026-08-05
test(e2e): fix resiliency and frr_restart pod readiness check

### LINKED ISSUE #661: e2e: frr restart and resiliency tests, check "waiting for the FRR container to restart and become ready" is ignored
Check "waiting for the FRR container to restart and become ready" in resiliency.go and frr_restart.go is ignored (at least during local reproduction) because after killing, the tests nearly immediately wait for pod ready. That status may erroneously still be reported as ready: true due to timing issues. We must make sure to Eventually() hit a Consistent ready state, e.g. with MustPassRepeated. In addition, even though we do not hit this in CI, we should take into account that kubernetes' max crashloopbackoff is hardcoded to 5 minutes. Therefore, the timeout must be North of 5 minutes (currently, it is 2 minutes for the pod ready [which immediately falls through] and exactly 5 minutes for the bgp session test which is not enough).

This can be reproduced locally with:

```
export GINKGO_ARGS='--focus="Alpha: Named netns and kernel objects survive FRR crash should preserve named netns"'; while make e2etests; do echo -e "\n\n\n=== Test succeeded, running it again ===\n\n\n"; done
```

At some point, the backoff becomes so long that the test will fail.

andreaskaris:

/kind flake

- **test(e2e): fix resiliency and frr_restart pod readiness check**
- **test(e2etest): resiliency.go and frr_restart.go Expect().To(Succeed())**

Check "waiting for the FRR container to restart and become ready" in
resiliency.go and frr_restart.go was ignored (at least during local
reproduction) because after killing, the tests nearly immediately waited
for pod ready. That status may erroneously still be reported as
ready: true due to timing issues.
This commit refactors the redundant code into a function, makes sure to
use MustPassRepeated so that the ready state is valid for at least 10
seconds. In addition, take into account that kubernetes' max
crashloopbackoff is hardcoded to 5 minutes. Therefore, the timeout must
be North of 5 minutes (6 minutes were chosen for this fix).

Fixes https://github.com/openperouter/openperouter/issues/661

Before this PR, command:

```
export GINKGO_ARGS='--focus="Alpha: Named netns and kernel objects survive FRR crash should preserve named netns"'; while make e2etests; do echo -e "\n\n\n=== Test succeeded, running it again ===\n\n\n"; done
```

triggers repeated crashloopbackoffs (counters IIRC are cleared only after 10 minutes) and after a while test runs will fail. With the fix, this does not happen.

Though tbh I'm not sure how useful this fix is ...

**Release note**:

```release-note
None
```

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test reliability by consistently verifying pod readiness after router restarts.
  * Added automated DaemonSet pod restart and readiness checks.
  * Strengthened validation for FRR restart, namespace recovery, and stretched-L2 resiliency scenarios.
  * Simplified test assertions and cleanup checks to provide clearer failure reporting.

### DIFF
--- a/e2etests/pkg/k8s/pods.go
+++ b/e2etests/pkg/k8s/pods.go
@@ -213,6 +213,27 @@ func PodIsReady(p *corev1.Pod) bool {
 		podConditionStatus(p, corev1.ContainersReady) == corev1.ConditionTrue
 }
 
+// WaitForPodReadyConsistently waits until the provided pod is consistently ready.
+// We make sure that the pod is consistently ready 10 consecutive times (roughly 10'ish seconds) to avoid that we catch
+// an invalid ready state early in case the pod is transitioning from ready to not ready.
+// Note: kubernetes' MaxCrashLoopBackOff is hardcoded to 5 minutes:
+// https://github.com/kubernetes/kubernetes/blob/ec9448ceef13adbbcd3fa9f09b4d3c1fa54cc674/pkg/kubelet/kubelet.go#L166
+// https://github.com/kubernetes/kubernetes/issues/57291
+// Therefore, before calling this function, make sure that the crashloop timer is reset, e.g. by a new rollout of the
+// pod.
+func WaitForPodReadyConsistently(cs clientset.Interface, podNamespace, podName string) {
+	By(fmt.Sprintf("waiting for pod %s/%s to become ready", podNamespace, podName))
+	Eventually(func(g Gomega) bool {
+		pod, err := cs.CoreV1().Pods(podNamespace).Get(context.Background(), podName, metav1.GetOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		return PodIsReady(pod)
+	}).
+		WithTimeout(2 * time.Minute).
+		WithPolling(time.Second).
+		MustPassRepeatedly(10).
+		Should(BeTrue())
+}
+
 func PodsForLabel(cs clientset.Interface, namespace, labelSelector string) ([]*corev1.Pod, error) {
 	pods, err := cs.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{
 		LabelSelector: labelSelector,
--- a/e2etests/pkg/openperouter/openperouter.go
+++ b/e2etests/pkg/openperouter/openperouter.go
@@ -3,20 +3,28 @@
 package openperouter
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"iter"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
 
 	"github.com/openperouter/openperouter/e2etests/pkg/executor"
 	"github.com/openperouter/openperouter/e2etests/pkg/k8s"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


andreaskaris:
@CodeRabbit review

e2etests/pkg/openperouter/openperouter.go
```diff
@@ -114,3 +122,85 @@ func DaemonsetRolled(oldRouters Routers, newRouters Routers) error {
...
+	By(fmt.Sprintf("waiting for DaemonSet %s/%s to become ready", namespace, name))
+	Eventually(func(g Gomega) bool {
+		ds, err := cs.AppsV1().DaemonSets(namespace).Get(context.Background(), name, metav1.GetOptions{})
+		g.Expect(err).NotTo(HaveOccurred())
+		return ds.Status.DesiredNumberScheduled > 0 && ds.Status.DesiredNumberScheduled == ds.Status.NumberReady
+	}).
+		WithTimeout(2 * time.Minute).
+		WithPolling(time.Second).
+		MustPassRepeatedly(10).
+		Should(BeTrue())
```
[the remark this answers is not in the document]
andreaskaris: correct. but we're deleting the daemonset pods precisely to minimize the risk of running into a long crash-loop backoff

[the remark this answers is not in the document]
andreaskaris: You're assuming that the pod's containers crash after the restart. With that assumption, they might as well crash indefinitely

e2etests/pkg/openperouter/openperouter.go
```diff
@@ -114,3 +122,100 @@ func DaemonsetRolled(oldRouters Routers, newRouters Routers) error {
 		return fmt.Errorf("unknown router type: %T", oldRouters)
 	}
 }
+
+// RestartDaemonSetPods restarts all pods that belong to the DaemonSet identified by namespace + label selector and
+// waits until the DaemonSet reports ready.
+func RestartDaemonSetPods(cs clientset.Interface, namespace, labelSelector string) {
```
fedepaol, line 128: so this is to reset the crashloopbackoff time?

e2etests/pkg/k8s/pods.go
```diff
@@ -213,6 +213,27 @@ func PodIsReady(p *corev1.Pod) bool {
...
 
+// WaitForPodReadyConsistently waits until the provided pod is consistently ready.
+// We make sure that the pod is consistently ready 10 consecutive times (roughly 10'ish seconds) to avoid that we catch
+// an invalid ready state early in case the pod is transitioning from ready to not ready.
+// Note: kubernetes' MaxCrashLoopBackOff is hardcoded to 5 minutes:
+// https://github.com/kubernetes/kubernetes/blob/ec9448ceef13adbbcd3fa9f09b4d3c1fa54cc674/pkg/kubelet/kubelet.go#L166
+// https://github.com/kubernetes/kubernetes/issues/57291
+// Therefore, before calling this function, make sure that the crashloop timer is reset, e.g. by a new rollout of the
+// pod.
+func WaitForPodReadyConsistently(cs clientset.Interface, podNamespace, podName string) {
```
fedepaol, line 224: this is going to add 10 seconds delay where is invoked.
Does it make sense to check the version of the pod we are checking the readiness of to exclude the ready too soon failure?

andreaskaris: > sense to check the version of the pod 

it's the same pod. let me check if I can work with the container status

e2etests/pkg/openperouter/openperouter.go
andreaskaris: yeah. I think there's no other way. maybe it's not needed ... running the test as often as I did locally is not something that happens in CI

PR #669, opened 2026-08-06
fix(e2e): give each leafkind its own ISIS net

qinqon:

/kind bug

`leafkind.tmpl` hardcoded `net 49.0001.0000.0000.0004.00`, and both
`LeafKind1Config` and `LeafKind2Config` render through it. Reconfiguring leafkind2
from a spec therefore hands it leafkind1's ISIS system ID. Both leaves then keep
receiving "their own" LSP back with a different checksum and answer by bumping the
sequence number and reflooding, so ISIS never converges and the routes learned over
it churn.

The clab boot config assigns the nets correctly, so the fabric only breaks once a
spec reconfigures a leaf.

Measured locally by flipping leafkind2's net and sampling leafkind1's own LSP
sequence number: 0/s with distinct nets, 3825/s with the duplicate. The CI leafkind
dumps carry over 400k `confused checksum` lines per run, on `main` as well.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/pkg/infra/leaf.go
+++ b/e2etests/pkg/infra/leaf.go
@@ -58,12 +58,14 @@ var (
 		SpinePeerAddress:  "192.168.1.4",
 		Container:         KindLeaf1Container,
 		ToSwitchInterface: "toswitch1",
+		ISISNet:           "49.0001.0000.0000.0004.00",
 	}
 	LeafKind2Config = LeafKind{
 		ASN:               64513,
 		SpinePeerAddress:  "192.168.1.6",
 		Container:         KindLeaf2Container,
 		ToSwitchInterface: "toswitch2",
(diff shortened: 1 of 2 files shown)


e2etests/pkg/infra/leaf.go
```diff
@@ -140,6 +143,10 @@ type LeafKind struct {
 	ASN               int
 	SpinePeerAddress  string
 	ToSwitchInterface string
+	// ISISNet must match the net the clab topology assigns to this leaf. Every
```
andreaskaris: I think we don't need this comment, we're also not adding similar explanations elsewhere

andreaskaris:
otherwise /lgtm

maiqueb (approved):
Tks

PR #670, opened 2026-08-06
test(e2e): fixed trafficTestResult interval for (systemd_)resiliency

andreaskaris:

/kind flake

measureTrafficLoss did not use a fixed interval for when it ran its
tests. Meaning that if a curl hung, each run would happen at ca.
2 seconds + 300 ms. Use a 500ms ticker instead and cap the curl at
400ms. Set maxAllowedFailures to 40 to allow for a maximum ~20 seconds
of downtime.
Also make sure that traffic actually works again before stopping and
counting.

Related to: https://github.com/openperouter/openperouter/issues/664
This PR adds a few improvements and increases the timeout by a few seconds.

However, the underlying RC for the issue seem to be repeated reconciles which in turn delay BGP session establishment and I'm not sure how to address this.

**Release note**:

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resiliency test validation by confirming traffic has recovered before measuring disruption.
  * Updated traffic monitoring to provide more consistent recovery results.
  * Added clearer failure reporting, including failure timestamps and recovery details.
  * Tests now detect missing traffic measurements and allow a defined level of disruption before failing.

### DIFF
--- a/e2etests/tests/resiliency.go
+++ b/e2etests/tests/resiliency.go
@@ -4,6 +4,7 @@ package tests
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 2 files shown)


andreaskaris:
@coderabbit review

e2etests/tests/resiliency.go
```diff
@@ -688,19 +694,23 @@ func measureTrafficLoss(exec executor.Executor, urlStr string) func() trafficTes
 	}
 }
 
+// eval evaluates the trafficTestResult. We are conservative and allow for up to
```
fedepaol: doesn't this mean we are more forgiving now?

andreaskaris: I rephrased that comment

e2etests/tests/resiliency.go
```diff
@@ -688,19 +694,23 @@ func measureTrafficLoss(exec executor.Executor, urlStr string) func() trafficTes
 	}
 }
 
+// eval evaluates the trafficTestResult. In order to reduce flakes, we allow for up to 40 failures,
+// Meaning that with a tick interval of 500ms, we allow up to ca. 20 seconds of downtime.
+// We need to be a bit more lenient here as depending on the reconcile timing, BGP sessions might take a while to
```
fedepaol: @maiqueb is this too much for your migration target?
If so, we need to understand if either we have a bug or we need to tighten up the tests

maiqueb: 20 seconds of downtime is not acceptable for virt. 

I think we need to better understand the error pattern before allowing the tests to be more lenient.

andreaskaris: the test did allow for 2.3 * 5 = ca. 12 seconds, so I'll tweak it to that value

Issue #400, opened 2026-05-11
E2E flake: BridgeRefresher E2E - Type 2 Route Persistence Pod migrates to a different node should maintain connectivity after pod migrates to another node [It] ipv6


andreaskaris:
https://github.com/openperouter/openperouter/actions/runs/25566667500/job/75052941017?pr=394
https://productionresultssa13.blob.core.windows.net/actions-results/589c8300-d74c-4118-a365-85c47b1a52c7/workflow-job-run-d7db6632-99c0-5e00-bfdd-83d07ed752bf/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-05-11T08%3A38%3A38Z&sig=mUJHhl%2BjrOoj3690rqRjdjtl841vDzLGqXcWMuzwFJM%3D&ske=2026-05-11T11%3A33%3A24Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-05-11T07%3A33%3A24Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-05-11T08%3A28%3A33Z&sv=2025-11-05

```
2026-05-08T16:53:01.0114360Z   [38;5;9m[FAILED][0m in [It] - /home/runner/work/openperouter/openperouter/e2etests/tests/bridgerefresh.go:320 [38;5;243m@ 05/08/26 16:53:01.011[0m
2026-05-08T16:53:01.0351710Z   [1mSTEP:[0m waiting for the router pod to rollout after removing the underlay [38;5;243m@ 05/08/26 16:53:01.034[0m
2026-05-08T16:53:19.1761930Z   router pods are:Pod openperouter-system/router-5k5zp: Running  Node: pe-kind-worker  IPs: [{10.244.1.32} {fd00:10:244:1::20}]  Containers:    - frr: quay.io/openperouter/router:main    - reloader: quay.io/openperouter/router:main
2026-05-08T16:53:19.1764907Z   Pod openperouter-system/router-rxnbz: Running  Node: pe-kind-control-plane  IPs: [{10.244.0.22} {fd00:10:244::16}]  Containers:    - frr: quay.io/openperouter/router:main    - reloader: quay.io/openperouter/router:main
[...]

PR #676, opened 2026-08-08
Bridge refresher E2E - fix potential flake, refactor and collect additional data

andreaskaris:
- **fix(e2e): fix potential flake in bridgerefresher due to neigh tbl prune**
- **refactor(e2e): refactor bridgerefresher e2e tests**
- **fix(e2e): collect ip monitor neigh for bridge refresher e2e tests**
- **fix(e2e): fix triggering bridge refresher STALE condition**

/kind flake

Address a potential cause for https://github.com/openperouter/openperouter/issues/400
The bridgerefresher E2E test requires that the neighbor entries for the migrated pod go to STALE or FAILED.
However, if the kernel garbage collects the STALE or FAILED entries, the test will fail.
Whereas I'm not sure if the flakes that we see are indeed caused by garbage collection, I can say with certainty
that the exact same symptoms can indeed be triggered by garbage collection.
Therefore, increase the gh_thresh values to their maximum to make garbage collection near impossible.
Refactor the tests slightly to gather some extra data.
And collect ip monitor neigh output in case flakes happen again in the future.

While testing, I found that the current test does not reliably reproduce the issue. 
a) The original test case checked for STALE | FAILED. However, a FAILED entry will be purged from the neighbor table and does not reproduce the issue. 
b) The entry must be in STALE _before_ the migration. Otherwise, if the pod's neighbor entry goes into FAILED after the pod was deleted, the issue cannot be reproduced.

**Release note**:

```release-note
None
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved migration validation by monitoring neighbor events across router nodes.
  * Added clearer diagnostics for migration and connectivity failures.
  * Prevented failed neighbor states from being treated as successful.

* **Testing**
  * Added configurable timeouts for pod connectivity checks.
  * Added automatic neighbor-cache configuration and restoration during migration tests.
  * Added cancellable command execution for more reliable end-to-end testing.
  * Improved handling of interrupted test commands and connectivity checks.

### DIFF
--- a/e2etests/pkg/executor/executor.go
+++ b/e2etests/pkg/executor/executor.go
@@ -3,6 +3,7 @@
 package executor
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
@@ -15,6 +16,11 @@ var Kubectl string
 
 type Executor interface {
 	Exec(cmd string, args ...string) (string, error)
+}
+
+type ExecutorWithContext interface {
+	Executor
+	CommandContext(ctx context.Context, cmd string, args ...string) (string, error)
 }
 
 type hostExecutor struct{}
@@ -35,7 +41,7 @@ func (hostExecutor) Exec(cmd string, args ...string) (string, error) {
 	return string(out), err
 }
 
-func ForContainer(containerName string) Executor {
+func ForContainer(containerName string) ExecutorWithContext {
 	return &containerExecutor{container: containerName}
 }
 
@@ -46,6 +52,12 @@ type containerExecutor struct {
 func (e *containerExecutor) Exec(cmd string, args ...string) (string, error) {
 	newArgs := append([]string{"exec", e.container, cmd}, args...)
 	out, err := exec.Command(ContainerRuntime, newArgs...).CombinedOutput()
+	return string(out), err
+}
+
+func (e *containerExecutor) CommandContext(ctx context.Context, cmd string, args ...string) (string, error) {
+	newArgs := append([]string{"exec", e.container, cmd}, args...)
+	out, err := exec.CommandContext(ctx, ContainerRuntime, newArgs...).CombinedOutput()
 	return string(out), err
 }
 
--- a/e2etests/pkg/sysctl/sysctl.go
+++ b/e2etests/pkg/sysctl/sysctl.go
@@ -1,0 +1,146 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package sysctl
+
+import (
+	"errors"
+	"fmt"
+	"log/slog"
+	"os"
+	"os/exec"
+	"strings"
+)
+
+const (
+	NeighDefaultGcThreshMax = "2147483647"
+)
+
+// Sysctl represents a sysctl setting to be enabled.
+type Sysctl struct {
+	Path        string // The sysctl path under /proc/sys/
+	Description string // Human-readable description for logging
+	Value       string // Value to configure or to read
+}
+
+// Ensure enables the given sysctls.
+// Each sysctl is checked and only written if not already set to the target value.
+func Ensure(sysctls ...Sysctl) error {
+	var errs []error
+	for _, s := range sysctls {
+		if err := ensureSysctl(s); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
+}
+
+// Read reads the given sysctls and stores the value in Value.
+func Read(sysctls ...*Sysctl) error {
+	for _, s := range sysctls {
+		if err := readSysctl(s); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func IPv4NeighDefaultGcThresh1(targetValue string) Sysctl {
+	return Sysctl{
+		Path: "net/ipv4/neigh/default/gc_thresh1",
+		Description: "Minimum number of entries to keep. " +
+			"Garbage collector will not purge entries if there are fewer than this number.",
+		Value: targetValue,
+	}
+}
+
+func IPv4NeighDefaultGcThresh2(targetValue string) Sysctl {
+	return Sysctl{
+		Path: "net/ipv4/neigh/default/gc_thresh2",
+		Description: "Threshold when garbage collector becomes more aggressive about purging entries. " +
+			"Entries older than 5 seconds will be cleared when over this number.",
+		Value: targetValue,
+	}
+}
+
+func IPv4NeighDefaultGcThresh3(targetValue string) Sysctl {
+	return Sysctl{
+		Path: "net/ipv4/neigh/default/gc_thresh3",
+		Description: "Maximum number of non-PERMANENT neighbor entries allowed. " +
+			"Increase this when using large numbers of interfaces and when communicating " +
+			"with large numbers of directly-connected peers.",
+		Value: targetValue,
+	}
+}
+
+func IPv6NeighDefaultGcThresh1(targetValue string) Sysctl {
+	return Sysctl{
+		Path: "net/ipv6/neigh/default/gc_thresh1",
+		Description: "Minimum number of entries to keep. " +
+			"Garbage collector will not purge entries if there are fewer than this number.",
+		Value: targetValue,
+	}
+}
+
+func IPv6NeighDefaultGcThresh2(targetValue string) Sysctl {
+	return Sysctl{
+		Path: "net/ipv6/neigh/default/gc_thresh2",
+		Description: "Threshold when garbage collector becomes more aggressive about purging entries. " +
+			"Entries older than 5 seconds will be cleared when over this number.",
+		Value: targetValue,
+	}
+}
+
+func IPv6NeighDefaultGcThresh3(targetValue string) Sysctl {
+	return Sysctl{
+		Path: "net/ipv6/neigh/default/gc_thresh3",
+		Description: "Maximum number of non-PERMANENT neighbor entries allowed. " +
+			"Increase this when using large numbers of interfaces and when communicating " +
+			"with large numbers of directly-connected peers.",
+		Value: targetValue,
+	}
+}
+
+// ensureSysctl reads the sysctl at the given path and writes the desired value with sudo if it differs
+// from the current one.
+func ensureSysctl(s Sysctl) error {
+	path := "/proc/sys/" + s.Path
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %w", path, err)
+	}
+	currentValue := strings.TrimSpace(string(data))
+	if currentValue == s.Value {
+		return nil
+	}
+
+	cmd := []string{"/bin/bash", "-c", fmt.Sprintf("echo %s > %s", s.Value, path)}
+	if out, err := exec.Command("sudo", cmd...).CombinedOutput(); err != nil {
+		return fmt.Errorf("failed to write to %s: %w (%s)", path, err, out)
+	}
+
+	data, err = os.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %w", path, err)
+	}
+	currentValue = strings.TrimSpace(string(data))
+	if currentValue != s.Value {
+		return fmt.Errorf("failed to set %s to target value %s, current: %s", path, s.Value, currentValue)
+	}
+
+	slog.Info("sysctl enabled", "path", s.Path, "value", s.Value)
+	return nil
+}
+
+// readSysctl reads the sysctl at the given path and writes it to s.Value.
+func readSysctl(s *Sysctl) error {
+	path := "/proc/sys/" + s.Path
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %w", path, err)
+	}
+	currentValue := strings.TrimSpace(string(data))
+	s.Value = currentValue
+
+	slog.Info("sysctl read", "path", s.Path, "value", s.Value)
+	return nil
+}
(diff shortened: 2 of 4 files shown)


e2etests/tests/bridgerefresh.go
```diff
@@ -70,17 +74,39 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
...
+
+		By("Setting neighbor GC thresholds to max")
+		sysctl.Ensure(
+			sysctl.IPv4NeighDefaultGcThresh1(sysctl.NeighDefaultGcThreshMax),
+			sysctl.IPv4NeighDefaultGcThresh2(sysctl.NeighDefaultGcThreshMax),
+			sysctl.IPv4NeighDefaultGcThresh3(sysctl.NeighDefaultGcThreshMax),
+			sysctl.IPv6NeighDefaultGcThresh1(sysctl.NeighDefaultGcThreshMax),
+			sysctl.IPv6NeighDefaultGcThresh2(sysctl.NeighDefaultGcThreshMax),
+			sysctl.IPv6NeighDefaultGcThresh3(sysctl.NeighDefaultGcThreshMax),
+		)
```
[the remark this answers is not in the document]
andreaskaris: I pushed a new commit, please check

e2etests/tests/bridgerefresh.go
```diff
@@ -200,19 +231,31 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
...
 
+		AfterEach(func() {
+			By("Printing ip neigh monitor output after the test for NodeA")
+			output, err := cancelIPNeighMonitorNodeA()
+			fmt.Fprintf(GinkgoWriter, "ip neigh monitor output: %s; err: %q\n", output, err)
+
+			By("Printing ip neigh monitor output after the test for NodeB")
+			output, err = cancelIPNeighMonitorNodeB()
+			fmt.Fprintf(GinkgoWriter, "ip neigh monitor output: %s; err: %q\n", output, err)
+		})
```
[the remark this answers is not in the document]
andreaskaris: I pushed a new commit, please check

andreaskaris:
@coderabbit review

andreaskaris:
@corerabbit review

e2etests/pkg/sysctl/sysctl.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Read reads the given sysctls and stores the value in Value.
+func Read(sysctls ...*Sysctl) error {
+	for _, s := range sysctls {
+		if err := readSysctl(s); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
andreaskaris: the values are separate. Read reads ipv4GcThresh1 which is later used to reset to the originally read value. The value is set to Max with 			Expect(sysctl.Ensure(
				sysctl.IPv4NeighDefaultGcThresh1(sysctl.NeighDefaultGcThreshMax),
so there's no "assign the same memory" here

e2etests/tests/validate_pods.go
```diff
@@ -39,14 +39,18 @@ func checkPodIsReachableWithExpected(exec executor.Executor, from, to, expected
...
+	if len(timeout) > 0 {
+		withTimeout = timeout[0]
+	}
 	Eventually(func(g Gomega) {
 		By(fmt.Sprintf("pinging %s via net1", ip))
 		out, err := exec.Exec("ping", "-c", "1", "-W", "2", "-I", "net1", ip)
 		g.Expect(err).ToNot(HaveOccurred(), "ping to %s failed: %s", ip, out)
 	}).
-		WithTimeout(40 * time.Second).
+		WithTimeout(time.Duration(withTimeout) * time.Second).
```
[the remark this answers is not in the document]
andreaskaris: I don't think that's necessary (it already worked like this just fine)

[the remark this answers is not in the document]
andreaskaris: I mean, technically, you're correct, but the ping runs for max 2 seconds

e2etests/tests/bridgerefresh.go
```diff
@@ -450,7 +468,7 @@ func checkNeighborStale(cs clientset.Interface, podIP string, vni int, nodeName
 	if out == "" {
 		return fmt.Errorf("no neighbor entry for %s on %s in router %s", podIP, bridgeDev, exec.Name())
 	}
-	if strings.Contains(out, "STALE") || strings.Contains(out, "FAILED") {
+	if strings.Contains(out, "STALE") {
```
andreaskaris, line 474: Explanation: waiting here for FAILED makes little sense - a FAILED entry will be removed from the ARP table (https://blogs.oracle.com/linux/arp-internals) and thus does not reproduce the issue

e2etests/tests/bridgerefresh.go
```diff
@@ -395,7 +413,7 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
 
 				By(fmt.Sprintf("Verifying migrating pod (%s) can ping the local gateway", migratingPod.Name))
 				newMigratingExec := executor.ForPod(testNamespace, migratingPod.Name, "agnhost")
-				canPingFromPod(newMigratingExec, l2GatewayIPOnly)
+				canPingFromPodWithTimeout(newMigratingExec, l2GatewayIPOnly, 60*time.Second)
```
andreaskaris, line 419: now that we reproduce the issue more reliably, increase this here to 60 seconds. The bridge refresher runs every 30 seconds and having the ping check time out after 40 is a bit too close for my taste for the refresher to ping and then for the ARP table to clean up and the l2vpn routes to be properly exchanged. 60 seconds sounds more reasonable

andreaskaris:
@coderabbit review

e2etests/pkg/sysctl/sysctl.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %w", path, err)
+	}
+	currentValue := strings.TrimSpace(string(data))
+	if currentValue == s.Value {
+		return nil
+	}
+
+	cmd := []string{"/bin/bash", "-c", fmt.Sprintf("echo %s > %s", s.Value, path)}
+	if out, err := exec.Command("sudo", cmd...).CombinedOutput(); err != nil {
```
andreaskaris, line 117: This must be set on the host system .. So we unfortunately need sudo afaict. I'm open for other suggestions but I think it's needed :-( If so, should the test be skipped if this fails?

e2etests/tests/validate_pods.go
```diff
@@ -40,13 +40,17 @@ func checkPodIsReachableWithExpected(exec executor.Executor, from, to, expected
 }
 
 func canPingFromPod(exec executor.Executor, ip string) {
+	canPingFromPodWithTimeout(exec, ip, 40)
```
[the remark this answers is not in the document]
andreaskaris: done

e2etests/tests/bridgerefresh.go
maiqueb: Is that behavior controller by a sysctl or something ? 

Is there another state we might want to track ?

e2etests/pkg/sysctl/sysctl.go
maiqueb: when we run using `make test` these are executed in a privileged container. We should be OK to run these tests there.

e2etests/tests/bridgerefresh.go
```diff
@@ -292,7 +297,7 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
 				)
 				Expect(err).NotTo(HaveOccurred())
 
-				By("Creating migrating pod on node A")
+				By(fmt.Sprintf("Creating migrating pod on node A (%s)", nodes[0].Name))
```
maiqueb: maybe we can drop the A / B naming, and be more descriptive - how about src and dst cluster ? or migration src + migration dst  ?

e2etests/tests/bridgerefresh.go
```diff
@@ -344,12 +351,15 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
 				}, 2*time.Minute, time.Second).Should(BeTrue())
 
 				// This is important to trigger the deadlock situation described in https://github.com/FRRouting/frr/issues/14156
-				By("Waiting for neighbor entry to go STALE on the router on the same node")
+				By(fmt.Sprintf("Waiting for neighbor entry (%s) on bridge br-pe-%d to go STALE or FAILED on the "+
```
maiqueb: do we actually wait for it to go stale ? I thought this PR said that was not possible.

e2etests/pkg/executor/executor.go
```diff
@@ -49,6 +55,12 @@ func (e *containerExecutor) Exec(cmd string, args ...string) (string, error) {
 	return string(out), err
 }
 
+func (e *containerExecutor) CommandContext(ctx context.Context, cmd string, args ...string) (string, error) {
+	newArgs := append([]string{"exec", e.container, cmd}, args...)
+	out, err := exec.CommandContext(ctx, ContainerRuntime, newArgs...).CombinedOutput()
+	return string(out), err
+}
```
maiqueb, line 62: maybe we can simple just replace this function with the previous one. 

Is there any interest in having both these funcs ? 

This is opinionated - we can proceed without it, and fix in a follow-up.

maiqueb (changes requested):
looks good, but I do have some questions.

e2etests/pkg/sysctl/sysctl.go
andreaskaris: Ah true - I had not considered that and was really worried because I couldn't find any other sysctl / sudo tests

e2etests/tests/bridgerefresh.go
andreaskaris: sounds good I'll address that inside the refactor commit

e2etests/tests/bridgerefresh.go
andreaskaris: ha, here I was on the other PR criticizing confusion when a later commit in the same series fixes a change introduced by newer one ...

checkNeighborStale internally checked for STALE or FAILED, that's why I had changed the message to reflect what was actually happening. That was before the last commit that moves things around.

In the last commit in the series, I try to make the scenario way more likely to happen and in the (too long) comment I explain why. The location that the wait command was at originally, it was very unlikely to trigger a STALE neighbor entry state and thus to actually observe the bridge refresher doing its work. Instead, because we deleted the pod first, then a packet might trigger FAILED state, then we waited for the state, Fede had to add a wait for STALE (issue reproduced) or FAILED (issue not reproduced) to unblock the test and avoid that it failed

With the last commit in this series, I change the location of the wait. I move it _before_ the migration, where it can wait for all traffic from the br-... gw to the pod to die down. If some stray traffic still occurs, the ARP entry might go back from STALE into REACHABLE to STALE, but with the pod there, no matter what, once no more traffic occurs between our perouter br-... gateway and the pod, the entry is guaranteed to go into STALE. (within ca. 30 seconds .. I tried finding the timer to shorten the interval but the ones that I tinkered with didn't work).
Thus, in the new location, the `Waiting for` does not need to wait for FAILED any more.

Once we pass the `Waiting for ..` part: Any subsequent state transition (before or after migration, depending on timing) from STALE to REACHABLE (before migration) or STALE to FAILED (after migration) is likely triggered by the bridge refresher's ping - and that's exactly what we want to see.

There might still be a small timing issue where the bridge refresher (runs every 30 seconds) could get an entry out of STALE into REACHABLE, then the entry goes back to STALE and then the ping refresher detects that before the `Waiting for` part ... but given that we check every 2 seconds vs the refreshers 30, we should be ok? I had also thought about Skip()'ping the test here in case of a problem, but I'd like to see if it still flakes in the future, and if so, maybe improve the test setup, instead of skipping and losing the signal (as we skip quite a few tests and so we wouldn't easily see if we skipped this one or not)

e2etests/pkg/executor/executor.go
andreaskaris: The problem is the signature. I'd have to change the interface to include the ctx and thus the code for every caller, right?
I'm not against that, I was just trying to make a smaller change.
wdyt?

andreaskaris:
@coderabbit review

e2etests/tests/bridgerefresh.go
andreaskaris: I think I answered that in the long answer that I gave earlier, hope it's clear now :)

e2etests/pkg/executor/executor.go
maiqueb: ok then. Let's keep it.

e2etests/tests/bridgerefresh.go
```diff
@@ -341,7 +341,11 @@ var _ = Describe("BridgeRefresher E2E - Type 2 Route Persistence", Ordered, func
 				By(fmt.Sprintf("Verifying stationary pod (%s) can reach migrating pod (%s) before migration",
 					stationaryPod.Name, migratingPod.Name))
 				stationaryExec := executor.ForPod(testNamespace, stationaryPod.Name, "agnhost")
-				checkPodIsReachable(stationaryExec, stationaryPodIPOnly, migratingPodIPOnly)
+				// Use a ping here instead of checkPodIsReachable which uses curl. The reason: TCP sessions exchange
+				// a bunch of packets during the FIN/ACK dance and we want to get reliably into STALE, so we want to
+				// test connectivity yet be sure that no more packets are exchanged between the hosts once connectivity
+				// is confirmed (stray packets might otherwise bring us from STALE to REACHABLE or from STALE to FAILED).
```
maiqueb, line 347: ah tks for this explanation.

maiqueb (approved):
thank you.

PR #693, opened 2026-08-21
ipam: remove +1 offset from RouterID derivation

### LINKED ISSUE #692: RouterID is chosen inconsistent with e.g. loopback IP
RouterID is chosen as follows, with a + 1 offset:

```
// RouterID returns the IP to be used for the router ID on the ith node.
func RouterID(pool string, index int) (string, error) {
	_, cidr, err := net.ParseCIDR(pool)
	if err != nil {
		return "", fmt.Errorf("failed to parse pool %s: %w", pool, err)
	}

	ip, err := gocidr.Host(cidr, index+1)
	if err != nil {
		return "", fmt.Errorf("failed to get router id for node %d from cidr %s: %w", index, cidr, err)
	}

	return ip.String(), nil
}
```

But e.g. the loopback addresses are picked with ParseCIDR and then offset by the nodeIndex:

```
// RouterID returns the IP to be used for the router ID on the ith node.
func RouterID(pool string, index int) (string, error) {
	_, cidr, err := net.ParseCIDR(pool)
	if err != nil {
		return "", fmt.Errorf("failed to parse pool %s: %w", pool, err)
	}

	ip, err := gocidr.Host(cidr, index+1)
	if err != nil {
		return "", fmt.Errorf("failed to get router id for node %d from cidr %s: %w", index, cidr, err)
	}

	return ip.String(), nil
}
```

The node index for non-static / systemd deployments starts with 0.

internal/controller/nodeindex/nodes_allocation.go

```
const OpenpeNodeIndex = "openpe.io/nodeindex"

func nodesToAnnotate(allNodes []v1.Node) []v1.Node {
	taken := make([]int, 0)

	res := make([]v1.Node, 0)
	for _, n := range allNodes {
		if n.Annotations[OpenpeNodeIndex] != "" {
			index, err := strconv.Atoi(n.Annotations[OpenpeNodeIndex])
			if err != nil { // not an int, let's ignore it so we'll add again
				res = append(res, n)
				continue
			}
			if alreadyExistsInSorted(taken, index) {
				res = append(res, n)
				continue
			}
			taken = insertSorted(taken, index)
			continue
		}
		res = append(res, n)
	}

	allocator := newAllocator(taken)
	for i := range res {
		if res[i].Annotations == nil {
			res[i].Annotations = make(map[string]string)
		}
		index := allocator.getNextFree()
		res[i].Annotations[OpenpeNodeIndex] = strconv.Itoa(index)
	}
	return res
}

type allocator struct {
	lastAssigned int
	upperIndex   int
	taken        []int
}

func newAllocator(taken []int) *allocator {
	return &allocator{
		lastAssigned: -1,
		upperIndex:   0,
		taken:        taken,
	}
}

func (a *allocator) getNextFree() int {
	nextFree := a.findNextFree()
	a.lastAssigned = nextFree
	return nextFree
}
```

for systemd deployments, users would prob pick auto-derived indexes from the interfaceName:

```
[root@master-0 /]# cat /etc/openperouter/node-config.yaml
nodeIndex:
  interfaceName: br0
logLevel: debug
```

Meaning that the nodeIndex is picked by the host part from the CIDR - meaning that in normal IPv4 subnetting, for the most common cases, the node ID will more than likely not start with 0:

```
func NodeIndexFromInterface(name, cidr string) (int, error) {
	iface, err := net.InterfaceByName(name)
	if err != nil {
		return 0, fmt.Errorf("failed to find interface %s: %w", name, err)
	}

	addrs, err := iface.Addrs()
	if err != nil {
		return 0, fmt.Errorf("failed to get addresses for interface %s: %w", name, err)
	}

	if cidr != "" {
		_, cidrNet, err := net.ParseCIDR(cidr)
		if err != nil {
			return 0, fmt.Errorf("invalid cidr %q: %w", cidr, err)
		}
		return findIndexFromCIDR(addrs, cidrNet, name, cidr)
	}
	return findIndexFromAddresses(addrs, name)
}

func findIndexFromCIDR(addrs []net.Addr, cidrNet *net.IPNet, ifaceName, cidr string) (int, error) {
	for _, addr := range addrs {
		ipNet, ok := addr.(*net.IPNet)
		if !ok {
			continue
		}
		if cidrNet.Contains(ipNet.IP) {
			return hostPartFromIPNet(ipNet), nil
		}
	}
	return 0, fmt.Errorf("no address matching cidr %s found on interface %s", cidr, ifaceName)
}
```

The loopback addresses are /32 or /128. Meaning they can very well end in a 0, without any issues.
The routerID is not even an IP address - it just happens to look the same.

Therefore, in order to be consistent, the best is to not offset the routerID.

andreaskaris:
/kind bug

RouterID was computed with a `+1` offset from the node index, but all
other address derivations (loopback IPs, veth IPs) use the raw index
directly. This is an aesthetic inconsistency but one that is confusing.

This PR removes the `+1` offset from `ipam.RouterID()` so that the
router ID is derived consistently with all other per-node addresses.

Fixes #692

The only production change is a single line in `internal/ipam/ipam.go`
(`index+1` → `index`). All other changes are test expectation updates
in `internal/conversion/frr_conversion_test.go`.

RouterIDs are *not* IP addresses (they just follow IPv4 syntax) so offsetting them by +1 made little sense.

The problem is less one of functionality but of the mental model of our FRR configuration. E.g. we currently get stuff like:

```
router bgp 65500
 bgp router-id 10.0.0.3    # <--- inconsistent
 no bgp ebgp-requires-policy
 no bgp enforce-first-as
 no bgp hard-administrative-reset
 no bgp default ipv4-unicast
 bgp cluster-id 10.255.255.255
 no bgp graceful-restart notification
 no bgp network import-check
 neighbor fc00:0:20::1 remote-as internal
 neighbor fc00:0:20::1 update-source fd00::2  # <--- inconsistent
 neighbor fc00:0:20::1 capability extended-nexthop
```

And this fix changes this to:
```
router bgp 65500
 bgp router-id 10.0.0.2
 no bgp ebgp-requires-policy
 no bgp enforce-first-as
 no bgp hard-administrative-reset
 no bgp default ipv4-unicast
 bgp cluster-id 10.255.255.255
 no bgp graceful-restart notification
 no bgp network import-check
 neighbor fc00:0:20::1 remote-as internal
 neighbor fc00:0:20::1 update-source fd00::2
 neighbor fc00:0:20::1 capability extended-nexthop
```

**Release note**:

```release-note
Fix RouterID derivation to be consistent with loopback and other
per-node address assignments by removing the +1 index offset.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected router ID assignment to use the configured node index consistently.
  * Updated generated route distinguishers to reflect the corrected router ID.
  * Improved consistency across underlay, VNI, VPN, SRv6, ISIS, passthrough, and Layer 2 gateway routing scenarios.
  * Prevented mismatches between router identifiers and associated network routes.

### DIFF
--- a/internal/conversion/frr_conversion_test.go
+++ b/internal/conversion/frr_conversion_test.go
@@ -58,7 +58,7 @@ func TestAPItoFRR(t *testing.T) {
 					TunnelEndpoint: &frr.TunnelEndpoint{
 						IPv4CIDR: "192.168.1.0/32",
 					},
-					RouterID: "10.0.0.1",
+					RouterID: "10.0.0.0",
 					Neighbors: []frr.NeighborConfig{
 						{
 							Name:                  "65001@192.168.1.1",
@@ -101,7 +101,7 @@ func TestAPItoFRR(t *testing.T) {
 					TunnelEndpoint: &frr.TunnelEndpoint{
 						IPv4CIDR: "192.168.1.0/32",
 					},
-					RouterID: "10.0.0.1",
+					RouterID: "10.0.0.0",
 					Neighbors: []frr.NeighborConfig{
 						{
 							Name:                  "external@192.168.1.1",
@@ -149,7 +149,7 @@ func TestAPItoFRR(t *testing.T) {
 					TunnelEndpoint: &frr.TunnelEndpoint{
 						IPv4CIDR: "192.168.1.0/32",
 					},
-					RouterID: "10.0.0.1",
+					RouterID: "10.0.0.0",
 					Neighbors: []frr.NeighborConfig{
 						{
 							Name:                  "internal@192.168.1.1",
@@ -197,7 +197,7 @@ func TestAPItoFRR(t *testing.T) {
 					TunnelEndpoint: &frr.TunnelEndpoint{
 						IPv4CIDR: "192.168.1.0/32",
 					},
-					RouterID: "10.0.0.1",
+					RouterID: "10.0.0.0",
 					Neighbors: []frr.NeighborConfig{
 						{
 							Name:                  "65000@192.168.1.1",
@@ -255,7 +255,7 @@ func TestAPItoFRR(t *testing.T) {
 					TunnelEndpoint: &frr.TunnelEndpoint{
 						IPv4CIDR: "192.168.1.0/32",
 					},
-					RouterID: "10.0.0.1",
+					RouterID: "10.0.0.0",
 					Neighbors: []frr.NeighborConfig{
 						{
 							Name: "65001@192.168.1.1",
@@ -275,7 +275,7 @@ func TestAPItoFRR(t *testing.T) {
 						ASN:      65000,
 						VNI:      200,
 						VRF:      "vrf1",
-						RouterID: "10.0.0.1",
+						RouterID: "10.0.0.0",
 						LocalNeighbor: &frr.NeighborConfig{
 							Addr: "192.168.2.2",
 							ID:   "192.168.2.2",
@@ -334,7 +334,7 @@ func TestAPItoFRR(t *testing.T) {
 					TunnelEndpoint: &frr.TunnelEndpoint{
 						IPv4CIDR: "192.168.1.0/32",
 					},
-					RouterID: "10.0.0.1",
+					RouterID: "10.0.0.0",
 					Neighbors: []frr.NeighborConfig{
 						{
 							Name: "65001@2001:db8::1:192:168:1:1",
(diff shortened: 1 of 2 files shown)


internal/conversion/frr_conversion_test.go
```diff
@@ -2646,7 +2646,7 @@ func TestAPItoFRR(t *testing.T) {
 							{Name: "lo", IPv6: true, IsPassive: true},
 						},
 					},
-					RouterID: "10.0.0.1",
+					RouterID: "10.0.0.0",
```
andreaskaris, line 2649: we can see in this example that the routerID and updateSource (IPv6 interface IP) were inconsistent (updateSource ended with 0); it's the same for the tunnelSource for IPv4 even though this doesn't show in the frr_conversion tests

andreaskaris:
@fedepaol ptal and wdyt

fedepaol:
to be fair, we can't have a .0 veth ip: https://github.com/openperouter/openperouter/blob/main/internal/ipam/ipam.go#L169

Said that, it's a matter of taste I guess.

PR #737, opened 2026-09-07
fix(api): validate port ranges and mark HostMaster union

qinqon:

/kind bug

- Restrict L2VNI and L3VNI `vxlanPort` to 1-65535.
- Correct `Neighbor.port` from 0-16384 to 1-65535.
- Add `+union` to `HostMaster` and `+unionDiscriminator` to its `Type` field, preserving existing CEL validation.
- Add 15 schema validation cases covering negative, zero, minimum, maximum, and out-of-range ports.
- Regenerate CRDs, Helm templates, operator bundle, all-in-one manifests, and API documentation.

N/S

**Release note**:

```release-note
L2VNI and L3VNI VXLAN ports and BGP neighbor ports are now validated against the range 1-65535. BGP neighbor ports above 16384 are now accepted. Action required: replace explicitly configured port 0 or out-of-range VXLAN ports with valid ports before updating affected resources.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for VXLAN ports, which must now be between 1 and 65,535.
  * Updated BGP neighbor port validation to allow values from 1 through 65,535; port 0 is no longer valid.

* **Documentation**
  * Updated API documentation to reflect the revised port validation ranges.

* **Tests**
  * Added coverage for valid boundary values and rejected out-of-range ports.


api/v1alpha1/neighbor.go
```diff
@@ -58,8 +58,8 @@ type Neighbor struct {
 	// port is the port to dial when establishing the session.
 	// Defaults to 179.
 	// +optional
-	// +kubebuilder:validation:Minimum=0
-	// +kubebuilder:validation:Maximum=16384
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
```
maiqueb, line 62: @fedepaol is there any reason why you've defined this maximum ? Was this a simple mistake ?

maiqueb (approved):
LGTM, but just want to double-check with the original author the motives (if any) for limiting the max port number.

api/v1alpha1/l3vni_types.go
```diff
@@ -43,6 +43,8 @@ type L3VNISpec struct {
 	VNI int32 `json:"vni,omitempty"`
 
 	// vxlanPort is the port to be used for VXLan encapsulation.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
```
ormergi, line 47: nit: Consider adding new type for a port and have it cary these validations: 
```
// +kubebuilder:validation:Maximum=65535
type Port *int32
```
Just a suggestion, it can also be done in different PR.

api/v1alpha1/neighbor.go
fedepaol: sounds like, yes

ormergi (approved):
Thanks for the fix, LGTM 

I left one suggestion, can be done in diff PR though.

PR #752, opened 2026-09-11
feat(systemd): host-network VNF proof of concept

qinqon:

/kind feature

Adds a PoC for running OpenPERouter as a dedicated VM/VNF with systemd Quadlets and static configuration directly in the host network namespace.

- Adds `--router-netns` to the host-mode controller, defaulting to `/var/run/netns/perouter`.
- Detects host networking by namespace identity, including bind-mounted aliases; no empty-path convention or separate host-network flag.
- Leaves host uplinks externally provisioned: no moves, link-group marking, restoration, or automatic veth MTU adjustment.
- Allows Underlay interfaces to be omitted for the host kernel datapath; separate-namespace conversion still requires them.
- Reconciles overlay removal independently of uplink bookkeeping, preserving host NICs and loopback addresses.
- Brings up both veth ends when the target is the same namespace.
- Includes ready-to-copy Quadlet drop-ins, a static VNF example, and installation/verification instructions in `systemdmode/host-network/README.md`.
- Regenerates CRDs, bundles, all-in-one manifests and API docs.

Draft PoC for experimentation on a dedicated router-owned VM. The controller must itself use host networking. The router pod uses `/proc/1/ns/net` as seen by host-side Podman, while the controller accesses it through `/hostproc/1/ns/net`.

PoC boundaries:
- Kernel datapath/systemd only; custom namespace lifecycle is installation-owned.
- MTUs and cleanup of removed/changed tunnel endpoint addresses are externally managed.
- VRF cleanup retains the existing router-owned-namespace assumption.
- Same-namespace host BGP sessions and full VM forwarding/reboot behavior still need validation.
- The shared CRD schema now admits omitted interfaces; the host conversion enforces the namespace-dependent requirement at reconciliation time.

Validation:
- `make test` passed, including privileged race-enabled networking tests.
- `make lint` passed (0 issues).
- `make manifests generate-all-in-one bundle` and `make api-docs` passed; bundle validation reported existing metadata warnings.
- Podman Quadlet generator dry-run verified the drop-ins select only the host namespace and clear named-namespace creation steps.
- Focused tests cover namespace identity through aliases/bind mounts, distinct namespaces, no-uplink static BGP/EVPN conversion, VNF overlay creation/removal preserving uplink state, and same-namespace veth setup.

Adding a privileged controller-package test exposed that the container test runner used the repository root rather than each package directory. The runner now uses package working directories so existing relative fixtures resolve.

**Release note**:

```release-note
Add an experimental systemd host-network VNF deployment with configurable router namespace and externally managed uplinks.
```

- [ ] I have reviewed all changes in this PR, including any AI-generated content, and I take full responsibility for its accuracy and correctness.


qinqon:
I will just keep my branch

PR #755, opened 2026-09-11
grout: suppress underlay veth link-local to fix IPv6 sessions

### LINKED ISSUE #720: Flake: Underlay explicit address family configuration / peers with the tor an...
Flaky test detected by nightly CI.

**Test path:**
```
Underlay explicit address family configuration peers with the tor and exchanges address families ipv4unicast and ipv6unicast
```

**Deployment(s):** operator-grout
**Failure state(s):** failed
**First seen:** 2026-08-30
**Run:** https://github.com/openperouter/openperouter/actions/runs/33288014996

---
flake-id: [grout] Underlay explicit address family configuration peers with the tor and exchanges address families ipv4unicast and ipv6unicast

maiqueb:

>
> /kind cleanup
> /kind feature
> /kind design

/kind flake

> /kind failing
> /kind documentation
> /kind regression
> /kind example

Grout adopts the backing veth's MAC (via remote=) for the underlay
port, so its control-plane kernel shadow `u_<iface>` derives the same
EUI-64 IPv6 link-local as the veth. 

When grout adds that link-local to the shadow it collides with the
veth's copy, DAD strips it, and the shadow ends up with no link-local.
FRR then cannot set the IPv6 nexthop, so any session carrying one
(ipv6unicast/ipv6vpn) never establishes.

This surfaced as the grout-only nightly flake in #720.

The veth's kernel link-local is unused (grout owns all forwarding), so
suppress it before creating the port: set `addr_gen_mode` to none and
flush any existing link-local. This removes the collision partner and
lets `u_<iface>` keep a valid link-local (shared only with grout's own
`tap_<iface>`, which is by design and harmless).

Fixes #720 

**Release note**:

```release-note
Suppress underlay veth link-local to fix IPv6 sessions, to prevent IPv6 DAD in the grout TAP, which prevented the BGP session from establishing.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate IPv6 link-local addresses on underlay interfaces, avoiding address conflicts during network setup.
  - Underlay configuration now suppresses automatically generated link-local addresses before creating associated ports.
  - Improved link-local address ownership handling so each underlay interface retains the correct address association.

- **Tests**
  - Enabled regression coverage verifying that link-local addresses are uniquely assigned across underlay interfaces and do not conflict with their backing connections.


maiqueb:
@zeeke please take a look - grout's your gig :)

e2etests/pkg/openperouter/interface.go
```diff
@@ -78,6 +78,32 @@ func InterfaceIPv4InNetns(nodeName, intf, ns string) (string, error) {
...
+	for line := range strings.SplitSeq(out, "\n") {
+		// e.g. "36: u_toswitch1    inet6 fe80::a8c1:abff:fe84:5cc8/128 scope link nodad"
+		fields := strings.Fields(line)
+		if len(fields) < 4 || fields[2] != "inet6" {
+			continue
+		}
+		iface := fields[1]
+		addr, _, _ := strings.Cut(fields[3], "/")
+		owners[addr] = append(owners[addr], iface)
+	}
```
qinqon, line 111: ip -o -6 addr show scope link includes addresses flagged tentative or dadfailed, but the parser discards those flags. The new test therefore passes when u_toswitch1 has exactly one unusable link-local address, as long as the backing veth does not share it. It can also succeed before DAD finishes.

Preserve address flags and require the shadow’s link-local to be neither tentative nor DAD-failed. Keep all addresses in the ownership check so duplicate detection remains effective.

qinqon (changes requested):

e2etests/pkg/openperouter/interface.go
maiqueb: @coderabbitai can you rephrase this ? I am having some trouble understanding what the suggestion is. 

Also, evaluate it for correctness.

qinqon: @maiqueb you should skip dadfailed and tentative

maiqueb: done.

qinqon (approved):
LGTM

zeeke (approved):

internal/hostnetwork/link_properties.go
```diff
@@ -108,6 +108,36 @@ func setAddrGenModeNone(l netlink.Link) error {
 	return netlink.LinkSetIP6AddrGenMode(l, 1)
 }
 
+// SuppressLinkLocal stops the kernel from managing an IPv6 link-local address on
+// the named interface: it sets addr_gen_mode to none so no new one is generated
+// and removes any that already exists. Called on grout underlay interfaces whose
```
fedepaol: I think the second part of the comment is not relevant to this function as we explain it already on teh calling site

internal/grout/underlay.go
```diff
@@ -166,6 +166,19 @@ func configureUnderlayPort(ctx context.Context, client *Client, underlayInterfac
 		return fmt.Errorf("failed to read underlay interface addresses: %w", err)
 	}
 
+	// grout adopts the backing veth's MAC (via remote=) for the port, so its
```
fedepaol: I'd reword this comment as what we are doing and why, as opposed to what is the issue and the fact that we are fixing it. Something like "suppress link local address to avoid conflicts to the shadow.."

No need to mention hte issue imo

e2etests/tests/sessions.go
```diff
@@ -1098,4 +1099,69 @@ var _ = Describe("Underlay explicit address family configuration", Ordered, Grou
...
+	// When grout reuses the backing veth's MAC for the underlay port, the port's
+	// kernel shadow u_<iface> derives the same EUI-64 link-local as the veth. The
+	// collision triggers DAD, which can strip u_<iface>'s link-local and break any
+	// session carrying an IPv6 nexthop (ipv6unicast/ipv6vpn). Assert the
+	// port owns a link-local that is not shared with its backing veth. Note the
+	// shadow shares its link-local with grout's own tap_<iface> by design; only a
+	// collision with the veth is the failure signature.
+	//
+	// This test is marked as failing until we fix this bug.
+	It("assigns each underlay port a unique IPv6 link-local", func() {
```
fedepaol: we should use the label instead of skipping

e2etests/tests/sessions.go
```diff
@@ -1098,4 +1099,69 @@ var _ = Describe("Underlay explicit address family configuration", Ordered, Grou
 			networklayerprotocol.NLP{AFI: networklayerprotocol.IPv6, SAFI: networklayerprotocol.Unicast},
 		)
 	})
+
+	// Regression test for https://github.com/openperouter/openperouter/issues/720.
```
fedepaol, line 1103: I am a bit torn about the value of this test, we fixed the issue and we'll carry this forever, adding more test time, dropping the comment to engage a discussion, but I can live with it

e2etests/pkg/openperouter/interface.go
```diff
@@ -78,6 +79,40 @@ func InterfaceIPv4InNetns(nodeName, intf, ns string) (string, error) {
...
 
+// NetnsLinkLocalOwners returns, for the given netns on nodeName, a map from
+// each usable IPv6 link-local (fe80::/10) address to the interfaces that carry
+// it. A correct setup has a single owner per address; a link-local shared
+// across interfaces (the issue #720 failure signature, caused by a duplicate
+// MAC) shows up as more than one owner.
+//
+// Addresses still in DAD (tentative) or that lost DAD (dadfailed) are skipped:
+// they are not usable and, being exactly the transient artifacts of the
+// collision this test guards against, would otherwise produce flaky ownership
```
fedepaol, line 88: please go over the comments, I think we are saying the same thing in multiple places.

e2etests/tests/sessions.go
maiqueb: this one is also a good candidate to be added to the `AGENTS.md`.

e2etests/pkg/openperouter/interface.go
maiqueb: Done.

e2etests/tests/sessions.go
maiqueb: I prefer to keep it since it'll be helpful to spot regressions. 

If we have another deterministic test that catches this regression we can delete. But I think we don't - otherwise, we wouldn't have the flake.

e2etests/tests/sessions.go
maiqueb: hm, but I don't follow. the label is already on the describe block. Best I can do is go to the automation script that runs the tests and manually add this test to the `--ginkgo.Skip` list of stuff.

Which I think is more trouble that it's worth.

internal/grout/underlay.go
maiqueb: done

internal/hostnetwork/link_properties.go
maiqueb: Done.

e2etests/tests/sessions.go
fedepaol: sorry I put the comment on the wrong line I guess. It's about the if below - 	if !GroutMode {

e2etests/tests/sessions.go
fedepaol: it won't spot regressions - because we fixed this. And it's a tradeoff between we'll break this again - and this is useful, or we won't break it because the code is there and written as such, and we'll pay the test running time toll forever.

Next time it will be another quirk on another bit that we set wrongly. This is on par on checking that we set the right port on the vxlan interface, which we are not doing, and we are not doing it because we see that the e2e flow works.

I am fine with leaving this, what I am afraid of is to pollute the suite with a ton of low-level, non functional tests because we'll keep finding corner cases that don't happen in the happy path - or they happen rarely.

Hence the discussion which you can feel free to ignore :-)

fedepaol:
lgtm

e2etests/tests/sessions.go
zeeke: I understand @fedepaol 's point, but without this test, if a regression happens (maybe due to a refactor), we will catch the problem after weeks of flakes.

If there are any better deterministic test case that catches this error, ok with dropping this. as of now, I'm +1 on keeping the test

maiqueb:
The grout lane [failed](https://github.com/openperouter/openperouter/actions/runs/34853681685/job/104040536277?pr=755) for a different reason.

As before, the test times out because the worker node's BGP session (192.168.11.4 = pe-kind-worker) never establishes. But, when tracing it into the grout dataplane on that node (router-q49qp), we see grout fails to configure the underlay port with a different error:

```
ERR: ADDRESS: ip6_iface_llocal_init: iface_addr_add: No space left on device (28)
grcli address add 2001:db8:11::4/64 iface u_toswitch1 failed: {"error":"...ENOSPC","errno":28}
```

Note the errno: 28 (ENOSPC / "No space left on device").

Grout's address table is exhausted. This PR fixes flake caused by errno 98 ("Address already in use").

It seems there's a second, independent grout problem.

This in turn, led to a leaked IPv6 link-local address, which left us again in the same problem.

Proceeding with the merge; will investigate in parallel the grout issue.

PR #760, opened 2026-09-14
Dockerfile: bundle the vlan and tuning CNI plugins

fivetime:

/kind feature

`CNIDevice` accepts CNI configurations, but the controller image does not contain `vlan` or `tuning`. A configuration requesting either plugin therefore fails plugin lookup unless users supply external binaries or rebuild the image.

The existing CNI builder already compiles both plugins. Copy them into `/opt/openperouter/cni/bin/` alongside the existing plugins and document the available bundle. This enables tagged VLAN underlays using the existing IPAM plugins and chained interface tuning, such as MTU changes.

- This is two additional Dockerfile COPY instructions plus the configuration reference update. The upstream CNI repository and pinned revision are unchanged.
- VLAN still requires IPAM at the pinned revision. Optional IPAM is proposed independently in https://github.com/containernetworking/plugins/pull/1302; this PR does not depend on that change or a personal fork.
- Built the complete image locally on Linux amd64 with `docker build -t local/openperouter:bundle-vlan-tuning-review .`.
- Both bundled binaries successfully handle `CNI_COMMAND=VERSION` inside the resulting image.
- A smoke test in an isolated container and child network namespace verified VLAN ID 123, static IPAM address assignment, a chained tuning ADD setting MTU 1400, and tuning/VLAN DEL cleanup.
- `git diff --check` passes. No live-cluster end-to-end test or local arm64 build was run.

**Release note**:

```release-note
The controller image now includes the vlan and tuning CNI plugins for tagged underlay interfaces and chained interface configuration.
```


fedepaol:
hey @fivetime! Sorry for the delay. We've been discussing about allowing extra cnis and we thought it'd be better to provide a user a way to bring their own CNI. 
This would require changing the helm chart a bit, because even though we provide a configurable cni folder https://github.com/openperouter/openperouter/blob/main/charts/openperouter/templates/controller.yaml#L65 , there is no way to mount a volume.

By doing this, a user could plumb their own binaries independently from what we ship. Are you interested in contributing this variant? If not we can work on that.

fedepaol:
I am going to close this for now, but we can keep the discussion either here or in an issue.

PR #771, opened 2026-09-21
fix(grout): classify grcli failures by errno

zeeke:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression
> /kind example

ensureAddress took any grout error whose text contains "already" for success. grout answers EADDRINUSE, "Address already in use", when it holds a dangling nexthop left over by an earlier failed add and the address was in fact not assigned; only EEXIST means the address is on the port. Taking EADDRINUSE for success loses the address, because the caller then drops it from the kernel interface it reads the underlay addresses from.

A failing grcli prints the errno of the underlying failure in its JSON output. Parse it into a groutError that wraps the command error, so the existing messages are unchanged, and match on the errno instead.

**Release note**:

```release-note
Grout: use error numbers when parsing grcli output
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of network command failures by interpreting structured error codes rather than relying on error message text.
  - Correctly treats existing addresses as successful no-op operations.
  - More reliably distinguishes missing interfaces from other network errors.
  - Preserves error reporting for address conflicts, storage limits, and other unexpected failures.


internal/grout/grout_client.go
```diff
@@ -100,11 +102,16 @@ func (c *Client) deletePort(ctx context.Context, name string) error {
 }
 
 // ensureAddress assigns an IP address (in CIDR notation) to a grout port.
-// If the address is already assigned, it is a no-op.
+// If the address is already assigned to that port, it is a no-op.
 func (c *Client) ensureAddress(ctx context.Context, ifaceName, cidr string) error {
 	slog.InfoContext(ctx, "assigning IP to grout port", "iface", ifaceName, "cidr", cidr)
 	err := c.run(ctx, "address", "add", cidr, "iface", ifaceName)
-	if err != nil && strings.Contains(err.Error(), "already") {
+	// Only EEXIST means the address is on the port. EADDRINUSE also reads as
```
fedepaol: this comment adds little value imo, as it refers to a previous bug we are now fixing.
We can add a comment on why EEXIST is the only value that we care about, but I would skip talking about the consequences and us loosing the IP because it's removed from the kernel.

Something like "EEXIST means the address is already bound to the port" and that's it

internal/grout/grout_client.go
```diff
@@ -179,17 +186,27 @@ var execCmd = func(ctx context.Context, name string, args ...string) ([]byte, er
...
+	}
+
+	cmdErr := fmt.Errorf("grcli %s failed: %w, output: %s", strings.Join(args, " "), err, output)
+	groutErr := &groutError{}
+	if jsonErr := json.Unmarshal([]byte(output), groutErr); jsonErr != nil || groutErr.Errno == 0 {
+		return output, cmdErr
 	}
-	return output, nil
+	groutErr.cmdErr = cmdErr
+	return output, groutErr
```
fedepaol, line 211: can we return nil in case of success?

fedepaol (changes requested):

internal/grout/grout_client.go
zeeke: success return path has been moved to L195.

fedepaol:
@zeeke re: release notes, please provide user facing rel notes. In this case we should say if we fixed some bug, or NONE if it's not user facing

